PR #4, opened 2016-11-23
Feature: Return requesting user when retrieving Permission Requests

hochung:


src/main/java/com/kryptnostic/datastore/services/PermissionsManager.java
```diff
@@ -94,12 +95,12 @@
 
     void removePermissionsRequestForEntitySet( UUID id );
     
-    Iterable<PropertyTypeInEntitySetAclRequest> getAllReceivedRequestsForPermissionsOfUsername( String username );
+    Iterable<PropertyTypeInEntitySetAclRequestWithRequestingUser> getAllReceivedRequestsForPermissionsOfUsername( String username );
```
geekbeast: These wrapper classes are getting out of hand.

hochung: Agree - not sure how to name things appropriately though :(

PR #12, opened 2016-12-05
Service fix instead of controller fix

hochung:


geekbeast:
@hochung Approved waiting on build to merge.

PR #14, opened 2016-12-08
Turn on Cql Timestamp to JodaTime DateTime codec

hochung:
Setup codec for Cql Timestamp, Date, Time to JodaTime DateTime, Local Date, LocalTime

### DIFF
--- a/src/main/java/com/kryptnostic/conductor/codecs/TimestampDateTimeTypeCodec.java
+++ b/src/main/java/com/kryptnostic/conductor/codecs/TimestampDateTimeTypeCodec.java
@@ -1,0 +1,68 @@
+package com.kryptnostic.conductor.codecs;
+
+import java.nio.ByteBuffer;
+
+import org.joda.time.DateTime;
+
+import com.datastax.driver.core.DataType;
+import com.datastax.driver.core.ParseUtils;
+import com.datastax.driver.core.ProtocolVersion;
+import com.datastax.driver.core.TypeCodec;
+import com.datastax.driver.core.exceptions.InvalidTypeException;
+
+/**
+ * TypeCodec between Cql Timestamp and Joda DateTime, modified from default TimestampCodec
+ * @author Ho Chung Siu
+ *
+ */
+public class TimestampDateTimeTypeCodec extends TypeCodec<DateTime> {
+    private static final TimestampDateTimeTypeCodec instance = new TimestampDateTimeTypeCodec();
+    
+    private TimestampDateTimeTypeCodec() {
+        super(DataType.timestamp(), DateTime.class);
+    }
+    
+    public static TimestampDateTimeTypeCodec getInstance(){
+        return instance;
+    }
+    
+    @Override
+    public DateTime parse(String value) {
+        if (value == null || value.isEmpty() || value.equalsIgnoreCase("NULL"))
+            return null;
+        // strip enclosing single quotes, if any
+        if (ParseUtils.isQuoted(value))
+            value = ParseUtils.unquote(value);
+
+        if (ParseUtils.isLongLiteral(value)) {
+            try {
+                return new DateTime(Long.parseLong(value));
+            } catch (NumberFormatException e) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/kryptnostic/conductor/codecs/TimestampDateTimeTypeCodec.java
```diff
@@ -0,0 +1,68 @@
...
+        } catch (IllegalArgumentException e) {
+            throw new InvalidTypeException(String.format("Cannot parse timestamp value from \"%s\"", value));
+        }
+    }
+
+    @Override
+    public String format(DateTime value) {
+        if (value == null)
+            return "NULL";
+        return Long.toString(value.getMillis());
```
geekbeast: What's the reason for not using the DateTime representation of a string?

src/main/java/com/kryptnostic/conductor/rpc/ResultSetAdapterFactory.java
```diff
@@ -75,12 +90,21 @@ public static Entity mapRowToEntity( Row row, Set<PropertyType> properties ) {
 	public static SetMultimap<FullQualifiedName, Object> mapRowToObject( Row row, Collection<PropertyType> properties ) {
 		SetMultimap<FullQualifiedName, Object> map = HashMultimap.create();
 		properties.forEach(property -> {
-			Object value = row.getObject( "value_" + property.getTypename() );
+			Object value = getObject( row, CassandraEdmMapping.getCassandraType( property.getDatatype() ), "value_" + property.getTypename() );
```
geekbeast: This probably legacy badness from way back-- do you remember why we're prefixing `value_`?

src/main/java/com/kryptnostic/conductor/codecs/TimestampDateTimeTypeCodec.java
hochung: TypeCodec.format is supposed to format the JavaType as a CQL string. TypeCodec.deserialize handles what you are thinking about I guess.

src/main/java/com/kryptnostic/conductor/rpc/ResultSetAdapterFactory.java
hochung: We probably can revamp this a little bit - the main reason is that when @Bloodyhacker creates the cache table, he prefixes the column names by `value_`
https://github.com/dataloom/kindling/blob/develop/src/main/java/com/kryptnostic/sparks/ConductorSparkImpl.java#L372

src/main/java/com/kryptnostic/conductor/codecs/TimestampDateTimeTypeCodec.java
geekbeast: I'm not really sure when format gets called here as it is supposed to prefer ByteBuffer serialization, but based off the docs it looks like the CQL string literal it has to be in ISO 8601 format. It does say you can use an integer for input, but I think that's means actual integer not string. What do other codecs do?

geekbeast: `InstantCodec` uses  `java.time.format.DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSxxx")`

src/main/java/com/kryptnostic/conductor/rpc/ResultSetAdapterFactory.java
geekbeast: Ah okay

src/main/java/com/kryptnostic/conductor/codecs/TimestampDateTimeTypeCodec.java
hochung: Hmm good catch. Apparently format is called in these occasions:
http://docs.datastax.com/en/drivers/java/3.1/com/datastax/driver/core/TypeCodec.html#format-T-
Anyway I updated the format function.

PR #19, opened 2017-01-07
Feature/write data

hochung:
- Update CassandraPropertyReader, Row Adapters to use UUID instead.
- Added Helper methods to get authorized properties among user-submitted ones.


src/main/java/com/kryptnostic/datastore/cassandra/CassandraPropertyReader.java
```diff
@@ -1,26 +1,26 @@
...
 
-import org.apache.olingo.commons.api.edm.FullQualifiedName;
+import java.util.UUID;
 
 import com.datastax.driver.core.Row;
 import com.google.common.base.Function;
 
 public class CassandraPropertyReader implements Function<Row, Object> {
-    private final FullQualifiedName     type;
+    private final UUID                  typeId;
```
geekbeast: This should stay FQN for our purposes or we will have to create to instance a UUID based one and an FQN based one. The reason is most humans will expect FQNs as columns when reading data and not UUIDs. Thus for data export, we should use FQNs.

PR #41, opened 2017-01-16
Fix getCurrentUser principal

mindblight:
getCurrentUser creates the Principal from the Auth0 user_id instead of the username

### DIFF
--- a/src/main/java/com/dataloom/authorization/Principals.java
+++ b/src/main/java/com/dataloom/authorization/Principals.java
@@ -30,7 +30,7 @@ public final class Principals {
         return new Principal(
                 PrincipalType.USER,
                 ( (Auth0UserDetails) SecurityContextHolder.getContext().getAuthentication().getPrincipal() )
-                        .getUsername() );
+                        .getAuth0Attribute("user_id").toString() );
     }
 
     public static Set<Principal> getCurrentPrincipals() {


src/main/java/com/dataloom/authorization/Principals.java
```diff
@@ -30,7 +30,7 @@ public static Principal getCurrentUser() {
         return new Principal(
                 PrincipalType.USER,
                 ( (Auth0UserDetails) SecurityContextHolder.getContext().getAuthentication().getPrincipal() )
-                        .getUsername() );
+                        .getAuth0Attribute("user_id").toString() );
```
hochung: Just wondering - did you test this? I will look at this and try this out very soon, but I vaguely remember that getting "user_id" always return null for me in the past.

mindblight: I ran it locally :p. This code is running against the standard auth0 integration, and the auth0 docs guarantee this property, but they have been wrong before. 

We definitely want to move to this - username is volatile and can be pretty easily faked. Auth0UserDetails tries to fill the username field with a lot of different properties, starting with `email`. That leaves accounts that have been created with username/pass open to an attack where a malicious user creates a FB account with the victims email, then signs in using FB.

hochung: Cool! If you tested that out it should be fine 👍

geekbeast:
I'm going to decline this one, because I have a change in my PR that includes the fix with the `user_id` magic string as a constant.

PR #42, opened 2017-01-16
move up the entity type existence check

katherinebernstein:


src/main/java/com/kryptnostic/datastore/services/EdmService.java
```diff
@@ -229,11 +229,11 @@ private void createEntitySet( EntitySet entitySet ) {
     public void createEntitySet( Principal principal, EntitySet entitySet ) {
         try {
             Principals.ensureUser( principal );
-
-            createEntitySet( entitySet );
-
             EntityType entityType = checkNotNull( entityTypes.get( entitySet.getEntityTypeId() ),
```
geekbeast: switch to  `checkState( entityType.containsKey(  ), <error message>  )`

PR #45, opened 2017-01-17
Feature/fix read principal

geekbeast:
Ho was right about getting NPE. It turns out that the only require field where auth0 puts the id is SUBJECT not USER_ID.


hochung:
Arion and I talked about this again earlier this afternoon. It turns out that it depends on where you get the JWT token from. If you get it from Auth0Test, that one (by default) won't be decorated with `user_id` and other fields. But if you get it from gallery, which is configured to have the params
```
         params: {
          scope: 'openid email user_metadata app_metadata nickname roles user_id'
        }
```
Then it would come with the user_id field. So using it from gallery would not be a problem.

geekbeast:
@hochung This PR means no adding mapstores without making sure R/W works with test data.

geekbeast:
@hochung Ah. That makes sense, I updated it to include user_id and it seems that for auth0 the JWT subject claim and the user_id claim are identical.

from rfc:
```
 The "sub" (subject) claim identifies the principal that is the
   subject of the JWT.  The claims in a JWT are normally statements
   about the subject.  The subject value MUST either be scoped to be
   locally unique in the context of the issuer or be globally unique.
   The processing of this claim is generally application specific.  The
   "sub" value is a case-sensitive string containing a StringOrURI
   value.  Use of this claim is OPTIONAL.
```

geekbeast:
Also sub is the right thing to use: https://auth0.com/forum/t/is-the-sub-a-reliable-way-to-use-the-user-id/3834

PR #60, opened 2017-01-22
bugfix/BACKEND-626

hochung:


geekbeast:
I added a commit to cleanup the lambdas a bit.

PR #75, opened 2017-01-24
enum set instead

hochung:


geekbeast:
Fixed stream serializer instead.

PR #80, opened 2017-02-07
Feature/index organizations

katherinebernstein:


src/main/java/com/kryptnostic/conductor/rpc/ConductorSparkApi.java
```diff
@@ -14,6 +14,7 @@
 import com.dataloom.data.requests.LookupEntitiesRequest;
 import com.dataloom.edm.internal.EntitySet;
 import com.dataloom.edm.internal.PropertyType;
+import com.dataloom.organization.Organization;
 import com.google.common.base.Optional;
 
 public interface ConductorSparkApi {
```
geekbeast: This starting to get big-- we should break into sub-interfaces when we get the chance.

src/main/java/com/kryptnostic/conductor/rpc/ConductorElasticsearchApi.java
```diff
@@ -39,6 +47,8 @@
 
 	Boolean initializeEntitySetDataModelIndex();
```
geekbeast: Prefer primitives to boxed types. Is there a reason all the boolean variables are boxed on this interface?
https://books.google.com/books?id=ka2VUBqHiWkC&pg=PA221&lpg=PA221&dq=prefer+primitive+types+to+boxed+primitives&source=bl&ots=yZKlOeo_NZ&sig=yNFIINhohAN2D398nEI1ZQt9Zio&hl=en&sa=X&ved=0ahUKEwiX6sbZ5IHSAhWCslQKHbYoAe0Q6AEINjAE#v=onepage&q=prefer%20primitive%20types%20to%20boxed%20primitives&f=false

src/main/java/com/kryptnostic/conductor/rpc/ConductorSparkApi.java
katherinebernstein: +1, I thought I had filed a ticket for myself on this but I can't find it

src/main/java/com/kryptnostic/conductor/rpc/ConductorElasticsearchApi.java
katherinebernstein: because the return type needs to be used by the lambdas for Function<ConductorSparkApi, Boolean>

PR #82, opened 2017-02-08
Feature/index entity data

katherinebernstein:


src/main/java/com/kryptnostic/conductor/rpc/EntityDataLambdas.java
```diff
@@ -0,0 +1,26 @@
+package com.kryptnostic.conductor.rpc;
+
+import java.io.Serializable;
+import java.util.Map;
+import java.util.UUID;
+import java.util.function.Function;
+
+public class EntityDataLambdas implements Function<ConductorSparkApi, Boolean>, Serializable {
```
geekbeast: If you change to implements `Predicate<ConductorSparkApi>, Serializable` and make necessary changes in `KryoSerializer` you can avoid boxing and everything else should pretty much stay the same.

katherinebernstein: does this involve rewriting ConductorCall?

geekbeast: Yeah, it might require creating a ConductorPredicate wrapper. I'm going to approve for now, but filed tracking bug.

PR #84, opened 2017-02-09
update entity sets/property types in entity sets

katherinebernstein:


src/main/java/com/kryptnostic/datastore/services/EdmService.java
```diff
@@ -380,6 +381,7 @@ public void renamePropertyType( UUID propertyTypeId, FullQualifiedName newFqn )
     public void renameEntitySet( UUID entitySetId, String newName ) {
         aclKeyReservations.renameReservation( entitySetId, newName );
         entitySets.executeOnKey( entitySetId, new RenameEntitySetProcessor( newName ) );
+        eventBus.post( new EntitySetMetadataUpdatedEvent( entitySetManager.getEntitySet( newName ) ) );
```
hochung: would be better to do 
```eventBus.post( new EntitySetMetadataUpdatedEvent( getEntitySet( entitySetId ) ) );```
instead - the other call is a secondary index lookup which could be expensive.

katherinebernstein: ah makes sense! fixing it

PR #85, opened 2017-02-09
add event bus for renaming property type

hochung:
(Opening a PR here just to make sure I won't forget about it later - this should be ready to merge once the issue about `List` vs `Set` is resolved)

### DIFF
--- a/src/main/java/com/dataloom/edm/properties/CassandraTypeManager.java
+++ b/src/main/java/com/dataloom/edm/properties/CassandraTypeManager.java
@@ -4,17 +4,16 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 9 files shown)


src/main/java/com/dataloom/edm/properties/CassandraTypeManager.java
```diff
@@ -3,6 +3,7 @@
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import org.spark_project.guava.collect.Iterables;
```
hochung: er, what do you mean?

hochung: oh do you mean the Iterables one?

PR #96, opened 2017-02-15
consts

katherinebernstein:
if this merges and the kindling branch doesn't then creating indices will be broken

### DIFF
--- a/src/main/java/com/kryptnostic/conductor/rpc/ConductorElasticsearchApi.java
+++ b/src/main/java/com/kryptnostic/conductor/rpc/ConductorElasticsearchApi.java
@@ -32,6 +32,22 @@ import com.dataloom.organization.Organization;
 import com.google.common.base.Optional;
 
 public interface ConductorElasticsearchApi {
+    
+    // settings consts
+    final String NUM_SHARDS = "number_of_shards";
+    final String NUM_REPLICAS = "number_of_replicas";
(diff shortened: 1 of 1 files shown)


src/main/java/com/kryptnostic/conductor/rpc/ConductorElasticsearchApi.java
```diff
@@ -32,17 +32,32 @@
...
+    final String ANALYSIS = "analysis";
+    final String FILTER = "filter";
+    final String ANALYZER = "analyzer";
+    final String ENCODER = "encoder";
+    final String REPLACE = "replace";
+    final String TOKENIZER = "tokenizer";
+    final String STANDARD = "standard";
+    final String LOWERCASE = "lowercase";
+    final String PHONETIC = "phonetic";
+    final String METAPHONE = "metaphone";
```
hochung: Maybe worthwhile to batch `METAPHONE`, `METAPHONE_FILTER` and `METAPHONE_ANALYZER` into one single enum constant, which will be passed in by the relevant field in property type instantiation.

katherinebernstein: those three consts all have different functions (encoder, filter, and analyzer), but yeah it would be good to have an enum for either different analyzers or different encoders once we allow for that

PR #97, opened 2017-02-16
creating SearchResult object and serializer

katherinebernstein:


src/main/java/com/dataloom/hazelcast/serializers/SearchResultStreamSerializer.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public void write( ObjectDataOutput out, SearchResult object ) throws IOException {
+        out.writeLong( object.getNumHits() );
+        out.writeObject( object.getHits() );
+    }
+
+    @Override
+    public SearchResult read( ObjectDataInput in ) throws IOException {
+        long numHits = in.readLong();
+        List<Map<String, Object>> hits = in.readObject();
```
geekbeast: In cases like where we know most of this stuff safely serialized as JSON, we can use the SmileMapper to generate an efficient binary representation. Alternatively, you could also use Kryo as well, but that is more complicated to setup due to thread safety issues.  https://jira.thedataloom.com/browse/BACKEND-832

PR #140, opened 2017-03-23
Feature/track latest sync ids

katherinebernstein:


src/main/java/com/kryptnostic/datastore/cassandra/CommonColumns.java
```diff
@@ -110,7 +110,8 @@
     REASON( DataType.text() ),
     BASE_TYPE( DataType.uuid() ),
     FLAGS( DataType.cboolean() ),
-    CATEGORY( DataType.text() );
+    CATEGORY( DataType.text() ),
+    SYNC_ID( DataType.uuid() );
```
katherinebernstein: ooh cool, I'll switch to that one. Also--should SYNCIDS be a ( DataType.list( DataType.timeuuid() ) ), instead of ( DataType.list( DataType.uuid() ) ) ?

PR #143, opened 2017-03-28
save entity type metadata updates to elasticsearch

katherinebernstein:


src/main/java/com/kryptnostic/datastore/services/EdmService.java
```diff
@@ -570,6 +570,7 @@ public void updateEntityTypeMetadata( UUID entityTypeId, MetadataUpdate update )
             aclKeyReservations.renameReservation( entityTypeId, update.getType().get() );
         }
         entityTypes.executeOnKey( entityTypeId, new UpdateEntityTypeMetadataProcessor( update ) );
+        eventBus.post( new EntityTypeCreatedEvent( getEntityType( entityTypeId ) ) );
```
UnsungHero97: How does this translate to "save updates to elastic search"?

katherinebernstein: because it saves the updated entity type to elasticsearch. check out SearchService

PR #144, opened 2017-03-30
add vertex/ fix typo; slight change to api method name

hochung:


src/main/java/com/dataloom/graph/core/LoomGraph.java
```diff
@@ -9,25 +9,35 @@
...
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.IMap;
 
 public class LoomGraph implements LoomGraphApi {
     
+    private static UUID DEFAULT_GRAPH_ID = new UUID(0, 0);
     private UUID graphId;
     
-    private IMap<GraphWrappedVertexId, LoomVertex> vertices;
+    private static IMap<GraphWrappedVertexId, LoomVertex> vertices;
```
UnsungHero97: Why are you changing `vertices` and `verticesLookup` to being static, and changing the constructor to a static `init` method?

hochung: All graphs are stored in one table/Imap, so all graph instances can share the same imap

PR #150, opened 2017-04-05
update data + delete entity functionality for elasticsearch

katherinebernstein:


src/main/java/com/kryptnostic/datastore/services/CassandraDataManager.java
```diff
@@ -351,6 +391,7 @@ public void deleteEntitySetData( UUID entitySetId ) {
             String entityId = RowAdapters.entityId( entityIdRow );
 
             results.add( asyncDeleteEntity( entitySetId, entityId ) );
+            eventBus.post( new EntityDataDeletedEvent( entitySetId, entityId, Optional.absent() ) );
```
hochung: I'm not sure what level of logging is necessary, but is it okay to just log the whole entity set being deleted rather than logging every single entry being deleted?

katherinebernstein: what logging?

PR #156, opened 2017-04-14
AUDIT_LOG table

UnsungHero97:
new `AUDIT_LOG` table to replace `AUDIT_EVENTS` and `AUDIT_METRICS`


src/main/java/com/kryptnostic/datastore/cassandra/CommonColumns.java
```diff
@@ -30,19 +30,22 @@
 
 public enum CommonColumns implements ColumnDef {
     ACLID( DataType.uuid() ),
+    AUDIT_ID( DataType.uuid() ),
```
UnsungHero97: @geekbeast you said audit ID is just a vertex ID. Should `AUDIT_ID` be a `UUID`?

src/main/java/com/kryptnostic/datastore/cassandra/CommonColumns.java
```diff
@@ -30,19 +30,22 @@
...
     FROZEN_PERMISSIONS( DataType.frozenSet( DataType.text() ) ),
     ACL_CHILDREN_PERMISSIONS( DataType.map( DataType.uuid(), FROZEN_PERMISSIONS.getType() ) ),
+    // TODO: ACL_KEYS should really be ACL_KEY
     ACL_KEYS( DataType.frozenList( DataType.uuid() ) ), // partition index within a table for distribution purpose,
     ACL_ROOT( DataType.frozenList( DataType.uuid() ) ),
     ALLOWED_EMAIL_DOMAINS( DataType.set( DataType.text() ) ),
     AUDIT_EVENT_DETAILS( DataType.text() ),
-    AUDIT_EVENT_TYPE( DataType.text() ),
     BIDIRECTIONAL( DataType.cboolean() ),
+    BLOCK_ID( DataType.uuid() ),
```
UnsungHero97: @geekbeast you said block ID is a pointer to a new table that we'll create for blockchain stuff. Should `BLOCK_ID` be a `UUID`?

src/main/java/com/kryptnostic/datastore/cassandra/CommonColumns.java
```diff
@@ -30,19 +30,22 @@
...
     CLOCK( DataType.timestamp() ),
     DATATYPE( DataType.text() ),
     DESCRIPTION( DataType.text() ),
     DEST( DataType.set( DataType.uuid() ) ),
     ENTITYID( DataType.text() ),
+    ENTITY_ID( DataType.uuid() ),
```
UnsungHero97: @geekbeast you said entity ID is just a vertex ID for the corresponding `AclKey`. Should `ENTITY_ID` be a `UUID`?

src/main/java/com/kryptnostic/conductor/rpc/odata/Table.java
```diff
@@ -137,13 +138,44 @@ static CassandraTableBuilder createTableDefinition( Table table ) {
...
+                                CommonColumns.EVENT_TYPE
+                        )
+                        .clusteringColumns(
+                                CommonColumns.PRINCIPAL_TYPE,
+                                CommonColumns.PRINCIPAL_ID,
+                                CommonColumns.TIME_UUID
+                        )
+                        .columns(
+                                CommonColumns.ENTITY_ID,
+                                CommonColumns.AUDIT_ID,
```
UnsungHero97: @geekbeast what purpose does `AUDIT_ID` serve?

src/main/java/com/kryptnostic/datastore/cassandra/CommonColumns.java
hochung: `AUDIT_ID` should be `UUID`

src/main/java/com/kryptnostic/datastore/cassandra/CommonColumns.java
hochung: see below

UnsungHero97: where?

src/main/java/com/kryptnostic/conductor/rpc/odata/Table.java
```diff
@@ -137,13 +138,44 @@ static CassandraTableBuilder createTableDefinition( Table table ) {
...
+                                CommonColumns.ACL_KEYS,
+                                CommonColumns.EVENT_TYPE
+                        )
+                        .clusteringColumns(
+                                CommonColumns.PRINCIPAL_TYPE,
+                                CommonColumns.PRINCIPAL_ID,
+                                CommonColumns.TIME_UUID
+                        )
+                        .columns(
+                                CommonColumns.ENTITY_ID,
```
hochung: I think this column is supposed to reference the actual data that got changed. In other words, this column should be a vertexId, or something like that.

Also since we haven't documented this anywhere, I will just leave a brief description of what vertex_id and entity key do. Essentially,
- one row of an entity set is identified by entity_id and sync id. Here entity_id is a string either the user specifies, or we compute in shuttle.
- therefore across all the entity sets, one row of data is uniquely identified by the triple (entity_set_id, entity_id, syncId). We call this an entity key.
- When a row of data is first written to the system, we would register it as a vertex which gives it a unique uuid. In particular, an entity key (hence a row of data) is uniquely identified by the uuid. This is the vertex id we've been talking about.

hochung: Also since we haven't documented this anywhere, I will just leave a brief description of what vertex_id and entity key do. Essentially,
- one row of an entity set is identified by entity_id and sync id. Here entity_id is a string either the user specifies, or we compute in shuttle.
- therefore across all the entity sets, one row of data is uniquely identified by the triple (entity_set_id, entity_id, syncId). We call this an entity key.
- When a row of data is first written to the system, we would register it as a vertex which gives it a unique uuid. In particular, an entity key (hence a row of data) is uniquely identified by the uuid. This is the vertex id we've been talking about.

src/main/java/com/kryptnostic/conductor/rpc/odata/Table.java
hochung: We will write the Audit metadata to a big entity set `Audit`. As with other entity sets we will just store such metadata in the data table. This `AUDIT_ID` should then reference to the row of data in data table, that references this audit metadata

src/main/java/com/kryptnostic/conductor/rpc/odata/Table.java
UnsungHero97: I'd be ok with `VERTEX_ID`, but the table contains 2 columns that store vertex IDs, so I wanted something more specific. What about `ENTITY_DATA_ID` + `AUDIT_ID`, or `ENTITY_VERTEX_ID` + `AUDIT_VERTEX_ID`?

hochung: I prefer the first one, I may even just use `DATA_ID` + `AUDIT_ID`, although I think all choices here are pretty reasonable

src/main/java/com/kryptnostic/datastore/cassandra/CommonColumns.java
geekbeast: 👍

src/main/java/com/kryptnostic/conductor/rpc/odata/Table.java
UnsungHero97: It sounds like there will be 2 writes that happen, the first for the metadata into the data table, and then the second into this `AUDIT_LOG` table. Is this right? 

If so, I'm confused. What's the audit metadata?

src/main/java/com/kryptnostic/conductor/rpc/odata/Table.java
UnsungHero97: Updated with `DATA_ID` + `AUDIT_ID`

UnsungHero97:
@geekbeast @hochung 

The new `AUDIT_LOG` table is meant to be where we log all events, and it is structured based off the schema we sketched out on the whiteboard. It seems that I'm missing the details around `AUDIT_ID` column of the new `AUDIT_LOG` table. I don't quite understand what the audit metadata is, or how it fits into the pipeline.

From what I can gather based off of your comments, it sounds like there will be 2 separate and sequential writes that happen: the first for the metadata into the data table, and then the second into this `AUDIT_LOG` table, which contains the `UUID` from the first write. Is this right?

If so, I'm confused. What's the audit metadata?

PR #157, opened 2017-04-14
basic structure

UnsungHero97:
Changes to `HazelcastAuthorizationService` are temporary, just in place to help me visualize how Neuron fits into the bigger picture, how to interact with it, etc.

This PR goes hand-in-hand with https://github.com/dataloom/conductor-client/pull/156. They are separated to make it easier to review.

### DIFF
--- a/src/main/java/com/dataloom/authorization/HazelcastAuthorizationService.java
+++ b/src/main/java/com/dataloom/authorization/HazelcastAuthorizationService.java
@@ -69,6 +69,7 @@ public class HazelcastAuthorizationService implements AuthorizationManager {
         this.aqs = checkNotNull( aqs );
         this.eventBus = checkNotNull( eventBus );
 
+        // TODO: this is just a placeholder to help with Neuron implementation. remove before merging into develop.
         aces.addEntryListener( (EntryUpdatedListener) event -> {
 
             // TODO: use actual values instead of null
--- a/src/main/java/com/dataloom/authorization/HazelcastAuthorizationService.java
+++ b/src/main/java/com/dataloom/authorization/HazelcastAuthorizationService.java
@@ -26,11 +26,19 @@ import com.dataloom.authorization.processors.PermissionMerger;
 import com.dataloom.authorization.processors.PermissionRemover;
 import com.dataloom.authorization.securable.SecurableObjectType;
 import com.dataloom.hazelcast.HazelcastMap;
+import com.dataloom.neuron.AuditableSignal;
+import com.dataloom.neuron.Neuron;
+import com.dataloom.neuron.Signal;
+import com.dataloom.neuron.SignalType;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.eventbus.EventBus;
+import com.hazelcast.core.EntryEvent;
+import com.hazelcast.core.EntryListener;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.IMap;
+import com.hazelcast.map.listener.EntryUpdatedListener;
 import com.kryptnostic.datastore.util.Util;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -55,10 +63,25 @@ public class HazelcastAuthorizationService implements AuthorizationManager {
     public HazelcastAuthorizationService(
             HazelcastInstance hazelcastInstance,
             AuthorizationQueryService aqs,
-            EventBus eventBus ) {
+            EventBus eventBus,
+            Neuron neuron ) {
         aces = hazelcastInstance.getMap( HazelcastMap.PERMISSIONS.name() );
         this.aqs = checkNotNull( aqs );
         this.eventBus = checkNotNull( eventBus );
+
+        aces.addEntryListener( (EntryUpdatedListener) event -> {
+
+            // TODO: use actual values instead of null
+            neuron.transmit( new AuditableSignal(
+                    SignalType.ACL_KEY_PERMISSION_UPDATE,
+                    null,
+                    null,
+                    null,
+                    null,
+                    null,
+                    null
+            ) );
+        }, true );
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


src/main/java/com/dataloom/authorization/HazelcastAuthorizationService.java
```diff
@@ -55,10 +63,25 @@
...
             HazelcastInstance hazelcastInstance,
             AuthorizationQueryService aqs,
-            EventBus eventBus ) {
+            EventBus eventBus,
+            Neuron neuron ) {
         aces = hazelcastInstance.getMap( HazelcastMap.PERMISSIONS.name() );
         this.aqs = checkNotNull( aqs );
         this.eventBus = checkNotNull( eventBus );
+
+        aces.addEntryListener( (EntryUpdatedListener) event -> {
```
geekbeast: You will probably have serialization problems here once you're not running in a single jvm

src/main/java/com/dataloom/neuron/Synapse.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * You can contact the owner of the copyright at support@thedataloom.com
+ */
+
+package com.dataloom.neuron;
+
+public interface Synapse<T extends Signal> {
```
geekbeast: There is already a class with this name and I think Receptor void signal( AuditableEvent event ) seems like a better pattern here as it will standardized the envelope instead of requiring the use of generics (at this level).

src/main/java/com/dataloom/neuron/Signal.java
```diff
@@ -0,0 +1,47 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.dataloom.client.serialization.SerializationConstants;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class Signal {
```
geekbeast: I'm not sure where you a going with this class. Is this supposed to be a re-implementation of AuditableEvent with SignalType being the enum of AuditableEventTypes?

src/main/java/com/dataloom/neuron/Neuron.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+
+public class Neuron {
+
+    private static final Logger logger = LoggerFactory.getLogger( Neuron.class );
+
+    private final HazelcastAuditLogService hzAuditLogService;
+
+    private final EnumMap<SignalType, List<Synapse>> synapses = Maps.newEnumMap( SignalType.class );
```
geekbeast: Okay SignalType makes sense-- I would update to LinkedHashSet<Receptor> (if order matters). 

I wouldn't have a HazelcastAuditLogService because we aren't going to be caching audit log events. Just do a query service.

src/main/java/com/dataloom/neuron/Neuron.java
```diff
@@ -0,0 +1,64 @@
...
+            synapses.get( type ).add( synapse );
+        } else {
+            synapses.put( type, Lists.newArrayList( synapse ) );
+        }
+    }
+
+    public void transmit( AuditableSignal signal ) {
+
+        // 1. audit event
+        this.hzAuditLogService.log( signal );
```
geekbeast: Just use AuditLogQueryService no need to use Hazelcast.

src/main/java/com/dataloom/neuron/Neuron.java
```diff
@@ -0,0 +1,64 @@
...
+    private final HazelcastAuditLogService hzAuditLogService;
+
+    private final EnumMap<SignalType, List<Synapse>> synapses = Maps.newEnumMap( SignalType.class );
+
+    public Neuron( HazelcastAuditLogService hzAuditLogService ) {
+
+        this.hzAuditLogService = hzAuditLogService;
+    }
+
+    public void activate( SignalType type, Synapse synapse ) {
```
geekbeast: Following the naming pattern: createSynapse( SignalType type,  Receptor receptor )

src/main/java/com/dataloom/neuron/Neuron.java
UnsungHero97: I was wondering whether it was necessary to have `HazelcastAuditLogService` as well, but I added it in because the existing auditing code has it. Will remove.

src/main/java/com/dataloom/authorization/HazelcastAuthorizationService.java
UnsungHero97: Good to know. Changes to `HazelcastAuthorizationService` are temporary, just in place to help me visualize how Neuron fits into the bigger picture, how to interact with it, etc. Will add a comment to remove before merging into develop.

src/main/java/com/dataloom/neuron/Synapse.java
UnsungHero97: I was initially thinking about this being called `Receptor` as well :)

src/main/java/com/dataloom/neuron/Signal.java
UnsungHero97: In short, yes.

Since we're doing the neuron theme, a "signal" is just an event. `Signal` is meant to be the base class for all events, similar to `AbstractSecurableObject`, and `AuditableSignal` is meant to be the re-implementation of `AuditableEvent` to match the new `AUDIT_LOG` table.

UnsungHero97: I'm not really sure where to go with this class either. Perhaps `AuditableSignal` could serve as the base class for all signals, and this might not be necessary.

Still thinking it all over; just wanted to gather feedback.

src/main/java/com/dataloom/neuron/Neuron.java
UnsungHero97: Done.

src/main/java/com/dataloom/neuron/Neuron.java
UnsungHero97: I changed it to `activateReceptor( SignalType type, Receptor receptor )`

PR #159, opened 2017-04-19
edge updates

katherinebernstein:


src/main/java/com/dataloom/data/storage/CassandraEntityDatastore.java
```diff
@@ -137,6 +137,20 @@ public EntitySetData getEntitySetData(
...
+            UUID entitySetId,
+            UUID syncId,
+            String entityId,
+            Map<UUID, PropertyType> authorizedPropertyTypes ) {
+        ResultSet entity = session.execute( entitySetQuery.bind()
+                .setUUID( CommonColumns.ENTITY_SET_ID.cql(), entitySetId )
+                .setString( CommonColumns.ENTITYID.cql(), entityId )
+                .setSet( CommonColumns.PROPERTY_TYPE_ID.cql(), authorizedPropertyTypes.keySet() )
+                .setUUID( CommonColumns.SYNCID.cql(), syncId ) );
+        return RowAdapters.entity( entity, authorizedPropertyTypes, mapper );
```
hochung: there's an `asyncLoadEntity` method in there, so you can do `asyncLoadEntity( ... ).getUninterruptibly()` to avoid code duplication

src/main/java/com/dataloom/graph/core/LoomGraph.java
```diff
@@ -102,19 +102,6 @@ public void addEdge( EntityKey src, EntityKey dst, EntityKey edgeLabel ) {
     }
```
hochung: looks like this should be deleted too.

src/main/java/com/dataloom/graph/edge/EdgeKey.java
```diff
@@ -31,22 +29,26 @@
...
+    private final UUID srcEntityKeyId;
+    private final UUID dstEntityKeyId;
     private final UUID dstTypeId;
     private final UUID edgeTypeId;
-    private final UUID dstVertexId;
     private final UUID edgeEntityKeyId;
 
-    public EdgeKey( UUID srcVertexId, UUID dstTypeId, UUID edgeTypeId, UUID dstVertexId, UUID edgeEntityKeyId ) {
-        this.srcVertexId = srcVertexId;
+    public EdgeKey( UUID srcEntityKeyId, UUID dstEntityKeyId, UUID dstTypeId, UUID edgeTypeId, UUID edgeEntityKeyId ) {
```
hochung: It may be better to order the five fields in the same order of `Table.java`, so that we don't need to worry about wrong order of columns

src/main/java/com/dataloom/graph/core/LoomGraph.java
katherinebernstein: which one?

hochung: `addEdge( EntityKey src, EntityKey dst, EntityKey edgeLabel)`

PR #160, opened 2017-04-19
(no longer super super untested) top utilizers logic

katherinebernstein:
SUPER NOT TESTED (just putting this here for visibility)


src/main/java/com/dataloom/data/DataGraphService.java
```diff
@@ -228,6 +241,57 @@ public void createEntitiesAndAssociations(
...
+        try {
+            queryId = ByteBuffer.wrap( ObjectMappers.getSmileMapper().writeValueAsBytes( topUtilizerDetailsList ) );
+        } catch ( JsonProcessingException e1 ) {
+            logger.debug( "Unable to generate query id." );
+            return null;
+        }
+        eds.getEntityKeysForEntitySet( entitySetId, syncId ).parallel().map( entityKey -> {
+            UUID vertexId = idService.getEntityKeyId( entityKey );
+            List<ResultSetFuture> countFutures = new ArrayList<>();
+            for ( TopUtilizerDetails details : topUtilizerDetailsList ) {
```
geekbeast: Do a flatMap above and here do `return topUtilizerDetailsList.stream().parallel().map( lm.getEdgeCount(...) )` This will get you a parallel `Stream<ResultSetFuture>`.  Then do `.map( ResultSetFuture::getUninterruptibly )` then do `.map( ResultSet::one )` then do `.mapToDouble( row -> 1.0D*row.getLong(0) )` and finally a `.sum`

PR #168, opened 2017-04-21
Audit EntitySet

UnsungHero97:
- auditing logic is moving into the `com.dataloom.neuron` package
- deprecating everything inside `com.dataloom.auditing` package
- Neuron is responsible for ensuring that the Audit Entity Set is initialized
- write signals to Audit EntitySet
- write signals to Audit Log table


src/main/java/com/dataloom/neuron/Receptor.java
```diff
@@ -19,6 +19,8 @@
 
 package com.dataloom.neuron;
 
+import com.dataloom.neuron.signals.Signal;
+
 public interface Receptor<T extends Signal> {
```
geekbeast: Why are you using generics here?

src/main/java/com/dataloom/neuron/audit/AuditLogQueryService.java
```diff
@@ -24,7 +25,7 @@
     private final Session           session;
     private final PreparedStatement storeQuery;
 
-    public AuditLogQueryService( String keyspace, Session session ) {
+    public AuditLogQueryService( CassandraConfiguration cassandraConfig, Session session ) {
```
geekbeast: Are you using things other than the keyspace?

src/main/java/com/dataloom/neuron/signals/Signal.java
```diff
@@ -17,31 +17,63 @@
...
+    private Principal        principal;
+    private Optional<String> details;
 
     @JsonCreator
-    public Signal( @JsonProperty( SerializationConstants.TYPE_FIELD ) SignalType type ) {
+    public Signal(
+            @JsonProperty( SerializationConstants.TYPE_FIELD ) SignalType type,
+            @JsonProperty( SerializationConstants.ACL_KEY ) List<UUID> aclKey,
+            @JsonProperty( SerializationConstants.PRINCIPAL ) Principal principal,
+            @JsonProperty( SerializationConstants.DETAILS_FIELD ) Optional<String> details ) {
```
geekbeast: If you want to embed arbitrary objects I would make this Optional<Object>

src/main/java/com/dataloom/neuron/Receptor.java
UnsungHero97: I don't have a great reason, but the initial thought process was that `Signal` would be a base class, and different services would have their own `Receptor`s that process their own specific `Signal`s.

src/main/java/com/dataloom/neuron/audit/AuditLogQueryService.java
UnsungHero97: No. I included `CassandraConfiguration` instead of just the keyspace because I wasn't sure how this class will evolve over time, and what it might need.

src/main/java/com/dataloom/neuron/signals/Signal.java
UnsungHero97: Are you referring to `Optional<String> details`?

Actually, now that you mention arbitrary objects, I wanted to talk to you about how that would work. I wasn't sure how to support storing `Signal`s with arbitrary properties, so I decided to start out with just a simple optional String.

UnsungHero97:
Closing this PR, will open a new one.

PR #171, opened 2017-04-25
fix order, remove duplicates

katherinebernstein:


src/main/java/com/dataloom/data/analytics/TopUtilizers.java
```diff
@@ -31,7 +34,11 @@ public void accumulate( LongWeightedId id ) {
     }
 
     public Stream<LongWeightedId> stream() {
-        return utilizers.stream();
+        List<LongWeightedId> temp = Lists.newArrayList();
```
geekbeast: Given that we need an ordered list, we should switch to a ConcurrentSkipListSet instead of a heap, since there aren't any advantages to using a heap if we have to sort after that fact.

PR #177, opened 2017-04-27
Audit EntitySet

UnsungHero97:
- auditing logic is moving into the `com.dataloom.neuron` package
- everything inside `com.dataloom.auditing` package is deprecated
- write signals to Audit EntitySet
- write signals to Audit Log table


src/main/java/com/dataloom/neuron/audit/AuditLogQueryService.java
```diff
@@ -24,23 +44,25 @@
...
+    // TODO: what would CassandraConfiguration be needed for in the future?
+    public AuditLogQueryService( CassandraConfiguration cassandraConfig, Session session ) {
 
         this.session = session;
         this.storeQuery = session.prepare( Table.AUDIT_LOG.getBuilder().buildStoreQuery() );
     }
 
     public void store( AuditableSignal signal ) {
 
+        // TODO: how do I deal with Optionals?
```
katherinebernstein: in what case is `signal.getAclKey()` absent?

UnsungHero97: In the case that an event is not related objects, like "You've successfully verified your email!".

UnsungHero97: That said, I'm still not sure what the base `Signal` class should look like.

src/main/java/com/dataloom/neuron/audit/AuditEntitySet.java
```diff
@@ -0,0 +1,179 @@
...
+    private static final Logger logger = LoggerFactory.getLogger( AuditEntitySet.class );
+
+    private static final String AUDIT_ENTITY_SET_NAME = "Loom Audit Entity Set";
+    private static final String LOOM_AUDIT_NAMESPACE  = "LOOM_AUDIT";
+
+    private static final FullQualifiedName DETAILS_PT_FQN = new FullQualifiedName( LOOM_AUDIT_NAMESPACE, "DETAILS" );
+    private static final FullQualifiedName TYPE_PT_FQN    = new FullQualifiedName( LOOM_AUDIT_NAMESPACE, "TYPE" );
+    private static final FullQualifiedName AUDIT_ET_FQN   = new FullQualifiedName( LOOM_AUDIT_NAMESPACE, "AUDIT" );
+
+    // TODO: where does this belong?
```
hochung: There probably should be a constants file somewhere that we can put it in. Right now there's `DatastoreConstants` in `loom-api`. It probably makes sense to move that to conductor-client and put all these info in there.

src/main/java/com/dataloom/neuron/audit/AuditEntitySet.java
```diff
@@ -0,0 +1,179 @@
...
+                ImmutableSet.of(),
+                EdmPrimitiveTypeKind.String
+        );
+
+        AUDIT_ENTITY_TYPE = new EntityType(
+                AUDIT_ET_FQN,
+                "Loom Audit",
+                "The Loom Audit Entity Type.",
+                ImmutableSet.of(),
+                Sets.newLinkedHashSet( Sets.newHashSet( TYPE_PROPERTY_TYPE.getId() ) ),
```
hochung: May want to guard against the case where some of the properties/entity type are already created, but entity set is not. In that case, the property type Id's/entity type Id's should be obtained by creating property type/creating entity type first, check if they return null and do a getPropertyType or getEntityType in that case.

src/main/java/com/dataloom/neuron/audit/AuditEntitySet.java
```diff
@@ -0,0 +1,179 @@
...
+        setProperties();
+
+        UUID syncId = dataSourceManager.createNewSyncIdForEntitySet( AUDIT_ENTITY_SET.getId() );
+        dataSourceManager.setCurrentSyncId( AUDIT_ENTITY_SET.getId(), syncId );
+    }
+
+    private void setProperties() {
+
+        PROPERTIES = ImmutableList.of(
+                TYPE_PROPERTY_TYPE,
```
hochung: See above for potential type id issue (if property types/entity type/entity set is partially created)

src/main/java/com/dataloom/neuron/audit/AuditLogQueryService.java
hochung: second Katherine's question, and similarly when is principal absent? 
1. `Optional.get()` would throw `IllegalStateException` when the value is absent. So you are really assuming that it must be present already. But in that case, why do we need `Optional`?
2. Also, Cassandra wouldn't allow partition key/clustering key to be null.

src/main/java/com/dataloom/neuron/signals/AuditableSignal.java
```diff
@@ -27,65 +27,54 @@
...
 
-    private List<UUID> aclKey;
-    private Principal  principal;
-    private UUID       timeId;
-    private UUID       entityId;
-    private UUID       auditId;
-    private UUID       blockId;
+    private UUID           auditId;
+    private UUID           timeId;
+    private Optional<UUID> entityId;
```
hochung: dataId?

src/main/java/com/dataloom/neuron/signals/AuditableSignal.java
```diff
@@ -27,65 +27,54 @@
...
             @JsonProperty( SerializationConstants.TYPE_FIELD ) SignalType type,
             @JsonProperty( SerializationConstants.ACL_KEY ) List<UUID> aclKey,
             @JsonProperty( SerializationConstants.PRINCIPAL ) Principal principal,
-            @JsonProperty( SerializationConstants.TIME_ID ) UUID timeId,
-            @JsonProperty( SerializationConstants.ENTITY_ID ) UUID entityId,
+            @JsonProperty( SerializationConstants.DETAILS_FIELD ) Optional<String> details,
             @JsonProperty( SerializationConstants.AUDIT_ID ) UUID auditId,
-            @JsonProperty( SerializationConstants.BLOCK_ID ) UUID blockId ) {
+            @JsonProperty( SerializationConstants.TIME_ID ) UUID timeId,
+            @JsonProperty( SerializationConstants.ENTITY_ID ) Optional<UUID> entityId,
```
hochung: dataId?

src/main/java/com/dataloom/neuron/signals/Signal.java
```diff
@@ -0,0 +1,82 @@
...
+    // TODO: should details be Optional<String>, or Optional<Object>?
+    @JsonCreator
+    public Signal(
+            @JsonProperty( SerializationConstants.TYPE_FIELD ) SignalType type,
+            @JsonProperty( SerializationConstants.ACL_KEY ) Optional<List<UUID>> aclKey,
+            @JsonProperty( SerializationConstants.PRINCIPAL ) Optional<Principal> principal,
+            @JsonProperty( SerializationConstants.DETAILS_FIELD ) Optional<String> details ) {
+
+        this.type = checkNotNull( type );
+        this.details = checkNotNull( details );
```
hochung: Anything wrapped with `Optional` is never null. If you intend whatever it wraps around to be non-null, you probably don't need the `Optional` wrapper.

src/main/java/com/kryptnostic/conductor/rpc/odata/Table.java
```diff
@@ -102,8 +102,8 @@ static CassandraTableBuilder createTableDefinition( Table table ) {
                                 CommonColumns.TIME_UUID
                         )
                         .columns(
-                                CommonColumns.DATA_ID,
                                 CommonColumns.AUDIT_ID,
+                                CommonColumns.DATA_ID,
                                 CommonColumns.BLOCK_ID
                         )
                         .sasi(
```
hochung: Don't use sasi, use ordinary secondary index for now. (Sasi gave us pain in reading data/doing linking earlier)

UnsungHero97: Done.

src/main/java/com/dataloom/neuron/signals/AuditableSignal.java
UnsungHero97: Done.

src/main/java/com/dataloom/neuron/signals/AuditableSignal.java
UnsungHero97: Done.

src/main/java/com/dataloom/neuron/signals/Signal.java
UnsungHero97: The `Optional` object itself can be null though.

src/main/java/com/dataloom/neuron/Neuron.java
```diff
@@ -51,14 +78,65 @@ public void activateReceptor( SignalType type, Receptor receptor ) {
...
 
-        // 2. hand off event to receptors
-        // List<Receptor> receptors = this.receptors.get( signal.getType() );
-        // receptors.forEach( synapse -> synapse.process( signal ) );
+    private UUID writeToAuditEntitySet( Signal signal ) {
+
+        try {
+
+            UUID auditEntitySetId = AuditEntitySet.getId();
+            UUID auditEntitySetSyncId = dataSourceManager.getCurrentSyncId( auditEntitySetId );
```
UnsungHero97: I agree that it makes sense to have the `syncId` be inside `AuditEntitySet`. However, it would have to be static, in which case, how would we do `getCurrentSyncId()`?

UnsungHero97: An afterthought... `dataSourceManager.getCurrentSyncId()` really just calls through to a Hazelcast Map. I could skip `dataSourceManager` entirely, and just use Hazelcast instead?

src/main/java/com/dataloom/neuron/Neuron.java
```diff
@@ -51,14 +78,65 @@ public void activateReceptor( SignalType type, Receptor receptor ) {
...
+            String auditEntityId = UUIDs.random().toString();
```
UnsungHero97: Done.

src/main/java/com/dataloom/neuron/Neuron.java
hochung: hm, first thought is to do something like
`if( entitySetId == AuditEntitySet.getId() ){ AuditEtitySet.setSyncId( syncId ); }`
in `dataSourceManager.setCurrentSyncId()` and `dataSourceManager.createNewSyncIdForEntitySet()`. Not sure if it's good.

I brought this issue up precisely to avoid stressing hazelcast/bringing unnecessary overhead, since audit entity set would be written to fairly often.

src/main/java/com/dataloom/neuron/audit/AuditEntitySet.java
UnsungHero97: Done.

src/main/java/com/dataloom/neuron/audit/AuditEntitySet.java
UnsungHero97: Done.

src/main/java/com/dataloom/neuron/Neuron.java
UnsungHero97: Got it. I agree, it's a good idea to store the `syncId` be inside `AuditEntitySet` and avoid unnecessarily overloading Hazelcast.

Your suggestion is good, but I think it's not enough. I'm not sure what the order of operations is, but isn't it theoretically possible that you'll invoke `AuditEtitySet.setSyncId()` before the EntitySet is actually initialized, therefore setting `syncId` to something not related to the Audit EntitySet?

It seems like we'll need the Audit EntitySet name as well...?

hochung: I'm assuming that `AuditEntitySet` is created when datastore starts up, and that `dataSourceManager` is not called at all during startup; I think both are true as of now.

UnsungHero97: Actually, ignore what I said. I think I confused myself. Updated.

src/main/java/com/dataloom/neuron/audit/AuditLogQueryService.java
UnsungHero97: Got rid of the `Optional`.

PR #182, opened 2017-05-02
refactor logic of uuid-entity key lookup table

hochung:
refactor logic of getting uuid from entity key - the part of writing to lookup table


src/main/java/com/dataloom/data/mapstores/EntityKeysMapstore.java
```diff
@@ -74,7 +82,18 @@ protected EntityKey mapValue( ResultSet rs ) {
         if ( r == null ) {
             return null;
         } else {
-            return RowAdapters.entityKey( r );
+            EntityKey value = RowAdapters.entityKey( r );
+            if( rs.one() != null ){
```
geekbeast: isn't there a has remaining or something similar here?

hochung: Done. :+1:

UnsungHero97:
@hochung take a look at the failing test:

https://bamboo.thedataloom.com/browse/LOOM-CONCLI276-JOB1-3

PR #184, opened 2017-05-03
transmit signal when permission request state is updated

UnsungHero97:


src/main/java/com/dataloom/requests/HazelcastRequestsManager.java
```diff
@@ -82,4 +111,52 @@ public void submitAll( Map<AceKey, Status> statusMap ) {
     public Stream<Status> getStatuses( Stream<AceKey> requestKeys ) {
         return requestKeys.map( Util.getSafeMapper( requests ) );
     }
+
+    private void signalPermissionRequestStatusUpdates( Map<AceKey, Status> aceKeyRequestStatusMap ) {
```
geekbeast: Consider: 
- Using subList instead of creating an entirely new ArrayLIst
- Inlining transmit of the signal.
- breaking it out into a function that processes one AceKey / Status pair at a time that is used a lambda argument

geekbeast: Also, I think you could just write EntryListeners and accomplish something similar without the added complexity of having signal code side by side.

PR #206, opened 2017-07-20
Feature/import data model

katherinebernstein:


src/main/java/com/kryptnostic/datastore/services/EdmService.java
```diff
@@ -838,4 +854,182 @@ public AssociationDetails getAssociationDetails( UUID associationTypeId ) {
...
+        edm.getPropertyTypes().forEach( pt -> {
+            PropertyType existing = null;
+            if ( pt.wasIdPresent() ) {
+                if ( checkPropertyTypeExists( pt.getId() ) ) existing = getPropertyType( pt.getId() );
+            } else {
+                UUID id = getTypeAclKey( pt.getType() );
+                if ( id != null ) existing = getPropertyType( id );
+            }
+            if ( existing == null
+                    || !pt.getType().equals( existing.getType() )
```
UnsungHero97: Do you have to do `.equals()` on all of the fields? Why can't you just do `pt.equals(existing)`?

src/main/java/com/kryptnostic/datastore/services/EdmService.java
```diff
@@ -838,4 +854,182 @@ public AssociationDetails getAssociationDetails( UUID associationTypeId ) {
...
+        edm.getEntityTypes().forEach( et -> {
+            EntityType existing = null;
+            if ( et.wasIdPresent() ) {
+                if ( checkEntityTypeExists( et.getId() ) ) existing = getEntityType( et.getId() );
+            } else {
+                UUID id = getTypeAclKey( et.getType() );
+                if ( id != null ) existing = getEntityType( id );
+            }
+            if ( existing == null
+                    || !et.getType().equals( existing.getType() )
```
UnsungHero97: Same comment about `.equals()` on the object instead of each field.

src/main/java/com/kryptnostic/datastore/services/EdmService.java
```diff
@@ -838,4 +854,182 @@ public AssociationDetails getAssociationDetails( UUID associationTypeId ) {
...
+            if ( at.getAssociationEntityType().wasIdPresent() ) {
+                if ( checkEntityTypeExists( at.getAssociationEntityType().getId() ) )
+                    existing = getAssociationType( at.getAssociationEntityType().getId() );
+            } else {
+                UUID id = getTypeAclKey( at.getAssociationEntityType().getType() );
+                if ( id != null ) existing = getAssociationType( id );
+            }
+            EntityType atEntityType = at.getAssociationEntityType();
+            if ( existing == null
+                    || !atEntityType.getType().equals( existing.getAssociationEntityType().getType() )
```
UnsungHero97: Same comment about `.equals()` on the object instead of each field.

src/main/java/com/kryptnostic/datastore/services/EdmService.java
katherinebernstein: because equals also checks for fields that we don't care about/can't update. like if they've tried to change the datatype of a property type or something, we don't want to prompt an update for that. also if they didn't specify the id of a property type, and we're finding it by the fqn (but the property type is the same), then .equals() will return that they are different. so I was only including checks for fields that will actually prompt an update

UnsungHero97: Got it. Thanks. In that case, I imagine this type of equality check would be needed elsewhere, so I think it makes sense to put this logic somewhere shared, like inside`PropertyType` for example.

geekbeast: I'm skeptical about this type of equality check being needed elsewhere as this highly specific to our rules for supported diffs, but I think the point about moving the diff logic into some other location is valid as it will dramatically improve readability of this function.

geekbeast: @UnsungHero97 Can you file bug for this? I have to merge.

PR #217, opened 2017-08-28
get entity function that takes list of property type ids to populate …

katherinebernstein:
…and omits others


geekbeast:
We're building up quite a lot of these functions that are all very similar.

PR #227, opened 2017-09-07
initialize entity set property types if not initialized

katherinebernstein:


src/main/java/com/kryptnostic/datastore/services/EdmService.java
```diff
@@ -1403,6 +1403,11 @@ else if ( !atEntityType.getType().equals( existing.getAssociationEntityType().ge
 
     @Override
     public EntitySetPropertyMetadata getEntitySetPropertyMetadata( UUID entitySetId, UUID propertyTypeId ) {
+        EntitySetPropertyKey key = new EntitySetPropertyKey( entitySetId, propertyTypeId );
+        if ( !entitySetPropertyMetadata.containsKey( key ) ) {
```
geekbeast: You should do a get and if it returns null do the setup. This isn't a big performance issue, but you are a network call for contains key, sometimes the writes writes, and then always the get. You can eliminate the contains key entirely. 

There's also a race condition when multiple people call this on an entity set. Not sure if has any bad side effects

PR #236, opened 2017-10-06
weighted linking vertex keys and stuff

katherinebernstein:


src/main/java/com/dataloom/linking/WeightedLinkingVertexKeySet.java
```diff
@@ -0,0 +1,11 @@
+package com.dataloom.linking;
+
+import java.util.Collection;
+import java.util.concurrent.ConcurrentSkipListSet;
+
+public class WeightedLinkingVertexKeySet extends ConcurrentSkipListSet<WeightedLinkingVertexKey> {
```
geekbeast: If we aren't concurrently updating this set there should be more efficient data structures than a skip list to use here (i.e a heap).

PR #245, opened 2017-11-06
Feature/app configs

katherinebernstein:
I haven't been able to test locally whether the role permissions actually get added to the entity sets, so I'm a bit worried about that...


src/main/java/com/dataloom/hazelcast/serializers/AclKeySetStreamSerializer.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override public void write( ObjectDataOutput out, AclKeySet object ) throws IOException {
+        SetStreamSerializers.serialize( out, object, aclKey -> {
+            AclKeyStreamSerializer.serialize( out, aclKey );
+        } );
+    }
+
+    @Override public AclKeySet read( ObjectDataInput in ) throws IOException {
+        Set<AclKey> aclKeys = SetStreamSerializers.deserialize( in, AclKeyStreamSerializer::deserialize );
```
geekbeast: Unnecessary double copy. I fixed this in develop don't know why it's not picking it up here.

PR #264, opened 2017-12-09
speed up loading authorized property types on an entity set

katherinebernstein:


src/main/java/com/dataloom/authorization/EdmAuthorizationHelper.java
```diff
@@ -54,12 +54,17 @@ public EdmAuthorizationHelper( EdmManager edm, AuthorizationManager authz ) {
...
-                        ptId ),
-                        Principals.getCurrentPrincipals(),
-                        requiredPermissions ) )
-                .collect( Collectors.toSet() );
+        return authz.accessChecksForPrincipals( selectedProperties.stream()
+                .map( ptId -> new AccessCheck( new AclKey( entitySetId, ptId ), requiredPermissions ) ).collect(
+                        Collectors.toSet() ), Principals.getCurrentPrincipals() )
+                .filter( authorization -> {
+                    boolean allPermissions = true;
+                    for ( Permission permission : requiredPermissions ) {
```
geekbeast: Since all permissions are required I think you can just do.
allPermissions = authorizations.getPermissions.values().stream().allMatch( Function.identity() );

Random comment-- authorizations should just be EnumSet<Permission> for the permissions you are granted. May need to make proper version of the API so front-end can start switching over.

katherinebernstein: Function.identity() wasn't compiling :/

PR #273, opened 2018-01-10
make mergeEdgeAsync actually be async

katherinebernstein:


src/main/java/com/dataloom/linking/aggregators/MergeEdgeAggregator.java
```diff
@@ -75,6 +77,7 @@ public UUID getMergedId( UUID graphId, UUID oldId ) {
         return Util.getSafely( newIds, new LinkingVertexKey( graphId, oldId ) );
     }
 
+    @Async
```
geekbeast: This only works for spring beans.

geekbeast:
LGTM. Have you done a manual test pass on this? If so I'll just merge.

PR #288, opened 2018-03-27
Bugfix/load entities

geekbeast:
This is an inelegant temporary fix for reading entities. Once we migrate and clean out sync ids we will have to rework this path as the current approach requires retrieving all the entity set ids via entity keys in order to load the data.


src/main/java/com/openlattice/data/storage/HazelcastEntityDatastore.java
```diff
@@ -181,13 +176,16 @@ public HazelcastEntityDatastore(
     public Stream<SetMultimap<Object, Object>> getEntities(
             Collection<UUID> ids,
             Map<UUID, PropertyType> authorizedPropertyTypes ) {
-        Predicate entitiesFilter = EntitySets.getEntities( ids );
-        Entities entities = data.aggregate( new EntitiesAggregator(), entitiesFilter );
+        Map<UUID, EntityKey> entityKeyIds = idService.getEntityKeys( ImmutableSet.copyOf( ids ) );
```
UnsungHero97: Remove the commented-out lines in this method.

PR #290, opened 2018-04-03
Bug fix/add apps to org constructor

mariereed:
This change, along with change to datastore OrganizationsController.java I think solves the issue of apps not being included in organization information.


geekbeast:
Woops looks like those changes are already in. GTG.

PR #301, opened 2018-05-24
delete entityKeyId when you delete an entity

katherinebernstein:


src/main/java/com/openlattice/data/storage/HazelcastEntityDatastore.java
```diff
@@ -462,6 +462,7 @@ public void deleteEntitySetData( UUID entitySetId ) {
     public void deleteEntity( EntityDataKey entityDataKey ) {
         try {
             asyncDeleteEntity( entityDataKey ).get();
+            idService.deleteEntityKeyId( entityDataKey.getEntityKeyId() );
```
UnsungHero97: The line above implies it's asynchronous. Could there be an issue with `deleteEntityKeyId()` possibly executing before `asyncDeleteEntity()`?

Also, is `entityDataKey.getEntityKeyId()` guaranteed to return the correct value if `asyncDeleteEntity()` successfully completes before it?

geekbeast: ```.get()``` is blocks until delete completes. It's a relic of us writing all our methods to be asynchronous, but most user invocations being blocking wrappers around the async operation.

PR #306, opened 2018-07-21
Feature/background indexing service

katherinebernstein:


src/main/java/com/openlattice/hazelcast/serializers/BulkEntityDataLambdasStreamSerializer.java
```diff
@@ -0,0 +1,85 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+@Component
+public class BulkEntityDataLambdasStreamSerializer extends Serializer<BulkEntityDataLambdas> {
```
geekbeast: I think you meant to extend StreamSerializer instead of Kryo's Serializer class.

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -698,76 +785,13 @@ public static AppType appType( ResultSet rs ) throws SQLException {
...
-                    default:
-                        objects = null;
-                        logger.error( "Unable to read property type {} for entity {}.",
-                                propertyType.getId(),
-                                entityKeyId );
-                }
-                if ( objects != null ) {
-                    data.putAll( propertyType.getType(), objects );
-                }
+            List<?> objects = propertyValue( rs, propertyType );
```
geekbeast: Nice!

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -52,6 +50,36 @@ const val BATCH_SIZE = 10000
 private val logger = LoggerFactory.getLogger(PostgresEntityDataQueryService::class.java)
 
 class PostgresEntityDataQueryService(private val hds: HikariDataSource) {
+
+    fun getEntitiesById(
```
geekbeast: There is already a streamEntitySet version of this that you can do a toMap() on. Was there are a reason for duplicating the code?

katherinebernstein: I didn't see a version with property type ids instead of FQNs

src/main/java/com/openlattice/hazelcast/serializers/BulkEntityDataLambdasStreamSerializer.java
katherinebernstein: this is using kryo-- following same model as EntityDataLambdasStreamSerializer

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
katherinebernstein: Or does that exist somewhere?

src/main/java/com/openlattice/hazelcast/serializers/BulkEntityDataLambdasStreamSerializer.java
geekbeast: Ah ok. Got it!

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
geekbeast: This has the same signature as ```fun streamableEntitySet(
            entitySetId: UUID,
            entityKeyIds: Set<UUID>,
            authorizedPropertyTypes: Map<UUID, PropertyType>,
            metadataOptions: Set<MetadataOption>,
            version: Optional<Long> = Optional.empty()
    )``` with parameters re-arranged.

katherinebernstein: Is there a way in kotlin to return different types depending on a parameter or something like that? Because one needs to return a `PostgresIterable<SetMultimap<FullQualifiedName, Any>>` and the other needs to return a `PostgresIterable<SetMultimap<UUID, Any>>`

geekbeast: Yes, but for expediency we can leave like this.

PR #315, opened 2018-08-16
add users to global org on initialization

katherinebernstein:


src/main/java/com/openlattice/users/Auth0Synchronizer.java
```diff
@@ -149,6 +153,8 @@ public void createPrincipal( Auth0UserBasic user, String userId ) {
 
         AclKey userAclKey = spm.lookup( principal );
 
+        organizationService.addMembers( organizationAclKey, ImmutableSet.of( principal ) );
```
geekbeast: This should be inside of the if statement for future-proofing. If we ever change the javascript logic in auth0 we don't want people to accidentally get the user role. If you think this is unlikely then you should remove the if statement entirely.

src/main/java/com/openlattice/users/Auth0Synchronizer.java
```diff
@@ -149,6 +153,8 @@ public void createPrincipal( Auth0UserBasic user, String userId ) {
...
+
         if ( user.getRoles().contains( SystemRole.AUTHENTICATED_USER.getName() ) ) {
```
geekbeast: Don't use spm to add principal to principal. Use ```organizationService.addRoletoPrincipalInOrganization```. That way if the logic is ever updated inside of organization this doesn't break.

PR #320, opened 2018-08-30
new functions for help with deleting entity neighbors

UnsungHero97:
https://github.com/openlattice/api/pull/217
https://github.com/openlattice/datastore/pull/279

### DIFF
--- a/src/main/kotlin/com/openlattice/data/DataGraphService.kt
+++ b/src/main/kotlin/com/openlattice/data/DataGraphService.kt
@@ -39,6 +39,7 @@ import com.openlattice.edm.EntitySet
 import com.openlattice.edm.type.PropertyType
 import com.openlattice.graph.core.GraphService
 import com.openlattice.graph.core.NeighborSets
+import com.openlattice.graph.edge.Edge
 import com.openlattice.graph.edge.EdgeKey
 import com.openlattice.hazelcast.HazelcastMap
 import com.openlattice.postgres.DataTables.COUNT_FQN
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -379,6 +380,17 @@ open class DataGraphService(
...
 
     override fun getNeighborEntitySets(entitySetId: UUID): List<NeighborSets> {
-        return lm.getNeighborEntitySets(entitySetId)
+        return graphService.getNeighborEntitySets(entitySetId)
+    }
+
+    override fun getNeighborEntitySetIds(entitySetId: UUID): Set<UUID> {
+        return getNeighborEntitySets(entitySetId)
+                .asSequence()
+                .flatMap { sequenceOf(it.srcEntitySetId, it.edgeEntitySetId, it.dstEntitySetId) }
```
geekbeast: Just making sure you intended to use asSequence here. Sequences cannot be inlined by the compiler and are for lazy evaluation. Given that NeighborSets being returned will most likely fit in memory it's more efficient to flat map with a listOf(...)

UnsungHero97: I used `AnalysisService` as an example:

```
val neighborEntitySets = dgm.getNeighborEntitySets(entitySetId)
val entitySetIds = neighborEntitySets.asSequence()
        .flatMap { sequenceOf(it.srcEntitySetId, it.edgeEntitySetId, it.dstEntitySetId) }
        .toSet()
```

geekbeast: Yeah, that's my bad. I shouldn't have used sequence there.

PR #321, opened 2018-09-06
Lattice 987 clear edges when vertex deleted

pckeyes:
Clears edge and EntityDataKey edge when a vertex is cleared

### DIFF
--- a/src/main/java/com/openlattice/datastore/services/PostgresEntitySetManager.java
+++ b/src/main/java/com/openlattice/datastore/services/PostgresEntitySetManager.java
@@ -20,6 +20,7 @@
 
 package com.openlattice.datastore.services;
 
+import com.google.common.collect.Sets;
 import com.openlattice.data.Entity;
 import com.openlattice.edm.EntitySet;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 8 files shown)


src/main/java/com/openlattice/datastore/services/PostgresEntitySetManager.java
```diff
@@ -112,4 +120,42 @@ public EntitySet getEntitySet( String entitySetName ) {
         }
     }
 
+    public Set<UUID> getAllPropertyTypeIds() {
```
geekbeast: Hazelcast is a more efficient way of gett all property type ids in a set. In fact it's a one liner in the edmService. ```propertyTypes.keySet()```

geekbeast: Also you could have used StreamUtil to collect directly into a set instead of manually managing with try-with-resources but that is pretty minor.

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -146,6 +146,13 @@ open class DataGraphService(
             entityKeyIds: Set<UUID>,
             authorizedPropertyTypes: Map<UUID, PropertyType>
     ): Int {
+        val edgeKeys = entityKeyIds
+                .flatMap { graphService.getEdgeKeysContainingEntity(entitySetId, it) }
+                .toSet()
+        graphService.clearEdges(edgeKeys)
+        edgeKeys.forEach {
```
geekbeast: LATTICE-1314

PR #329, opened 2018-09-18
Feature/newmodel

jokedurnez:
Added in the new features: probability of first and last names.  Please test model performance before merging (I can't).


jokedurnez:
wrong branch

PR #347, opened 2018-10-12
Feature/lattice 1473 read linked data

you-leee:
First bigger change of reading linking entity sets.


src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -668,18 +675,30 @@ fun selectEntitySetWithPropertyTypesAndVersion(
     //@formatter:off
     val columns = setOf(
             ID_VALUE.name,
-            if(metadataOptions.contains(MetadataOption.LAST_WRITE) ) { LAST_WRITE.name } else { "" },
```
geekbeast: columns = setOf(ID_VALUE.name) + metadataOptions.map { it.name }  + authorizedPropertyTypeValues.map(::quote)

geekbeast:
Looks good! A couple of minor updates in requested.

src/main/java/com/openlattice/hazelcast/serializers/EntitySetStreamSerializer.java
```diff
@@ -46,23 +47,38 @@ public void write( ObjectDataOutput out, EntitySet object )
...
     public EntitySet read( ObjectDataInput in ) throws IOException {
-        UUID id = UUIDStreamSerializer.deserialize( in );
+        Optional<UUID> id = Optional.of(UUIDStreamSerializer.deserialize( in ));
         UUID entityTypeId = UUIDStreamSerializer.deserialize( in );
         String name = in.readUTF();
         String title = in.readUTF();
         Optional<String> description = Optional.of( in.readUTF() );
         Set<String> contacts = SetStreamSerializers.deserialize( in, ObjectDataInput::readUTF );
+        Optional<Boolean> linking = Optional.of( in.readBoolean() );
+        Optional<Set<UUID>> linkedEntitySets = Optional.of( SetStreamSerializers.orderedDeserialize(
```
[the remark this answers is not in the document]
you-leee: It does that actually in the underlying AbstractUUIDStreamSerializer class

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
you-leee: Shouldn't we use the Postgres...Definitions for names instead of metadataoptions?

src/main/java/com/openlattice/hazelcast/serializers/EntitySetStreamSerializer.java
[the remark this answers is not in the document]
geekbeast: I don't think so. There's a fastOrderedUUIDSetDeserialize which calls processEntries. The orderedDeserialize simply reads the longs one at a time and is less efficient (except for very small lists).

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
geekbeast: Yes. The idea was MetadataOption class is in API while Postgres...Definitions only exist server side. You'll have to transform what the client passes (MetadataOption) into Postgres...Definition classes to do this.

PR #355, opened 2018-11-01
Feature/lattice 1559 s3 blob service

pckeyes:
Tested with medialocal and medialocalaws profiles. Ready for review.


src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -102,14 +101,18 @@
 import org.apache.olingo.commons.api.edm.FullQualifiedName;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
```
geekbeast: No way this or Inject annotation can have much of an effect in this file.

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -766,30 +766,39 @@ public static AppType appType( ResultSet rs ) throws SQLException {
...
+            ByteBlobDataManager byteBlobDataManager) throws SQLException {
         final Map<UUID, Set<Object>> data = new HashMap<>();
 
         for ( PropertyType propertyType : authorizedPropertyTypes.values() ) {
             List<?> objects = propertyValue( rs, propertyType );
 
             if ( objects != null ) {
-                data.put( propertyType.getId(), new HashSet<>( objects ) );
+                if ( propertyType.getDatatype() == EdmPrimitiveTypeKind.Binary ) {
+                    data.put( propertyType.getId(), new HashSet<>( byteBlobDataManager.getObjects( objects ) ) );
```
geekbeast: This should be returning pre-signed urls not the raw objects themselves.

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -833,15 +845,18 @@ public static AppType appType( ResultSet rs ) throws SQLException {
...
         final Set<PropertyType> allPropertyTypes = authorizedPropertyTypes.values().stream()
                 .flatMap( propertyTypesOfEntitySet -> propertyTypesOfEntitySet.values().stream() )
                 .collect( Collectors.toSet() );
         for ( PropertyType propertyType : allPropertyTypes ) {
             List<?> objects = propertyValue( rs, propertyType );
 
             if ( objects != null ) {
-                data.putAll( propertyType.getType(), objects );
+                if ( propertyType.getDatatype() == EdmPrimitiveTypeKind.Binary ) {
+                    data.putAll( propertyType.getType(), byteBlobDataManager.getObjects( objects ) );
```
geekbeast: This should return pre-signed URLs and not the actual objects.

src/main/java/com/openlattice/datastore/configuration/DatastoreConfiguration.kt
```diff
@@ -0,0 +1,27 @@
...
+        @JsonProperty("secretAccessKey") val secretAccessKey: String): Configuration {
+    val key = SimpleConfigurationKey("datastore.yaml")
+
+    @JsonIgnore
+    override fun getKey(): ConfigurationKey {
+        return key
+    }
+
+    @JsonIgnore
+    fun key(): ConfigurationKey { return key }
```
geekbeast: This will NPE and you don't need the @JsonIgnore.  Check out LinkingConfiguration in indexer for the correct pattern.

src/test/kotlin/com/openlattice/data/LocalAwsBlobDataServiceTest.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    fun testPutAndGetObject() {
+        val data = ByteArray(10)
+        Random().nextBytes(data)
+        for (i in 1..3) {
+            key1 = key1.plus(UUID.randomUUID().toString())
+        }
+        key1 = key1.plus(data.hashCode())
```
geekbeast: You will have to use Hasher or check out the PostgresDataHasher class, which has reasonable implementations for the EdmPrimitiveTypeKinds.

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
pckeyes: It does. The objects stored that are being retrieved are the strings that are keys to the s3 bucket. The getObjects method will generate a presigned url for each key in that list.

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
pckeyes: Same as above

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
pckeyes: It does not, I'll go through and clean up the import statements

src/main/java/com/openlattice/datastore/configuration/DatastoreConfiguration.kt
pckeyes: Where/when will the NPE occur? I've been able to load all these values and use them when needed in the local aws and local blob data services.

pckeyes: I added the companion object like in LinkingConfiguration and it works for the bamboo test so I think we should be good on this one

src/test/kotlin/com/openlattice/data/LocalAwsBlobDataServiceTest.kt
pckeyes: I implemented this differently but am not sure it is what you are looking for. Also changed in PostgresEntityDataQueryService

PR #359, opened 2018-11-15
LATTICE-1624 fix toputilizers query for linking entity sets

you-leee:


src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -297,12 +300,52 @@ class Graph(private val hds: HikariDataSource, private val edm: EdmManager) : Gr
...
+                    authorizedFilteredRanking.filteredRanking.associationAggregations,
+                    ASSOC
+            ).map { it.value to authorizedFilteredRanking.associationPropertyTypes[it.key]!!.datatype } +
+                    buildAggregationColumnMap(
+                            index,
+                            authorizedFilteredRanking.entitySetPropertyTypes,
+                            authorizedFilteredRanking.filteredRanking.neighborTypeAggregations,
+                            ENTITY
+                    ).map { it.value to authorizedFilteredRanking.entitySetPropertyTypes[it.key]!!.datatype } +
+                    (associationCountColumnName(index) to EdmPrimitiveTypeKind.Int64 )+
```
geekbeast: What's up with the weird whitespace?

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -356,8 +382,16 @@ class Graph(private val hds: HikariDataSource, private val edm: EdmManager) : Gr
                 filteredRankings.mapIndexed { index, authorizedFilteredRanking ->
                     val tableSql = buildAssociationTable(index, entitySetIds, authorizedFilteredRanking, linked)
                     //We are guaranteed at least one association for a valid top utilizers request
+
                     if (index == 0) {
-                        "SELECT *,$scoreColumn FROM ($tableSql) as assoc_table$index "
+                        val selectedColumns =
+                                if( linked ) {
+                                    " *, UUID('${linkingEntitySetId.get()}') as $SELF_ENTITY_SET_ID, ${LINKING_ID.name} as $SELF_ENTITY_KEY_ID, $scoreColumn "
```
geekbeast: Shouldn't you alias linking id to SELF_ENTITY_KEY_ID?

you-leee: It is: ${LINKING_ID.name} as $SELF_ENTITY_KEY_ID

src/main/kotlin/com/openlattice/graph/Graph.kt
you-leee: What do you mean by that?

src/main/kotlin/com/openlattice/graph/Graph.kt
geekbeast: Ah, for some reason it got cut off by github. LGTM.

src/main/kotlin/com/openlattice/graph/Graph.kt
geekbeast: There is a space between Int64 and paranthesis but not so on the next line. There is also no space before the plus, but other plus signs have a space.

PR #370, opened 2018-11-30
improve authorization check perf by 1000x

katherinebernstein:


src/main/java/com/openlattice/authorization/HazelcastAuthorizationService.java
```diff
@@ -261,8 +261,9 @@ public boolean checkIfHasPermissions(
             AclKey key,
             Set<Principal> principals,
             EnumSet<Permission> requiredPermissions ) {
-        Set<Permission> permissions = getSecurableObjectPermissions( key, principals );
-        return permissions.containsAll( requiredPermissions );
+        return accessChecksForPrincipals( ImmutableSet.of( new AccessCheck( key, requiredPermissions ) ), principals )
```
geekbeast: This API has a lot of overhead from setting up maps for handling checking acl keys in bulk so its better to use AuthorizationAggregator if you want to take advantage of aggregator infrastructure.

geekbeast: Something likes this: 

```final var permissionsMap = new EnumMap<Permission, Boolean>( Permission.class );
        final var authzMap = ImmutableMap.of( key, permissionsMap );

        for ( Permission permission : requiredPermissions ) {
            permissionsMap.put( permission, false );
        }

        final EnumMap<Permission, Boolean> result =
                aces.aggregate( new AuthorizationAggregator( authzMap ), matches( authzMap.keySet(), principals ) )
                        .getPermissions()
                        .get( key );
        if ( result == null ) {
            return false;
        } else {
            return requiredPermissions.stream().allMatch( result::get );
        }```

geekbeast: Also through in an ```@Timed``` as this is used by several external components.

PR #372, opened 2018-12-07
Fix/lattice 1692 delete edges

pckeyes:


src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -616,6 +623,12 @@ private val INSERT_COLUMNS = setOf(
         VERSIONS
 ).map { it.name }.toSet()
 
+private val SET_ID_COLUMNS = setOf(
```
geekbeast: Kotlin default implementation is an ordered set so this will work, but could break at some point in the future as the interface makes no such guarantees. For this use case it makes more sense to use linkedSetOf or just a plain listOf as you don't need the set behavior.  In addition given your current usage setOf( ... ).joinToString(" OR ") { "${it.name} = ?" } will accomplish your goal and allow you to generate the clause more directly.

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -668,6 +681,8 @@ private val CLEAR_SQL = "UPDATE ${EDGES.name} SET version = ?, versions = versio
         "WHERE ${KEY_COLUMNS.joinToString(" = ? AND ")} = ? "
 private val DELETE_SQL = "DELETE FROM ${EDGES.name} WHERE ${KEY_COLUMNS.joinToString(" = ? AND ")} = ? "
 
+private val DELETE_BY_SET_SQL = "DELETE FROM ${EDGES.name} WHERE ${SET_ID_COLUMNS.joinToString(" = ? OR ")} = ? "
```
geekbeast: See previous comment.

src/main/kotlin/com/openlattice/graph/Graph.kt
pckeyes: My thought here was that the order doesn't actually matter because for each of those columns, I'm checking if the same entity set id is present in any of them. This is based on my understanding that if an edges contains the entity set as the src, dst, or edge and that entity set gets deleted then the edge doesn't exist anymore.

geekbeast: Good point. Should comment as that is pretty subtle to pick up in a cursory look without.

PR #373, opened 2018-12-11
implemented batch deletion for S3

pckeyes:


src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -417,22 +418,30 @@ class PostgresEntityDataQueryService(
     }
 
 
-    fun deletePropertyInEntitySetFromS3(propertyTable: String, fqn: String, fqnColumn: String, entitySetId: UUID) {
+    fun deletePropertiesInEntitySetFromS3(propertyTable: String, fqn: String, entitySetId: UUID) {
```
geekbeast: PostgresIterable -> asSequence -> chunked -> forEach should make this code way more concise.

PR #374, opened 2018-12-11
Lattice 1723 base 64 image api

pckeyes:


src/main/kotlin/com/openlattice/data/storage/LocalBlobDataService.kt
```diff
@@ -66,4 +67,12 @@ class LocalBlobDataService(private val hds: HikariDataSource) : ByteBlobDataMana
     fun selectEntitySql(s3Key: String): String {
         return "SELECT \"object\" FROM mock_s3_bucket WHERE key = '$s3Key'"
```
geekbeast: This seems super unsafe and is only okay, because the local blob data service is only used for local development and testing. Normally you would wrap this in a prepared statement that invokes a function, where the function safely executes the statement with safe quoting.

PR #375, opened 2018-12-12
Bugfix/multiplicity in entities

jokedurnez:
#### Input:

| ID | FirstName | LastName |
| --- | --- | --- |
| 123 | Ariana | Grande |
| 123 | Ari | Grande |

#### Expected output:

| nc.SubjectIdentification | nc.PersonGivenName | nc.PersonSurName |
| --- | --- | --- |
| 123 | [Ariana, Ari] | [Grande] |

#### Observed output:

| nc.SubjectIdentification | nc.PersonGivenName | nc.PersonSurName |
| --- | --- | --- |
| 123 | [Ari] | [Grande] |

## Detailed description of bug and fix:
The original data will be transformed into *entities* to pass to DataGraphService.integrateEntitiesAndAssociations() as:
```
entities:
    - key: 
        entitySetId: 0000
        entityId: 0000
      value:
        UUID(nc.PersonGivenName): "Ariana"
        UUID(nc.PersonSurName): "Grande"
        UUID(nc.SubjectIdentification): 123
    - key: 
        entitySetId: 0000
        entityId: 0000
      value:
        UUID(nc.PersonGivenName): "Ari"
        UUID(nc.PersonSurName): "Grande"
        UUID(nc.SubjectIdentification): 123
```

**Expected result**
```
entitiesToCreate:
    0000:
        UUID(nc.PersonGivenName): Set(Ariana, Ari)
        UUID(nc.PersonSurName): Set(Grande)
        UUID(nc.SubjectIdentification): Set(123)       
```

**What actually happens when looping over entities (depending on the order):**
Iteration 1:
```
entitiesToCreate:
    entityId:
        UUID(nc.PersonGivenName): Set(Ariana)
        UUID(nc.PersonSurName): Set(Grande)
        UUID(nc.SubjectIdentification): Set(123)       
```
Iteration 2:
```
entitiesToCreate:
    entityId:
        UUID(nc.PersonGivenName): Set(Ari) 
        UUID(nc.PersonSurName): Set(Grande)
        UUID(nc.SubjectIdentification): Set(123)       
```

In iteration 2, the original set is overwritten by `entitiesToCreate[entity.entityId] = entity.details`.  This fix checks whether the entityId already exists in the map, and creates a mergedEntity if so.

### DIFF
--- a/src/main/kotlin/com/openlattice/data/DataGraphService.kt
+++ b/src/main/kotlin/com/openlattice/data/DataGraphService.kt
@@ -339,14 +339,13 @@ open class DataGraphService(
             associations: Set<Association>,
             authorizedPropertiesByEntitySetId: Map<UUID, Map<UUID, PropertyType>>
     ): IntegrationResults? {
-        val entitiesByEntitySet = HashMap<UUID, MutableMap<String, Map<UUID, Set<Any>>>>()
-
+        val entitiesByEntitySet = HashMap<UUID, MutableMap<String, MutableMap<UUID, MutableSet<Any>>>>()
         for (entity in entities) {
             val entitiesToCreate = entitiesByEntitySet.getOrPut(entity.entitySetId) { mutableMapOf() }
-            val entityDetails = entitiesToCreate.getOrPut(entity.entityId) { entity.details }.toMutableMap()
+            val entityDetails = entitiesToCreate.getOrPut(entity.entityId) { entity.details }
             if(entityDetails !== entity.details) {
                 entity.details.forEach { propertyTypeId, values ->
-                    entityDetails.getOrPut(propertyTypeId){ mutableSetOf() }.toMutableSet().addAll( values )
+                    entityDetails.getOrPut(propertyTypeId){ mutableSetOf() }.addAll( values )
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


geekbeast:
Thanks that clarifies what is going wrong. Is the above pattern common and if so how does it arise? It would explain why Kim is seeing deadlocks on integrations. @pckeyes @kengie 

The system is definitely not expecting duplicate entities (or at least entities with identical entity keys) to be integrated (hence the use of Set instead of List) in a single call. This may also explain the conflicts in entity keys we were seeing that made no sense.

@jokedurnez Can you elaborate on how this scenario is arising during integrations?

jokedurnez:
This came up during the demo integration of @jason2e which was specifically designed to have a lot of multiplicity.  It does not cause deadlocks, it just silently passes the bug and misses data.
Shuttle integrates by 10K entities I think, so as soon as duplicated primary keys are present, this problem should arise (I'm guessing a lot).

jokedurnez:
FYI this is where shuttle sends, every 100K entities, the data:
https://github.com/openlattice/shuttle/blob/develop/src/main/java/com/openlattice/shuttle/Shuttle.java#L375

PR #384, opened 2018-12-21
persistent search alerts and indexing versions

katherinebernstein:


src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -747,10 +754,32 @@ public static AppType appType( ResultSet rs ) throws SQLException {
...
+            if ( objects != null ) {
+                if ( propertyType.getDatatype() == EdmPrimitiveTypeKind.Binary ) {
+                    data.put( propertyType.getId(), new HashSet<>( byteBlobDataManager.getObjects( objects ) ) );
+                } else {
+                    data.put( propertyType.getId(), new HashSet<>( objects ) );
+                }
+            }
+        }
+
+        data.put( ConductorElasticsearchApi.VERSION, ImmutableSet.of( lastWrite( rs ) ) );
```
geekbeast: I see. If you are doing this remember property types are random uuids, so you can have uuid constants that can't be generated or you can reserve make an acl key reservation that will never be usable for property types. This avoids the type issues. 

One more thing is that you used lastWrite as opposed to the actual version. Are you using last write instead of version because it's easier to index in elasticsearch?

src/main/kotlin/com/openlattice/search/renderers/AlprAlertEmailRenderer.kt
```diff
@@ -0,0 +1,188 @@
...
+            values["model"] = getFirstValue(vehicleRead, MODEL_FQN)
+            values["year"] = getFirstValue(vehicleRead, YEAR_FQN)
+            values["color"] = getFirstValue(vehicleRead, COLOR_FQN)
+            values["coordinates"] = getFirstValue(vehicleRead, LOCATION_COORDS_FQN)
+
+
+            return values.filterValues { it != null }
+        }
+
+        private fun extractVehicleImages(vehicleRead: SetMultimap<FullQualifiedName, Any>): List<ByteArrayAttachment> {
```
geekbeast: This function will work and probably can't be changed until content-type is set on upload by piper.

src/main/kotlin/com/openlattice/search/renderers/AlprAlertEmailRenderer.kt
```diff
@@ -0,0 +1,188 @@
...
+
+        private fun getImage(imageUrlPath: String?, fileType: String): ByteArrayOutputStream? {
+            if (imageUrlPath == null) {
+                return null
+            }
+
+            var stream: ByteArrayOutputStream? = null
+
+            try {
+                val imageUrl = URL(imageUrlPath)
```
geekbeast: I'm sure this is working, but if you have binary data loaded from s3 (say using the aws s3 java sdk) and content type I think you can avoid lots of overhead here. Even if you don't fix it now file a future JIRA ticket to come back and fix this.

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
katherinebernstein: so you mean just use like UUID(0, 0) as version / lastWrite?

I guess we could index all the version numbers instead of just the last write -- I think elasticsearch should be fine to still do range queries on a list of values.

src/main/java/com/openlattice/data/storage/HazelcastEntityDatastore.java
```diff
@@ -263,30 +244,46 @@ private void signalEntitySetDeleted( UUID entitySetId ) {
                 false ).stream();
     }
 
+    @Override
+    @Timed
+    public Stream<SetMultimap<FullQualifiedName, Object>> getEntitiesWithVersion(
```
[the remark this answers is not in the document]
katherinebernstein: the version is included in the entity you are getting (although last write would be more accurate than version)

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
geekbeast: Yes

geekbeast: I think using UUId(0,0) is more important than switching last write to bersion.

PR #390, opened 2019-01-19
Feature/lattice 1664 index linking entitysets

you-leee:


src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
```diff
@@ -437,10 +420,31 @@ internal fun buildFilterClause(fqn: String, filter: Set<Filter>): String {
...
 
 internal fun buildLinkingEntitiesClause(linkingIds:Set<UUID>): String {
-    return " AND ${LINKING_ID.name} IN ( ${linkingIds.joinToString(",") { "'$it'" }} )"
+    return " AND ${LINKING_ID.name} IN ( ${linkingIds.joinToString(",") { "'$it'" }} ) "
+}
+
+private fun getJoinColumns(linking: Boolean, omitEntitySetId: Boolean): List<String> {
+    return if (linking) {
+        if(omitEntitySetId) {
+            listOf(LINKING_ID.name)
```
geekbeast: When do we omit entity set id?

src/main/java/com/openlattice/edm/PostgresEdmManager.java
```diff
@@ -186,20 +188,54 @@ public void revoke(
         }
     }
 
-    public Set<UUID> getAllPropertyTypeIds() {
-        String getAllPropertyTypeIds = String.format("SELECT id from %1$s", PROPERTY_TYPES);
+    /**
+     * Duplicate of
+     * {@link com.openlattice.data.storage.PostgresEntityDataQueryService#getLinkingEntitySetIdsOfEntitySet(UUID)}
+     */
+    public Iterable<EntitySet> getAllLinkingEntitySetsForEntitySet( UUID entitySetId ) {
```
geekbeast: Use Hazelcast for this as it will be significantly faster and keep load of the database. If you add an index in the entity set mapstore for the linked entity sets field there is even support for indexing array types.

src/main/java/com/openlattice/data/storage/PostgresDataManager.java
```diff
@@ -1,106 +0,0 @@
...
-import java.sql.*;
-import java.util.Set;
-import java.util.UUID;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-public class PostgresDataManager {
```
geekbeast: Hopefully you didn't copy over this class as it contains bad code.

src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
you-leee: When we want the data merged by linking id: searching, getting a linking entity, getting the whole linking entity set.
This way we don't have to afterprocess and merge all properties by linking id

src/main/java/com/openlattice/edm/PostgresEdmManager.java
you-leee: Will do.

src/main/java/com/openlattice/edm/PostgresEdmManager.java
```diff
@@ -186,20 +198,40 @@ public void revoke(
...
     }
 
-    public Set<UUID> getAllPropertyTypeIds() {
-        String getAllPropertyTypeIds = String.format("SELECT id from %1$s", PROPERTY_TYPES);
+    /**
+     * Duplicate of
+     * {@link com.openlattice.data.storage.PostgresEntityDataQueryService#getLinkingEntitySetIdsOfEntitySet(UUID)}
+     */
+    public Set<EntitySet> getAllLinkingEntitySetsForEntitySet( UUID entitySetId ) {
+        return entitySets.values().stream()
```
geekbeast: This pulls all entity sets locally and then filters. Look at ```PermissionMapstore``` and ```HazelcastAuthorizationService``` for an example of how to use Hazelcast index to this quickly and efficiently. You'll want to override ```getMapConfig()``` so that it adds a map index config ```linkedEntitySets[any]``` that is not ordered. This will allow you to use ```entitySets.value(...)``` to do an indexed predicate query that will only retrieve the entity sets satisfying the query with the filtering done at all the nodes.

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -359,6 +359,10 @@ public static UUID entitySetId( ResultSet rs ) throws SQLException {
         return rs.getObject( ENTITY_SET_ID.getName(), UUID.class );
     }
 
+    public static Set<UUID> entitySetIds( ResultSet rs ) throws SQLException {
+        return Sets.newHashSet( (UUID[]) rs.getArray( ENTITY_SET_ID.getName() ).getArray() );
```
geekbeast: ```Sets.newHashSet( PostgresArrays.getUuidArray( rs, ENTITY_SET_ID.getName() ) )```

src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
geekbeast: Let me clarify, why do we omit instead of doing an array aggregate?

you-leee: You mean add an extra outer select?

geekbeast: If we're doing a group by linking id, we can also collect array_agg(entity_set_id) in the same clause. I think this is fine for now-- we may have to modify when we need to return the data source along with the entity.

src/main/java/com/openlattice/edm/PostgresEdmManager.java
```diff
@@ -186,20 +201,43 @@ public void revoke(
...
 
-    public Set<UUID> getAllPropertyTypeIds() {
-        String getAllPropertyTypeIds = String.format("SELECT id from %1$s", PROPERTY_TYPES);
+    /**
+     * Duplicate of
+     * {@link com.openlattice.data.storage.PostgresEntityDataQueryService#getLinkingEntitySetIdsOfEntitySet(UUID)}
+     */
+    @SuppressWarnings( "unchecked" )
+    public Set<EntitySet> getAllLinkingEntitySetsForEntitySet( UUID entitySetId ) {
+        return ImmutableSet.copyOf(
```
geekbeast: Almost. I'll just fix this.

PR #391, opened 2019-01-23
Lattice 1744 slow toputilizers neighbor filters

you-leee:


src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
```diff
@@ -77,8 +77,11 @@ fun selectEntitySetWithCurrentVersionOfPropertyTypes(
             .union(returnedPropertyTypes.map { propertyTypes[it]!! })
             .joinToString(",")
 
+    // for performance, we place inner joins of filtered property tables at the end
+    val orderedPropertyTypesByJoin = propertyTypes.toList()
```
geekbeast: I am not sure this should matter, but if it made it faster...

you-leee: It was also strange to me.. but it avoids an extra nested loop this way

PR #393, opened 2019-01-24
delete organization's app configs when deleting the org

katherinebernstein:


src/main/java/com/openlattice/organizations/HazelcastOrganizationService.java
```diff
@@ -252,6 +239,7 @@ public void destroyOrganization( UUID organizationId ) {
         securePrincipalsManager.deletePrincipal( aclKey );
         allMaps.stream().forEach( m -> m.delete( organizationId ) );
         reservations.release( organizationId );
+        appConfigs.removeAll( Predicates.equal( "__key#organizationId", organizationId ) );
```
geekbeast: Use the constant.

PR #396, opened 2019-02-02
Bugfix/lattice 1761 delete matched entries

you-leee:


geekbeast:
This will eventually live in weave microservice

PR #403, opened 2019-02-09
bulk load entity set property metadata for multiple entity sets

katherinebernstein:


src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -1584,6 +1584,15 @@ public EntityDataModelDiff getEntityDataModelDiff( EntityDataModel edm ) {
                         propertyTypeId -> getEntitySetPropertyMetadata( entitySetId, propertyTypeId ) ) );
     }
 
+    @Override
+    public Map<UUID, Map<UUID, EntitySetPropertyMetadata>> getAllEntitySetPropertyMetadataForIds( Set<UUID> entitySetIds ) {
```
geekbeast: First group by entity type id (entity type id -> list entity set ids). Use that to efficiently generate the default metadata for any properties that are missing it (you only have to load entity type once instead of repeatedly) and create a set of EntitySetPropertyMetadataKeys along the way. Finally do a getAll with all EntitySetPropertyMetadataKey. It would look something likes this: ```ImmutableMap.copyOf(
                Maps.transformValues(
                        entitySetPropertyMetadata.getAll( keys )
                                .entrySet()
                                .stream()
                                .collect( Collectors.groupingBy( e -> e.getKey().getEntitySetId() ) ),
                        value -> value.stream().collect(
                                Collectors.toMap( e -> e.getKey().getPropertyTypeId(), e -> e.getValue() ) ) ) );```

geekbeast: getAll is dramatically faster for bulkloading items vs parallel get calls to hazelcast.

katherinebernstein: How do you generate the default metadata for EntitySetPropertyMetadataKeys missing it without checking whether each value is contained? I had originally tested/timed both this version and a version doing getAll on entitySets and entityTypes to load properties and then loading/creating the metadata individually, but that ended up taking longer (although basically the same -- i think it was like 1.3 seconds vs 1.2 seconds) on 800ish entity sets, so I went with this one, but I can switch to the other. Just still not sure how to check which keys are missing metadata without checking individually -- can you somehow use predicates to check not in, or something?

PR #408, opened 2019-02-16
Feature/lattice 1824 organization integration credential

geekbeast:


src/main/kotlin/com/openlattice/assembler/Assembler.kt
```diff
@@ -72,17 +92,47 @@ class Assembler(
...
+        }
+    }
+
+    fun initializeMissingOrganizations() {
+        val currentOrganizations =
+                securableObjectTypes.keySet(
+                        Predicates.equal("this", SecurableObjectType.Organization)
+                ).map { it.first() }.toSet()
+
+        val initializedOrganizations = assemblies.keySet(Predicates.equal(INITIALIZED_INDEX, true))
```
you-leee: Why not just use false in the predicate?

PR #415, opened 2019-02-26
lots of things...

katherinebernstein:
- on data updates in postgres, return WriteEvents (with version) instead of just num entities updated
- implement entity set flags
- some auditing updates (initializer task, wiring setup, etc)


src/main/kotlin/com/openlattice/auditing/AuditRecordEntitySetsManager.kt
```diff
@@ -56,53 +60,106 @@ class AuditRecordEntitySetsManager(
...
-        if (aclKey.size == 1
-                && securableObjectTypes[aclKey] == SecurableObjectType.EntitySet
-                && auditRecordEntitySetConfigurations.keySet(Predicates.equal(ANY_AUDITING_ENTITY_SETS, aclKey.first())).isNotEmpty()) {
-            return aclKey.first()
+
+        if (edmAuditTypes.contains(eventType)) {
+            return getEdmAuditEntitySetId()
+        }
+
+        val aclKeyRoot = aclKey.first() // TODO do we always only care about the base id?
```
katherinebernstein: @katherinebernstein look into this

PR #417, opened 2019-02-28
Bugfix/lattice 1846 handle entityset schema changes

you-leee:
Changes:
- grant select in entity_set_id and id too on entity set materialized views, if user has read permission to at least 1 property
- grant select to edges to each user, who has read on any of the involved entitysets (will be fixed later as row level security)
- fix select edges when creating materialized view
- handle entity set created/property type added
- delete org database, when real organization is deleted


src/main/kotlin/com/openlattice/assembler/Assembler.kt
```diff
@@ -170,6 +191,31 @@ class Assembler(
...
     }
 
+    private fun createOrUpdateProductionForeignSchemaOfEntitySet(organizationId: UUID, entitySetId: UUID, propertyTypeAdded: Boolean) {
+        val dbname = PostgresDatabases.buildOrganizationDatabaseName(organizationId)
+        acm.connect(dbname).use { datasource ->
+            datasource.connection.use { conn ->
+                conn.createStatement().use { stmt ->
+                    val tableName = "\"$entitySetId\""
+                    // (re-)import table
+                    stmt.execute("DROP FOREIGN TABLE IF EXISTS $PRODUCTION_FOREIGN_SCHEMA.$tableName CASCADE")
```
geekbeast: This seems dangerous. You could wipe out customer created materialized views and tables if you cascade.

geekbeast: Also this should most likely happen in an entry processor to avoid multiple property changes from colliding. That is we need to serialize these changes being applied to an organization in some way.

you-leee: It will only need to drop and re-create it, if entity set is changed (property type is added). I can modify it, so, that it will only alter that table and add the new property column(s).

geekbeast: the worry is if you have mat view x and customer does creates mat view y as select * from x, cascade will remove y.

other change looks good.

you-leee: I already changed this code. Now I'm only dropping and re-importing public tables entity_sets, property_types and entity_types.

PR #419, opened 2019-03-05
Bugfix/lattice 1854 handle delete associations

you-leee:


src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -169,71 +176,197 @@ open class DataGraphService(
...
+                    it.key,
+                    it.value,
+                    authorizedPropertyTypesOfAssociations[it.key]!!)
+        }.sum()
+
+        logger.info("Deleted {} associations when deleting entities from entity set {}",
+                associationDeleteCount, entitySetId)
+    }
+
+    private fun collectAssociationsEntityKeyIds(
```
geekbeast: This is dangerous from a memory perspective. Highly linked objects will OOM the server. It possible re-work to use a Sequence and PostgresIterables.

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -300,6 +300,19 @@ class PostgresEntityDataQueryService(
         )
     }
 
+    fun getEntityKeyIdsInEntitySet(entitySetId: UUID): Set<UUID> {
```
geekbeast: Why convert to a set here? Just use distinct in sql query if you need ids to be distinct.

you-leee: They are distinct anyways, since this is only getting the ids of from entity_key_ids table. It needs to be a set later on in the call stack, when it is called from DataController

geekbeast: Use a collection with less overhead (or even a PostgresIterable)

geekbeast: That is the way you are using this it could return hundred of millions of UUIDs so it needs to be streaming.

PR #422, opened 2019-03-08
Bugfix/lattice 1854 handle delete associations

you-leee:
Moving access checks to controllers


src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -204,12 +200,33 @@ open class DataGraphService(
...
-        clearAssociations(entitySetId, Optional.of(entityKeyIds))
-
         return clearEntityDataAndVertices(entitySetId, entityKeyIds, authorizedPropertyTypes)
     }
 
+    override fun clearAssociationsBatch(
+            entitySetId: UUID,
+            associationsEdgeKeys: PostgresIterable<EdgeKey>,
+            authorizedPropertyTypes: Map<UUID, Map<UUID, PropertyType>>
+    ): Set<WriteEvent> {
```
geekbeast: Don't use a Set where a List will suffice.

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -289,78 +283,45 @@ open class DataGraphService(
...
-        } else {
-            getEdgeKeysOfEntitySet(entitySetId)
-        }
-        val associationsEntityKeyIds = HashMap<UUID, MutableSet<UUID>>()
-        associations.forEach {
-            val previousValue = associationsEntityKeyIds
-                    .putIfAbsent(it.edge.entitySetId, mutableSetOf(it.edge.entityKeyId))
-            if (previousValue != null) {
-                associationsEntityKeyIds[it.edge.entitySetId]!!.add(it.edge.entityKeyId)
+    override fun deleteAssociationsBatch(
```
geekbeast: Same as above, don't use set unless necessary.

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -289,78 +283,45 @@ open class DataGraphService(
...
+            entitySetId: UUID,
+            associationsEdgeKeys: PostgresIterable<EdgeKey>,
+            authorizedPropertyTypes: Map<UUID, Map<UUID, PropertyType>>): Set<WriteEvent> {
+        var associationDeleteCount = 0
+        val edgeKeysIterator = associationsEdgeKeys.iterator()
+        val writeEvents = HashSet<WriteEvent>()
+
+        while (edgeKeysIterator.hasNext()) {
```
geekbeast: I think you borrowed this pattern from linker. For various reasons I think you might benefit from doing ```associationEdgeKeys.asSequence().chunked(...)``` here to read a fixed number of items from the underlying iterable.

It's important to make sure the calling iterable has fetchSize set so it doesn't load the full thing into memory (i.e Postgres will stream data).

PR #423, opened 2019-03-08
Bugfix/lattice 1862 create auth0sync initialization task

you-leee:


you-leee:
They both need to override run(), so it seemed more clean to put them in different classes.

PR #429, opened 2019-03-12
Lattice 1867 propagate members to assemblies

you-leee:


src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -140,25 +140,43 @@ class AssemblerConnectionManager(
...
+            addMembersToOrganization(dbname, datasource, organization.members)
 
             createForeignServer(datasource)
 //                materializePropertyTypes(datasource)
 //                materialzieEntityTypes(datasource)
         }
     }
 
+    fun addMembersToOrganization(dbName: String, dataSource: HikariDataSource, members: Set<Principal>) {
+        members.filter { it.id != "openlatticeRole" && it.id != "admin" }
```
katherinebernstein: do we have constants for these strings somewhere?

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -140,25 +140,43 @@ class AssemblerConnectionManager(
...
+                .filter {
+                    securePrincipalsManager.principalExists(it)
```
katherinebernstein: can we add logging here to identify the bad ones?

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -140,25 +140,43 @@ class AssemblerConnectionManager(
...
+                            buildPostgresUsername(securablePrincipal)
+                    )
+                }
+    }
+
+    fun removeMembersFromOrganization(dbName: String, dataSource: HikariDataSource, members: Set<Principal>) {
+        members.filter { it.id != "openlatticeRole" && it.id != "admin" }
+                .filter {
+                    securePrincipalsManager.principalExists(it)
+                } //There are some bad principals in the member list some how-- probably from testing.
```
katherinebernstein: same comments as above

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -555,6 +565,18 @@ class AssemblerConnectionManager(
         }
     }
 
+    private fun revokeConnectAndSchemaUsage(datasource: HikariDataSource, dbname: String, userId: String) {
+        val dbUser = DataTables.quote(userId)
+        logger.info("Removing user {} from database {} and schema usage of $MATERIALIZED_VIEWS_SCHEMA", userId, dbname)
+
+        datasource.connection.use { conn ->
+            conn.createStatement().use { stmt ->
+                stmt.execute("REVOKE CREATE, CONNECT, TEMPORARY, TEMP ON DATABASE ${quote(dbname)} FROM $dbUser")
```
katherinebernstein: might be a good idea to store an array of permissions somewhere that we reference here and on granting, so that if we ever add/remove a permission we don't end up with extra floating permissions / only have to update in one place

src/main/kotlin/com/openlattice/assembler/processors/AddMembersToOrganizationAssemblyProcessor.kt
```diff
@@ -0,0 +1,70 @@
...
+    : AbstractRhizomeEntryProcessor<UUID, OrganizationAssembly, Void?>(), Offloadable {
+
+    @Transient
+    private var acm: AssemblerConnectionManager? = null
+
+    override fun process(entry: MutableMap.MutableEntry<UUID, OrganizationAssembly?>): Void? {
+        val organizationId = entry.key
+        val assembly = entry.value
+        if (assembly == null) {
+            logger.error("Encountered null assembly while trying to add new members $newMembers  to organization " +
```
katherinebernstein: is this expected whereas `acm == null` is not? Why does this one only log an error and the other null case throws an exception?

you-leee: In general: Acm needs to be initialized with init() func first to be able to use the processor. The assembly on the other hand can be null, if the organization(id) is not yet in the HazelcastMap.
In this case and in case of some other processors, it is maybe worth to throw an error, since we expect the organizationassembly to be already initialized, I followed the logic on the existing processors on this one.

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
you-leee: I only see admin in SystemRole

you-leee: Will move openLatticeRole to SystemRole too

src/main/java/com/openlattice/organizations/HazelcastOrganizationService.java
```diff
@@ -346,20 +324,13 @@ public void removeMembers( UUID organizationId, Set<Principal> members ) {
                         .filter( m -> m.getType().equals( PrincipalType.USER ) )
                         .map( securePrincipalsManager::lookup ) );
         membersOf.submitToKey( organizationId, new OrganizationMemberRemover( members ) );
```
geekbeast: Unrelated to this PR-- I think we need execute on key here... submitToKey returns a future and means the call can return before the actual removal is complete.

src/main/kotlin/com/openlattice/assembler/processors/AddMembersToOrganizationAssemblyProcessor.kt
geekbeast: Assembly can be null, but this case should never happen as when organization is created its assembly is initialized. It's probably fine to throw an exception here but is not a fatal error. 

The ACM not being initialized on the other hand is a fatal error and should cause things to explode.

src/main/kotlin/com/openlattice/assembler/processors/AddMembersToOrganizationAssemblyProcessor.kt
```diff
@@ -0,0 +1,70 @@
...
+                    "$organizationId.")
+        } else {
+            if(acm == null) {
+                throw IllegalStateException(NOT_INITIALIZED)
+            }
+            val dbName = PostgresDatabases.buildOrganizationDatabaseName(organizationId)
+            acm!!.connect(dbName).use { dataSource -> acm!!.addMembersToOrganization(dbName, dataSource, newMembers) }
```
geekbeast: This !! is unnecessary. Surprised the compiler doesn't flag this. Either put inside else so it does or use ?. syntax with ?: to throw the exception.

you-leee: It says than mutable property could have changed overtime

you-leee: "Smart cast to 'AssemblerConnectionManager' is impossible, because 'acm' is a mutable property that could have been changed by this time"

geekbeast: I see-- you could make acm a lateinit var which will cause an uninitialized exception if it is call before it is initialized. This is probably best since it will reveal any layer violations.

PR #431, opened 2019-03-15
LATTICE-1855 add function for getting acls in bulk by their names

you-leee:


geekbeast:
Change it to getAclKeyIds or something of the sort. We should at some point change the underlying maps and what not as well.

PR #436, opened 2019-03-21
Fix id generation to avoid duplicate entity ids per entity set

geekbeast:


src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
```diff
@@ -82,15 +88,53 @@ class PostgresEntityKeyIdService(
         return entityKeyIds
     }
 
-    override fun reserveEntityKeyIds( entityKeys: Set<EntityKey> ): Set<UUID> {
-        val entityIdsByEntitySet = entityKeys.groupBy ({ it.entitySetId },{it.entityId}).mapValues { it.value.toSet() }
+    private fun countUpEntityKeys(entityKeys: Set<EntityKey>) {
+        idRefCounts.executeOnKeys(entityKeys, IdRefCountIncrementer())
```
you-leee: Do we ever add entries to idRefCounts?

src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
```diff
@@ -82,15 +88,53 @@ class PostgresEntityKeyIdService(
...
+
+        return entityKeys.map { key ->
+            val elem = q.take()
+            val assignedId = idMap.putIfAbsent(key, elem)
+            return@map if (assignedId == null) {
+                val assignedPair = key to elem
+                storeEntityKeyIds(mapOf(assignedPair))
+                assignedPair
+            } else {
+                q.put(elem)
```
you-leee: Maybe I'm missing something here, but why do we put back the existing id into the queue?

geekbeast: At the time we generate an id we are not guaranteed that it will be assigned, because another node made assign it first (hence the put if absent). If that is not the assigned entity key id then it is returned to the pool of entity key from which keys can be drawn. 

This does make me realize there can be an issue as queue is fixed size and may block on accepting more elements.

src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
geekbeast: Entry processor can add entries through set value

you-leee: Yeah, but if there are no matching keys, will it still execute?

geekbeast: Yes.

you-leee: Ah, okay! Thanks for clarifying that!

PR #437, opened 2019-03-21
LATTICE-1873 add version > 0 to entity_key_ids table needing indexing…

you-leee:
…, needing linking, needing linked indexing and needing propagation indexes


katherinebernstein:
if we add this, we'll also need to separately load entity_key_ids in need of indexing with version < 0 in order to remove them from the elasticsearch indices, otherwise they'll still come back in search results even after being deleted.

geekbeast:
Well since we're going to be re-indexing all data soon... also we will have drop these indexes before we start indexer next and it will probably take several hours to recreate concurrently (similar to starting linker).

you-leee:
I'm not sure now, if I need to create a mechanic update for this or we will re-index, re-link and then re-linkindex every entity?

geekbeast:
You don't need to create a mechanic update. https://jira.openlattice.com/browse/LATTICE-1906

Thought I just realized that you need to add an index for items which have a version < 0 (unfortunately expression indices are super specific).

PR #440, opened 2019-03-26
Initial cut for cleaning up members from organizations on delete

makosblade:
This PR:
- Adds an entryprocessor that removes users from the PrincipalSet of every organization
- Adds a streamserializer for the above entryprocessor
- Updates the `OrganizationMembersMapstore` with an index on the values stored in that mapstore
- Updates Auth0SyncTask to use the above entryprocessor to remove users from organizations on user expiration from the `UserMapstore`


makosblade:
In hindsight, the Auth0Sync task user removal step looks like it expires users from the mapstore if they've previously been removed from Auth0. Doesn't HZ have something for auto-expiring mapstore entries?

geekbeast:
It has to mechanism for auto-expiring entries. Unfortunately, we can only expire after successfully completing a sync. If anything goes wrong with the sync then we wend up with no users in the mapstore.

PR #442, opened 2019-03-26
make sure removing/setting permissions won't leave aclKeys without an…

katherinebernstein:
…y owners of type USER


src/main/java/com/openlattice/authorization/HazelcastAuthorizationService.java
```diff
@@ -112,11 +111,30 @@ public void addPermission(
...
+                .and( hasAnyAclKeys( aclKeys ),
+                        hasPrincipalType( PrincipalType.USER ),
+                        hasExactPermissions( EnumSet.of( Permission.OWNER ) ) );
+
+        aces.entrySet( p ).stream().collect( Collectors
+                .groupingBy( entry -> entry.getKey().getAclKey(),
+                        Collectors.mapping( entry -> entry.getKey().getPrincipal(), Collectors.toSet() ) ) )
+                .entrySet().forEach( entry -> {
+            if ( Sets.difference( entry.getValue(), principals ).isEmpty() ) {
+                throw new IllegalArgumentException(
```
geekbeast: IllegalStateException seems accurate.

src/main/java/com/openlattice/authorization/HazelcastAuthorizationService.java
```diff
@@ -112,11 +111,30 @@ public void addPermission(
                 new PermissionMerger( permissions, securableObjectType, expirationDate ) );
     }
 
+    private void ensureAclKeysHaveOtherUserOwners( Set<AclKey> aclKeys, Set<Principal> principals ) {
```
geekbeast: This is a lot of overhead as it requires locally loading a bunch of data on every request. 

A more efficient way of implementing this is to use a fast aggregator (Aggregators.count()) with the predicate p below and another predicate that only checks for owner and aclKeys. If the counts are different than it is safe to delete and requires way less serialization and over head.

katherinebernstein: updated

PR #444, opened 2019-04-02
Update BULK_NEIGHBORHOOD_SQL to select edges by edge entity set ID an…

makosblade:
…d edge entity key ID instead of just selecting over src and dest set/key ids

This PR:
- fixes delete to correctly delete entities in association entity sets
- LATTICE-1895


src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -814,7 +816,8 @@ private val NEIGHBORHOOD_SQL = "SELECT * FROM ${EDGES.name} WHERE " +
 
 private val BULK_NEIGHBORHOOD_SQL = "SELECT * FROM ${EDGES.name} WHERE " +
         "(${SRC_ENTITY_SET_ID.name} = ? AND ${SRC_ENTITY_KEY_ID.name} IN (SELECT * FROM UNNEST( (?)::uuid[] ))) OR " +
-        "(${DST_ENTITY_SET_ID.name} = ? AND ${DST_ENTITY_KEY_ID.name} IN (SELECT * FROM UNNEST( (?)::uuid[] )))"
+        "(${DST_ENTITY_SET_ID.name} = ? AND ${DST_ENTITY_KEY_ID.name} IN (SELECT * FROM UNNEST( (?)::uuid[] ))) OR " +
```
geekbeast: Please replace ``` IN (SELECT ...)``` with ```= ANY(?)``` as it is more efficient (try doing an EXPLAIN to see why). The short version is that IN queries are internally re-written to = ANY queries and this is adding an additional unnest operation to the mix.

makosblade: Will do for now. 
I was hoping to do a separate PR of this form to get rid of all of the INs:
```
"SELECT * FROM ${EDGES.name} WHERE " +
        "INNER JOIN ( VALUES " +
            "(?::uuid),(?),(?),(?),(?),(?),(?),(?),(?),(?)" + // This is the only part I haven't finished
        ") vals(v) ON (" +
        "(${SRC_ENTITY_SET_ID.name} = ? AND ${SRC_ENTITY_KEY_ID.name} = v) OR " +
        "(${DST_ENTITY_SET_ID.name} = ? AND ${DST_ENTITY_KEY_ID.name} = v) OR " +
        "(${EDGE_ENTITY_SET_ID.name} = ? AND ${EDGE_ENTITY_KEY_ID.name} = v))"
```

makosblade:
Depends on build passing after merging https://github.com/openlattice/conductor-client/pull/443

src/main/kotlin/com/openlattice/graph/Graph.kt
geekbeast: Cool this would be nice! The other thing that might be worth while is automating the creation of temporary tables for joins likes this so it can be re-used in many places (not sure though).

makosblade: Agreed. My impression is that doing the above is the most efficient case for any size under ~1000 values(I think) and that above ~1000 or so, temp tables are superior.

PR #446, opened 2019-04-10
better logging for datatype NPE

UnsungHero97:
Related to https://github.com/openlattice/api/pull/306

### DIFF
--- a/src/main/java/com/openlattice/postgres/JsonDeserializer.java
+++ b/src/main/java/com/openlattice/postgres/JsonDeserializer.java
@@ -76,8 +76,14 @@ public class JsonDeserializer {
             EdmPrimitiveTypeKind dataType,
             UUID propertyTypeId,
             Object value ) {
+
         if ( value == null ) {
             return null;
+        }
+
+        if ( dataType == null ) {
+            logger.error( "received a null datatype for property type {}", propertyTypeId.toString() );
+            throw new NullPointerException( propertyTypeId.toString() );
         }
 
         switch ( dataType ) {


src/main/java/com/openlattice/postgres/JsonDeserializer.java
```diff
@@ -76,10 +76,16 @@ private static Object validateFormatAndNormalize(
             EdmPrimitiveTypeKind dataType,
             UUID propertyTypeId,
             Object value ) {
+
         if ( value == null ) {
             return null;
         }
 
+        if ( dataType == null ) {
```
katherinebernstein: are you seeing this happen? this shouldn't be possible...

you-leee: There is no check for datatype in PropertyType constructor though, so I think it could happen

katherinebernstein: the column in postgres is not null though, so it should fail the write

you-leee: ah, yes, you are totally right!

UnsungHero97: We run into this locally every once in a while. Yes, the write fails with NPE. I'm just adding logging to make it easier to debug.

geekbeast: I think it makes more sense to add this check in the constructor then to prevent a property type with a null data type?

UnsungHero97: I agree. I'll add a check in the constructor as well. However, I tried it locally, and logging only in the constructor is not enough. Logging in JsonDeserializer seems to be necessary for this particular edge case.

UnsungHero97: https://github.com/openlattice/api/pull/306

katherinebernstein: so the issue isn't property types not having a datatype -- it's when someone attempts to write data to a property type that either doesn't exist on an entity set or isn't authorized for an entity set (since these values come from a Map<UUID, EdmPrimitiveTypeKind> of authorized property types)

PR #447, opened 2019-04-11
collect entity types to list when deleting property type

katherinebernstein:


src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -239,9 +239,9 @@ public void deletePropertyType( UUID propertyTypeId ) {
 
     @Override
     public void forceDeletePropertyType( UUID propertyTypeId ) {
-        Stream<EntityType> entityTypes = entityTypeManager
+        List<EntityType> entityTypes = entityTypeManager
```
geekbeast: If this is important enough... do predicates query on the EntityType Map with properties[any] index to retrieve a collection directly.

geekbeast:
Fixed in #449

PR #452, opened 2019-04-22
fix reading out of iterator

geekbeast:


src/main/java/com/openlattice/ids/IdCatchupEntryProcessor.java
```diff
@@ -77,6 +77,8 @@ public PreparedStatement prepareExistQuery( Connection connection ) throws SQLEx
 
     public boolean exists( PreparedStatement ps, UUID id ) throws SQLException {
         ps.setObject( 1, id );
+        //Count query always guaranteed to have one row.
+        ps.executeQuery().next();
         return ResultSetAdapters.count( ps.executeQuery() ) > 0;
```
you-leee: This will produce a new ResultSet, which will still throw error. 
        final var rs = ps.executeQuery();
        rs.next();
        return ResultSetAdapters.count( rs ) > 0;

geekbeast: nice catch

PR #457, opened 2019-04-27
Feature/lattice 1929 unify permission handling linking entitysets

you-leee:


src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
```diff
@@ -137,7 +221,7 @@ public EdmAuthorizationHelper( EdmManager edm, AuthorizationManager authz ) {
             EnumSet<Permission> requiredPermissions,
             Set<Principal> principals ) {
         return ( entitySetIds.isEmpty() )
-                ? ImmutableMap.of()
+                ? Maps.newHashMap()
```
geekbeast: Does the map have to be mutable?

src/main/java/com/openlattice/search/SearchService.java
```diff
@@ -253,7 +244,7 @@ public SearchResult executeOrganizationKeywordSearch( SearchTerm searchTerm ) {
         Set<AclKey> authorizedOrganizationIds = authorizations
                 .getAuthorizedObjectsOfType( Principals.getCurrentPrincipals(),
                         SecurableObjectType.Organization,
-                        EnumSet.of( Permission.READ ) ).collect( Collectors.toSet() );
+                        READ_PERMISSION ).collect( Collectors.toSet() );
```
geekbeast: Good idea.

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
```diff
@@ -112,15 +112,99 @@ public EdmAuthorizationHelper( EdmManager edm, AuthorizationManager authz ) {
...
+                linkingEntitySet, requiredPermissions, Principals.getCurrentPrincipals() );
+    }
+
+    public Map<UUID, PropertyType> getAuthorizedPropertyTypesOfLinkingEntitySet(
+            EntitySet linkingEntitySet,
+            EnumSet<Permission> requiredPermissions,
+            Set<Principal> principals ) {
+        if ( linkingEntitySet.getLinkedEntitySets().isEmpty() ) {
+            return Maps.newHashMap();
+        } else {
```
katherinebernstein: this doesn't need to be in an else

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
```diff
@@ -112,15 +112,99 @@ public EdmAuthorizationHelper( EdmManager edm, AuthorizationManager authz ) {
...
+            final var authorizedPropertyTypesOfNormalEntitySets = getAuthorizedPropertiesOnEntitySets(
+                    linkingEntitySet.getLinkedEntitySets(), requiredPermissions, principals );
+
+            return authorizedPropertyTypesOfNormalEntitySets
+                    .values().stream()
+                    .reduce( ( properties, nextProperties ) ->
+                            Maps.difference( properties, nextProperties ).entriesInCommon() )
```
katherinebernstein: Does this account for partial permission matches? i.e. if a linking entity set has 2 normal entity sets (es1, es2) and you have READ and WRITE on [es1, pt1], and only READ on [es2, pt1], will this set difference match and filter permissions to READ only?

src/main/java/com/openlattice/search/SearchService.java
```diff
@@ -480,18 +471,20 @@ public SearchResult executeFQNPropertyTypeSearch( String namespace, String name,
...
+            entityKeyIdsByLinkingId.values().forEach( entityKeyIds::addAll );
+            entityKeyIds.removeAll( entityKeyIdsByLinkingId.keySet() ); // remove linking ids
+
+            // normal entity sets within 1 linking entity set are only authorized if all of them is authorized
+            var authorizedLinkedEntitySetIds = linkingEntitySets.stream()
+                    .map( EntitySet::getLinkedEntitySets )
+                    .filter( esIds -> esIds.stream()
+                            .allMatch( esId -> authorizations
+                                    .checkIfHasPermissions( new AclKey( esId ), principals, READ_PERMISSION ) ) )
+                    .flatMap( Set::stream )
```
katherinebernstein: is this supposed to return the original linked entity set id or its normal entity set ids?

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
you-leee: Well, yes, because when collecting authorized properties on multiple entity sets, we collect it for linking entity sets and normal entity sets separately, but after representing it as just 1 map. I think it's faster to just put all elements into the first map, rather to create a new map with the copy of the first map and do the same things after.

src/main/java/com/openlattice/search/SearchService.java
you-leee: The normal entity set ids, because they are stored in edges table what we query after.

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
```diff
@@ -112,15 +112,99 @@ public EdmAuthorizationHelper( EdmManager edm, AuthorizationManager authz ) {
     }
 
     /**
-     * @see EdmAuthorizationHelper#getAuthorizedPropertiesOnEntitySets
+     * @see EdmAuthorizationHelper#getAuthorizedPropertiesByNormalEntitySets(EntitySet, EnumSet, Set)
      */
-    public Map<UUID, Map<UUID, PropertyType>> getAuthorizedPropertiesOnEntitySets(
-            Set<UUID> entitySetIds,
+    public Map<UUID, Map<UUID, PropertyType>> getAuthorizedPropertiesByNormalEntitySets(
```
geekbeast: Maybe we should call them UnitaryEntitySets?

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
geekbeast: Yeah, I would change to isNotEmpty() and then just have a default return as the cleaner approach.

src/main/java/com/openlattice/search/SearchService.java
katherinebernstein: ok makes sense -- could you rename this to `authorizedNormalEntitySetIds` or something then? Name is a bit misleading (although I guess this carries over from the previous terminology we were using where linked/linking entity set was used pretty much interchangeably)

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
geekbeast: Good catch!

you-leee: Nope, it just returns property types for which the user has the required permissions on all normal entity sets. The example what you described can be achieved with 2 separate calls, 1 for checking WRITE and 1 for READ.

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
geekbeast: Makes sense unless the algorithm changes, which I think it should. See general comment.

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
you-leee: Ah, yes, thanks!

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
katherinebernstein: ah true -- I was thinking of the get permissions calls in HazelcastAuthorizationService (btw is that set up for the new linked entity set permissions model?)

src/main/java/com/openlattice/search/SearchService.java
you-leee: Yes, I just left the name as it is, but will change it.

src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
you-leee: We already started to call them normal entity sets, so unless there is a strong reason, I wouldn't change it again. (it's just getting too confusing)

src/main/kotlin/com/openlattice/authorization/aggregators/AuthorizationSetAggregator.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.openlattice.authorization.Permission
+import java.util.*
+import java.util.Map
+
+data class AuthorizationSetAggregator(
+        val permissionsMap: MutableMap<AclKey, EnumSet<Permission>>
+) : Aggregator<Map.Entry<AceKey, AceValue>, EnumSet<Permission>>() {
+
+    override fun accumulate(input: Map.Entry<AceKey, AceValue>) {
+        val permissions = checkNotNull(input.value.permissions, "Permissions shouldn't be null")
```
geekbeast: Avoid exceptions if you can safely in an aggregator.

src/main/kotlin/com/openlattice/authorization/aggregators/AuthorizationSetAggregator.kt
```diff
@@ -0,0 +1,69 @@
...
+                    val reducedPermissions = Sets.intersection(acc, permissionSet)
+                    if (reducedPermissions.isEmpty()) {
+                        return@fold EnumSet.noneOf(Permission::class.java)
+                    }
+
+                    EnumSet.copyOf(reducedPermissions)
+                }
+    }
+
+    override fun toString(): String {
```
geekbeast: Kotlin provides a useful toString implementation for data class. What is the reason for overriding it?

src/main/kotlin/com/openlattice/authorization/aggregators/AuthorizationSetAggregator.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    override fun aggregate(): EnumSet<Permission> {
+        if (permissionsMap.isEmpty()) {
+            return EnumSet.noneOf(Permission::class.java)
+        }
+
+        return permissionsMap.values
+                .fold(EnumSet.allOf(Permission::class.java)) { acc, permissionSet ->
+                    val reducedPermissions = Sets.intersection(acc, permissionSet)
```
geekbeast: Kotlin sets have a built in intersection call. This is preferable to using the Sets.intersection call which returns a dynamically computed view.

src/main/kotlin/com/openlattice/authorization/aggregators/AuthorizationSetAggregator.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    override fun accumulate(input: Map.Entry<AceKey, AceValue>) {
+        val permissions = checkNotNull(input.value.permissions, "Permissions shouldn't be null")
+        permissionsMap[input.key.aclKey] = permissions
+    }
+
+    override fun combine(aggregator: Aggregator<*, *>?) {
+        if (aggregator is AuthorizationSetAggregator) {
+            aggregator.permissionsMap.forEach {
+                permissionsMap[it.key]!!.addAll(it.value)
```
geekbeast: permissionsMap.getValue(it.key).addAll(it.value)

src/main/kotlin/com/openlattice/authorization/aggregators/AuthorizationSetAggregator.kt
you-leee: Oh, it's a leftover from first implementation I did

src/main/kotlin/com/openlattice/authorization/aggregators/AuthorizationSetAggregator.kt
you-leee: Weird, that it did not give me a warning

PR #461, opened 2019-04-30
bring back lastWrite in search responses, and return lastReadDateTime…

katherinebernstein:
… from PersistentSearchMessengerTask where present


src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
```diff
@@ -375,11 +378,23 @@ internal fun selectEntityKeyIdsWithCurrentVersionSubquerySql(
...
+                selectColumns += ", max(${LAST_WRITE.name}) AS ${LAST_WRITE.name}"
+            }
+        } else {
+            selectColumns += ", $metadataColumns"
+        }
+    }
+
+    val groupBy = if (linking) "GROUP BY ${LINKING_ID.name}" else ""
+
+    return "(SELECT $selectColumns FROM ${IDS.name} WHERE ${VERSION.name} > 0 $entitiesClause $groupBy ) as $ENTITIES_TABLE_ALIAS"
```
you-leee: distinct is missing from the select, if query is for linking entities

katherinebernstein: it's not needed anymore because of the groupBy

you-leee: I see, nvm then!

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -91,7 +102,7 @@ class PostgresEntityDataQueryService(
         }
         return streamableEntitySet(
                 linkingIdsByEntitySetId, authorizedPropertyTypesByEntitySetId,
-                EnumSet.noneOf(MetadataOption::class.java), Optional.empty(), adapter, true
+                metadataOptions, Optional.empty(), adapter, true
```
you-leee: I think this needs a different adapter like implicitEntityValuesByIdWithLastWrite, when metadatoptions are not empty. And since linking entityset authorization changed, you can pass any of the authorized property maps from authorizedPropertyTypesByEntitySetId

katherinebernstein: oooh good call

you-leee: I think it should be only used, when metadatoptions is not empty, because we call getLinkedEntityDataWithMetadata with empty metadataOptions too, but implicitEntityValuesByIdWithLastWrite expects the last_write column to be there

katherinebernstein: +1 you right

PR #463, opened 2019-05-01
Handle PostgresIndexType updates in UpdatePropertyTypeMetadataProcessor

makosblade:
This PR:
- PRocesses indextype updates in the UpdatePropertyTypeMetadataProcessor

### DIFF
--- a/src/main/java/com/openlattice/edm/types/processors/UpdatePropertyTypeMetadataProcessor.java
+++ b/src/main/java/com/openlattice/edm/types/processors/UpdatePropertyTypeMetadataProcessor.java
@@ -20,14 +20,13 @@
 
 package com.openlattice.edm.types.processors;
 
-import java.util.Map.Entry;
-import java.util.UUID;
-
+import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor;
 import com.openlattice.edm.requests.MetadataUpdate;
 import com.openlattice.edm.type.PropertyType;
-import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor;
-
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+import java.util.Map.Entry;
+import java.util.UUID;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/openlattice/edm/types/processors/UpdatePropertyTypeMetadataProcessor.java
```diff
@@ -56,6 +55,9 @@ public Object process( Entry<UUID, PropertyType> entry ) {
             if ( update.getPii().isPresent() ) {
                 pt.setPii( update.getPii().get() );
             }
+            if ( update.getIndexType().isPresent() ) {
+                pt.setPostgresIndexType( update.getIndexType().get() );
```
katherinebernstein: is there logic somewhere to actually update the index type in postgres?

makosblade: Uhh now that you mention it, I'm not really seeing anything to do that. The update logic seems pretty constrained

katherinebernstein: I think we should probably include that piece with the ability to change the indexType field on property types, otherwise we could end up in a weird state where the indexType field doesn't actually reflect the index type in postgres

katherinebernstein: although actually we might already be in that state...

makosblade: 👍  will do!

makosblade: Actually let's scratch that and make a ticket for it

makosblade: https://jira.openlattice.com/browse/LATTICE-1983

makosblade:
To be completed here:
https://jira.openlattice.com/browse/LATTICE-1983

PR #465, opened 2019-05-02
Bugfix/lattice 1976 edges update deadlock

you-leee:
Made some minor improvements on how we delete the edges of associations and in graph queries in general. 
Will keep the issue in review until we find out more on how the deadlock occurred, if that's alright with everyone.


src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -179,33 +179,47 @@ open class DataGraphService(
...
 
-        logger.info("Cleared {} associations when deleting entities from entity set {}", associationDeleteCount,
+        logger.info("Cleared {} associations when deleting entities from entity set {}", associationClearCount,
                 entitySetId)
 
         return writeEvents
     }
 
-    private fun clearEntityDataAndVertices(entitySetId: UUID,
+    private fun clearEntityDataAndVerticesOfAssociayions(entitySetId: UUID,
```
makosblade: Typo!

you-leee: Oh, thanks!

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -279,6 +294,21 @@ open class DataGraphService(
         return writeEvents
     }
 
+    private fun deleteEntityDataAnVerticesOfAssociations(
```
katherinebernstein: typo (And)

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -785,43 +810,52 @@ internal fun getTopUtilizersFromDst(entitySetId: UUID, filters: SetMultimap<UUID
...
         "WHERE ${KEY_COLUMNS.joinToString(" = ? AND ")} = ? "
+
 private val CLEAR_BY_SET_SQL = "UPDATE ${EDGES.name} SET version = ?, versions = versions || ? " +
         "WHERE ${SET_ID_COLUMNS.joinToString(" = ? OR ")} = ? "
+
 private val CLEAR_BY_VERTICES_SQL = "UPDATE ${EDGES.name} SET version = ?, versions = versions || ? WHERE " +
-        "(${SRC_ENTITY_SET_ID.name} = ? AND ${SRC_ENTITY_KEY_ID.name} IN (SELECT * FROM UNNEST( (?)::uuid[] ))) OR " +
-        "(${DST_ENTITY_SET_ID.name} = ? AND ${DST_ENTITY_KEY_ID.name} IN (SELECT * FROM UNNEST( (?)::uuid[] ))) OR " +
-        "(${EDGE_ENTITY_SET_ID.name} = ? AND ${EDGE_ENTITY_KEY_ID.name} IN (SELECT * FROM UNNEST( (?)::uuid[] )))"
+        "( ${SRC_ENTITY_SET_ID.name} = ? AND ${SRC_ENTITY_KEY_ID.name} = ANY( (?)::uuid[] ) ) OR " +
```
katherinebernstein: This can be simplified as `... = ANY( ? )` (same thing for instances below)

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -785,43 +810,52 @@ internal fun getTopUtilizersFromDst(entitySetId: UUID, filters: SetMultimap<UUID
...
-        "(${DST_ENTITY_SET_ID.name} = ? AND ${DST_ENTITY_KEY_ID.name} = ?)"
-
-private val BULK_NEIGHBORHOOD_SQL = "SELECT * FROM ${EDGES.name} WHERE " +
-        "(${SRC_ENTITY_SET_ID.name} = ? AND ${SRC_ENTITY_KEY_ID.name} = ANY( (?)::uuid[] )) OR " +
-        "(${DST_ENTITY_SET_ID.name} = ? AND ${DST_ENTITY_KEY_ID.name} = ANY( (?)::uuid[] )) OR " +
-        "(${EDGE_ENTITY_SET_ID.name} = ? AND ${EDGE_ENTITY_KEY_ID.name} = ANY( (?)::uuid[] ))"
+        "( ${SRC_ENTITY_SET_ID.name} = ? AND ${SRC_ENTITY_KEY_ID.name} = ? ) OR " +
+        "( ${DST_ENTITY_SET_ID.name} = ? AND ${DST_ENTITY_KEY_ID.name} = ? )"
 
 private val BULK_BULK_NEIGHBORHOOD_SQL = "SELECT * FROM ${EDGES.name} WHERE " +
```
katherinebernstein: since `BULK_NEIGHBORHOOD_SQL` is gone, can we rename this one to that instead of BULK_BULK_... ?

src/main/kotlin/com/openlattice/graph/Graph.kt
you-leee: Thanks, will do!

src/main/kotlin/com/openlattice/graph/Graph.kt
you-leee: Well, I'm all in to rename it, but the naming follows a pattern and 1 BULK in the name infers, that we want to match multiple entity key ids and 2 BULKs that we want to match also multiple entity set ids.

katherinebernstein: are there any other BULK_BULKs? I'd also be up for changing it to `ENTITY_SET_NEIGHBORHOOD_SQL` or something that makes what it does a little clearer

katherinebernstein: or `BULK_ENTITY_SET_NEIGHBORHOOD_SQL` since it takes multiple entity sets. I think `BULK` should imply that an action is performed on multiple values, and in most cases entityKeyIds is implied unless otherwise specified

you-leee: Yeah, BULK_ENTITY_SET_NEIGHBORHOOD_SQL sound better! I will change it to that

geekbeast:
I'm assuming we'll have a separate PR for fixing the deadlock.

you-leee:
Actually I already started the change on this branch

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -141,134 +137,254 @@ class Graph(private val hds: HikariDataSource, private val edm: EdmManager) : Gr
...
-            ps.setObject(1, arr)
-            ps.setObject(2, arr)
-            ps.setObject(3, arr)
-            return ps.executeUpdate()
+        hds.connection.use { connection ->
+            connection.autoCommit = false
+
+            val arr = PostgresArrays.createUuidArray(connection, vertices)
+
+            val psLock = connection.prepareStatement(LOCK_BY_VERTICES_SQL)
```
geekbeast: This SQL query doesn't use ```= ANY(?)```. Are you sure this will work?

PR #466, opened 2019-05-02
Feature/lattice 1929 unify permission handling linking entitysets

you-leee:
1. New function in EdmAutorizationHelper to get authorized properties of a mix of normal and linking entity sets
2. Fixed AuthorizationSetAggregator to accumulate permissions for matched principals for each aclkey
3. Update view in openlattice when normal entity set is added/removed to/from linking entity set


src/main/java/com/openlattice/authorization/EdmAuthorizationHelper.java
```diff
@@ -102,10 +104,10 @@ public EdmAuthorizationHelper( EdmManager edm, AuthorizationManager authz ) {
                                 principals ) ) );
     }
 
-    public Set<UUID> getAuthorizedPropertiesOnEntitySet(
+    public Set<UUID> getAuthorizedPropertiesOnNormalEntitySet(
```
katherinebernstein: I thought it would be useful to have a version of this that can take either a normal or a linked entity set and return its set of authorized property types so that whenever this is needed, you don't need to special case linked entity sets vs normal ones (basically the same way `getAuthorizedPropertiesOnNormalEntitySets` has a parallel that's entity-set-type-agnostic)

you-leee: Good idea, I will start to refactor the calls to a general one

geekbeast: :+1:

PR #468, opened 2019-05-08
Bugfix/lattice 1990 src dst edge type check

you-leee:


src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -465,6 +475,35 @@ open class DataGraphService(
         return null
     }
 
+    private fun checkAssociationEntityTypes(associations: Set<DataEdgeKey>) {
+        associations.forEach {
```
geekbeast: Looping over the days keys twice seems unnecessarily expensive. On top of that the API calls we are making do not use a projection yet (that's a separate ticket) so they loading they entire entity set over and over again over the network (quite expensive). Accumulate a Map<UUID, Map<UUID, Set<UUID>>> (edge entity set id -> src entity set id -> dst entity set ids) then load just the distinct entity sets, grab their entity types like you are currently doing, and then follow your current approach except that you are only evaluating distinct entity set combinations (which will tend to be much smaller) then total items submitted.

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -375,8 +380,13 @@ open class DataGraphService(
                 .asMap(associations)
                 .forEach {
                     val entitySetId = it.key
-                    val entities = it.value.map { it.data }
-                    val (ids, entityWrite) = createEntities(entitySetId, entities, authorizedPropertiesByEntitySetId[entitySetId]!!)
+
+                    // check entity types of associations before creation
+                    checkAssociationEntityTypes(entitySetId, it.value)
```
geekbeast: This is better but still loads the whole AssociationTypes one at a time. You could add a ```getAssociationTypes( Set<UUID> associationTypeIds )``` to EdmManager and then call it ```getAssociationTypes(associations.keySet())```. Then make checkAssociationEntityTypes take the association type instead of the entitySetId.

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -391,45 +401,61 @@ open class DataGraphService(
...
+                        if(associationEntitySetIds.getValue(edgeEsId)
+                                        .putIfAbsent(srcEsId, mutableSetOf(dstEsId)) != null) {
+                            associationEntitySetIds.getValue(edgeEsId).getValue(srcEsId).add(dstEsId)
+                        }
+                    }
+
+                    DataEdgeKey(src, dst, edge)
+                }
+                .toSet()
+        checkAssociationEntityTypes(associationEntitySetIds)
```
geekbeast: Similar to above you could make a call getAssociationEntityTypes and pass in that map to streamline look ups.

PR #470, opened 2019-05-11
Fix startup issues.

geekbeast:


src/main/kotlin/com/openlattice/assembler/Assembler.kt
```diff
@@ -325,15 +325,15 @@ class Assembler(
...
+                entitySetIds.associate { it to Optional.empty<Set<UUID>>() },
                 propertyFqns,
                 authorizedPropertyTypes.values.map(PropertyType::getId),
                 mapOf(entitySetId to authorizedPropertyTypes.keys),
                 mapOf(),
                 EnumSet.allOf(MetadataOption::class.java),
                 authorizedPropertyTypes.mapValues { it.value.datatype == EdmPrimitiveTypeKind.Binary },
                 entitySet.isLinking,
-                false //Always provide entity set id
+                entitySet.isLinking //false //Always provide entity set id
```
you-leee: I think this will break materialize in AssemblerConnectionManager, because we select entity_set_id and linking_id for materialized view and with this, omitEntitySetId will always be true for linking entitysets.

geekbeast: Yup. You'll have to modify how it's done. @katherinebernstein made a change that does a group by during linking entity sets to fix indexing, which she only tested with omitEntitySetId = true for linked entity sets.

src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
```diff
@@ -378,8 +381,13 @@ internal fun selectEntityKeyIdsWithCurrentVersionSubquerySql(
 ): String {
     val metadataColumns = getMetadataOptions(metadataOptions, linking).joinToString(",")
 
-
-    var selectColumns = joinColumns.joinToString(",")
+    var selectColumns = joinColumns.joinToString(",") { column ->
```
you-leee: Why not include entity_set_id in group by?

geekbeast: linking_id crosses multiple entity_set_ids.

PR #471, opened 2019-05-11
Group by entity_set_id and linking_id

geekbeast:


src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
```diff
@@ -377,33 +378,43 @@ internal fun selectEntityKeyIdsWithCurrentVersionSubquerySql(
...
+                if (linking) {
+                    if (metadataOptions.contains(MetadataOption.LAST_WRITE)) {
+                        ", max(${LAST_WRITE.name}) AS ${LAST_WRITE.name}"
+                    } else {
+                        ""
+                    } + if (metadataOptions.contains(MetadataOption.ENTITY_SET_IDS)) {
+                        "array_agg(entity_set_id) as entity_set_ids"
+                    } else {
+                        ""
+                    } + if (metadataOptions.contains(MetadataOption.ENTITY_KEY_IDS)) {
```
katherinebernstein: will there ever be multiple of these metadata options passed? if so spacing might be off

katherinebernstein: /commas

PR #481, opened 2019-05-17
Feature/temp query table

geekbeast:


src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -282,12 +296,21 @@ class PostgresEntityDataQueryService(
...
+                            .executeQuery("DELETE FROM ${QUERIES.name} WHERE ${QUERY_ID.name} = $queryId")
+
+                    StatementHolder(connection, statement, rs, listOf(cleanupStatement), listOf(cleanupResultSet))
                 },
                 adapter
         )
     }
 
+    fun createQueryTable(entityKeyIds: Map<UUID, Optional<Set<UUID>>>) {
+
```
makosblade: Magic happens here?

geekbeast: Ended up going in a different direction :)

PR #484, opened 2019-05-17
fix authorized props for assembler view

geekbeast:


src/main/kotlin/com/openlattice/assembler/Assembler.kt
```diff
@@ -372,7 +372,7 @@ class Assembler(
                 entitySetIds.associate { it to Optional.empty<Set<UUID>>() },
                 propertyFqns,
                 authorizedPropertyTypes.values.map(PropertyType::getId),
-                mapOf(entitySetId to authorizedPropertyTypes.keys),
+                entitySetIds.map { it to authorizedPropertyTypes.keys) }.toMap(),
```
geekbeast: associate :)

PR #487, opened 2019-05-17
Perform filtering directly in with query

geekbeast:


src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
```diff
@@ -127,17 +127,17 @@ internal fun buildWithClause(queryId: UUID, entityKeyIds: Map<UUID, Optional<Set
...
     val linkingClause = if (linking) "AND ${LINKING_ID.name} IS NOT NULL" else ""
 
-    val emptyEntitySetIds = entityKeyIds.asSequence().filter { it.value.isEmpty }
-    val completeEntitySetsUnionSql = if (emptyEntitySetIds.iterator().hasNext()) { // if there are entity set ids with no entity ids in entityKeyIds
-        val completeEntitySets = emptyEntitySetIds.joinToString(",") { "'${it.key}'" }
-        "UNION SELECT $selectColumns FROM ${IDS.name} WHERE ${ENTITY_SET_ID.name} IN ($completeEntitySets)"
-    } else {
-        ""
-    }
+    val dataKeySql = "AND " + entityKeyIds.map { (entitySetId, entityKeyIds) ->
```
you-leee: Can we put this back into buildEntitiesClause? Or just uncomment that part and use it here?

geekbeast: Once we make sure those queries aren't causing perf issues.

geekbeast: Why not? I brought it back.

PR #500, opened 2019-05-25
LATTICE-2017 use row-by-row locking and deleting/clearing function fo…

you-leee:
…r edges

### DIFF
--- a/src/main/java/com/openlattice/graph/core/GraphService.java
+++ b/src/main/java/com/openlattice/graph/core/GraphService.java
@@ -41,21 +41,21 @@ public interface GraphService {
 
     WriteEvent createEdges( Set<DataEdgeKey> keys );
 
-    int clearEdges( Set<DataEdgeKey> keys );
+    int clearEdges( Iterable<DataEdgeKey> keys );
 
-    int clearVerticesInEntitySet( @Nonnull UUID entitySetId );
+    int clearVerticesInEntitySetWithoutLocking( @Nonnull UUID entitySetId );
 
-    int clearVertices( @Nonnull UUID entitySetId, @Nonnull Set<UUID> vertices );
+    int clearVerticesWithoutLocking( @Nonnull UUID entitySetId, @Nonnull Set<UUID> vertices );
 
-    int clearVerticesOfAssociations( UUID entitySetId, Set<UUID> vertices );
+    int clearVerticesOfAssociationsWithoutLocking( UUID entitySetId, Set<UUID> vertices );
 
-    WriteEvent deleteEdges( Set<DataEdgeKey> keys );
+    WriteEvent deleteEdges( Iterable<DataEdgeKey> keys );
 
-    int deleteVerticesInEntitySet( UUID entitySetId );
+    int deleteVerticesInEntitySetWithoutLocking( UUID entitySetId );
 
-    int deleteVertices( UUID entitySetId, Set<UUID> vertices );
+    int deleteVerticesWithoutLocking( UUID entitySetId, Set<UUID> vertices );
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -158,7 +158,7 @@ class Graph(private val hds: HikariDataSource, private val edm: EdmManager) : Gr
 
     /* Delete  */
 
-    override fun clearEdges(keys: MutableSet<DataEdgeKey>): Int {
+    override fun clearEdges(keys: Iterable<DataEdgeKey>): Int {
```
katherinebernstein: I think we should include component type in the SQL to clear by vertex (or vertices) -- doesn't matter in cases where an entity was deleted, but if we're ever trying to delete a specific edge A->B->C (but none of the entities were deleted), if there was also an edge from C->B->A we wouldn't want to accidentally delete that one too.

katherinebernstein: Actually I realize this should never happen because we shouldn’t be reusing association entities but the system would support this case

src/main/java/com/openlattice/graph/core/GraphService.java
```diff
@@ -41,21 +41,21 @@
 
     WriteEvent createEdges( Set<DataEdgeKey> keys );
 
-    int clearEdges( Set<DataEdgeKey> keys );
+    int clearEdges( Iterable<DataEdgeKey> keys );
 
-    int clearVerticesInEntitySet( @Nonnull UUID entitySetId );
+    int clearVerticesInEntitySetWithoutLocking( @Nonnull UUID entitySetId );
```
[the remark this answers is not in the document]
you-leee: They are not used right now, I just wasn't sure, if we should delete them, in case row-by row locking will be too slow.

src/main/kotlin/com/openlattice/graph/Graph.kt
you-leee: We do include Component type in the SQL, in case of A->B->C, we store it in 3 rows:
ACB, Comp type src
CBA, Comp type dst
BAC, Comp type edge
C->B->A would have row CBA, Comp type src, so that won't be deleted.

katherinebernstein: But it looks like CLEAR_BY_VERTEX_SQL doesn't use the comp_type column

katherinebernstein: (or LOCK_BY_VERTEX_SQL)

you-leee: Oh, you are totally right, for some reason I was under the impression, that addSrcKeyIds, addDstKeyIds and addEdgeKeyIds included Component type too.

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -820,8 +790,9 @@ enum class ComponentType {
 private val KEY_COLUMNS = setOf(
         ID_VALUE,
         EDGE_COMP_1,
-        EDGE_COMP_2
-).map { col -> col.name }.toList()
+        EDGE_COMP_2,
+        COMPONENT_TYPES
+).map { it.name }.toSet()
```
katherinebernstein: kotlin will keep this ordered right?

you-leee: Yes

geekbeast: Why .toSet?

PR #501, opened 2019-05-28
avoid direct indexing of audit entities

katherinebernstein:


src/main/java/com/openlattice/data/storage/HazelcastEntityDatastore.java
```diff
@@ -263,6 +264,11 @@ private void signalDeletedEntities( UUID entitySetId, Set<UUID> entityKeyIds ) {
                 .forEach( this::markMaterializedEntitySetDirty );
     }
 
+    private boolean shouldIndexDirectly( UUID entitySetId, Set<UUID> entityKeyIds ) {
+        return entityKeyIds.size() < BATCH_INDEX_THRESHOLD && !edmManager.getEntitySet( entitySetId ).getFlags()
```
geekbeast: This is quite the expensive call. We should make more use of projections :-/

PR #504, opened 2019-05-30
LATTICE-2023 add batch for each linking id when locking

you-leee:


src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
```diff
@@ -49,16 +49,19 @@ private const val AVG_SCORE_FIELD = "avg_score"
...
+
+        val psLocks = connection.prepareStatement(LOCK_CLUSTERS_SQL)
+        clusters.forEach {
+            psLocks.setObject(1, it)
+            psLocks.addBatch()
+        }
+        psLocks.executeBatch()
+        connection.commit()
+
+        return StatementHolder(connection, psLocks, null)
```
geekbeast: Just make a statement holder that doesn't take a a result set. Somewhat curious where this statement holder is used as the purpose of a statement holder is so that a postgres iterable can instantiate a postgres iterator on demand.

you-leee: It's used in BackgroundLinkingService, to lock these rows until we insert the new matches. I also should remove the commit() call there.

PR #505, opened 2019-05-31
Bugfix/lattice 1996 include entitykeyids in materialized sets

you-leee:


src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
```diff
@@ -384,19 +386,22 @@ internal fun selectEntityKeyIdsWithCurrentVersionSubquerySql(
         omitEntitySetId: Boolean,
         joinColumns: List<String>
 ): String {
-    val metadataColumns = getMetadataOptions(metadataOptions, linking).joinToString(",")
+    val metadataColumns = getMetadataOptions(metadataOptions.minus(MetadataOption.ENTITY_KEY_IDS), linking)
```
katherinebernstein: will this MetadataOption always have to be special cased? If so it might be good to push that filtering into `getMetadataOptions`

you-leee: Nope, it's just because for non-linking queries, the name of column is the same as the metadataoption, except for ENTTITY_KEY_IDS, where you do an aggregate for both linking and non-linking queries. It will be used in materialized views, so we can join on edge tables (we need the same type of column in materialized views, so we can run the same function to do the edges table join)

katherinebernstein: would that not be true for all non-linking queries then? Like if someone tries to read data from a normal entity set and the ENTITY_KEY_IDS MetadataOption is included, no special handling would be required?

you-leee: It's only required here, because we only aggregate on it here, in other cases, we only select the already aggregated column

katherinebernstein: Got it, makes sense -- thanks!

PR #506, opened 2019-06-04
Bugfix/lattice 2030 integrateassociations id assignment

you-leee:


src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -490,13 +479,32 @@ open class DataGraphService(
...
                     entityKeyIds.putAll(ids)
                     entitySetId to ids.asSequence().map { it.key.entityId to it.value }.toMap()
                 }.toMap()
-        //Retrieve the src/dst keys
+
+        // Retrieve the src/dst keys (it adds all entitykeyids to mutable entityKeyIds)
         idService.getEntityKeyIds(entityKeys, entityKeyIds)
 
+        val edges = associations
+                .asSequence()
```
makosblade: It seems like this adds an extra iteration across `associations` vs having it above in the `.forEach`. I'm not sure what the scale of `associations` usually is so 🤷‍♂

katherinebernstein: I'm not sure it's possible to avoid iterating over it twice and still ensure the src/dst types are allowed before writing any data -- but the scale is pretty large so if it is possible it would definitely be preferable

PR #522, opened 2019-06-13
Fix subscription code for new neighborhood query.

geekbeast:


src/main/kotlin/com/openlattice/graph/PostgresGraphQueryService.kt
```diff
@@ -645,9 +645,20 @@ class PostgresGraphQueryService(
         return "SELECT * FROM ${EDGES.name} WHERE $idsClause $dstEntitySetIdsClause $associationEntitySetIdsClause $componentTypeClause"
     }
 
-    private fun inClause(uuids: Set<UUID>): String {
-        check(uuids.isNotEmpty()) { "Ids must be provided." }
-        return "IN (" + uuids.joinToString(",") { id -> "'$id'" } + ")"
+    private fun inClause(ids: Set<UUID>): String {
+        return "IN (" + ids.joinToString(",") { "'$it'" } + ")"
```
katherinebernstein: why remove the non-empty check?

katherinebernstein: oh nvm it's below

katherinebernstein: inClause is still called from outside dataKeysClause though so it might be worthwhile to keep here

geekbeast: ?

PR #525, opened 2019-06-14
Bugfix/lattice 2048 edge audit entityset deletion

you-leee:


src/main/kotlin/com/openlattice/auditing/AuditingComponent.kt
```diff
@@ -63,7 +63,9 @@ interface AuditingComponent {
         return if (auditingConfiguration.isAuditingInitialized()) {
             events
                     .groupBy { ares.getActiveAuditEntitySetIds(it.aclKey, it.eventType) }
-                    .filter { (auditEntitySetConfiguration, entities) -> auditEntitySetConfiguration.auditRecordEntitySet != null }
+                    .filter { (auditEntitySetConfiguration, entities) ->
+                        auditEntitySetConfiguration.auditRecordEntitySet != null &&
+                                auditEntitySetConfiguration.auditEdgeEntitySet != null }
```
katherinebernstein: filtering out the audit edge entity sets will prevent auditing EDM events (and eventually organizations once we implement auditing there) so we might actually want to make the audit entity nullable and handle the case where they don't exist

geekbeast: +1

you-leee: Oh, I see! I will remove it from her. Do we want to initialize edge audit entity set if it's null here or just skip creating those associations?

katherinebernstein: probably just skip creating the associations since there's no EDM entity set or organization entity set, so there's no entity to create an edge to

PR #526, opened 2019-06-15
Feature/lattice 1842 materialized view refresh automatization

you-leee:


src/main/kotlin/com/openlattice/assembler/tasks/MaterializedEntitySetsDataRefreshTask.kt
```diff
@@ -0,0 +1,121 @@
...
+    private fun getAuthorizedPropertiesForMaterialization(organizationId: UUID, entitySetId: UUID): Map<UUID, PropertyType> {
+        val organizationPrincipal = getDependency().organizations.getOrganizationPrincipal(organizationId)
+        //This will be rare, since it is unlikely you have access to an organization that does not exist.
+                ?: throw ResourceNotFoundException("Organization does not exist.")
+
+        // check materialization on all linking and normal entity sets
+        val entitySet = getDependency().edm.getEntitySet(entitySetId)
+        val allEntitySetIds = mutableSetOf(entitySet.id) + entitySet.linkedEntitySets
+
+        allEntitySetIds.forEach { ensureMaterialize(it, organizationPrincipal) }
```
katherinebernstein: If an organization has MATERIALIZE on a linked entity set but is missing permissions on one of its normal entity sets, we might still want to materialize the linked entity set without data from the entity set missing permissions instead of blocking materialization altogether (basically mimic read behavior).

you-leee: Right now, we only allow read, when the calling user has read on all normal entity sets, so this is mimicking it. I don't remember the whole conversation, but it has to do with not being able to filter values by user for materialized entity sets, because we don't know which value is from which normal entityset, so we only allow read, if user has read on all normal entity sets, hence do the same for materializing.

katherinebernstein: Ah true we are checking read on all for reading linked entities -- I don't think necessarily want to be doing that. The partial permissions concern was for the case where you only have READ on first name in entity set A, and only READ on last name in entity set B, and linked A+B allows you to reconstruct full names. This was the rationale for requiring READ on all entity sets contributing data for a property type, and also isn't a concern if an entire entity set is filtered out.

(Requiring READ on all entity sets of a property type also doesn't actually prevent people from reconstructing names by doing things like creating associations with PII to normal entities involved in a linked entity set.)

you-leee: True, but still, after we materialize a linking entity set and we start to add read permissions on the postgres columns, we cannot filter out values for which the organization member has or has not read permissions for in the normal entity sets.
If we start to filter out whole normal entity sets from materialization (and loading linking entity sets), we will still require the organization member to have read on that whole sub-set of normal entity sets, which provided the data for that linking entity set and the org principal had materialized on.

katherinebernstein: That makes sense -- theoretically we can compute user permissions per column based on the entity sets the organization is authorized to materialize, but that could get overly complex. I think this is fine for now but we might want to revisit the way linking permissions are granted

src/main/java/com/openlattice/edm/PostgresEdmManager.java
```diff
@@ -226,6 +224,37 @@ public void revoke(
         }
     }
 
+    public Iterable<UUID> getAllEntitySetIdsForType( UUID entityTypeId ) {
+        final var getEntitySetIdsByType = String.format( "SELECT %1$s FROM %2$s WHERE %3$s = ?",
+                ID.getName(),
+                ENTITY_SETS,
+                ENTITY_TYPE_ID_FIELD );
+        return new PostgresIterable<>(
```
katherinebernstein: Is there a reason this is read from postgres instead of hazelcast?

src/main/kotlin/com/openlattice/assembler/tasks/MaterializedEntitySetsDataRefreshTask.kt
you-leee: I will include this on wiki page and and a TODO

src/main/java/com/openlattice/edm/PostgresEdmManager.java
you-leee: Now, that you mention it, I don't see any reason, just wanted to have a version, where we don't return the whole entity set, only the id. I will turn it into a hazelcast map query with filtering on entity type index. 
Btw, we have a lot of calls from EdmService to PostgresEdmManager.getAllEntitySetsForType, although there is a version in EdmService, which does exactly the same but using hazelcast. I will change those calls too to use hazelcast.

katherinebernstein: Yeah avoiding loading the entire entity set is for sure ideal

PR #527, opened 2019-06-18
edm sync is HazelcastInitializationTask + run before AuditInitializat…

katherinebernstein:
…ionTask


src/main/kotlin/com/openlattice/edm/tasks/EdmSyncInitializerTask.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.openlattice.tasks.PostConstructInitializerTaskDependencies
+import com.openlattice.tasks.Task
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
+import org.apache.olingo.commons.api.edm.FullQualifiedName
+import org.slf4j.LoggerFactory
+
+class EdmSyncInitializerTask : HazelcastInitializationTask<EdmSyncInitializerDependencies> {
+
+    companion object {
+        private val OL_AUDIT_FQN = FullQualifiedName("OPENLATTICE_AUDIT", "AUDIT")
```
you-leee: I think we should also include audit edge type here. And it would be nice to get the fqns from auditing.yaml.

katherinebernstein: I don't think the actual audit entity types break the EDM import -- this one is legacy (we should probably just delete it from the EDM actually)

you-leee: Oh, I see. Totally confused this entity type / property types with current auditing. Yeah, the sync task works fine without removing these from edm.

PR #530, opened 2019-06-19
when an entity set is moved to a different organization, also move it…

katherinebernstein:
…s audit entity sets to the new organization

### DIFF
--- a/src/main/java/com/openlattice/datastore/services/EdmService.java
+++ b/src/main/java/com/openlattice/datastore/services/EdmService.java
@@ -987,6 +987,28 @@ public class EdmService implements EdmManager {
         }
         entitySets.executeOnKey( entitySetId, new UpdateEntitySetMetadataProcessor( update ) );
         eventBus.post( new EntitySetMetadataUpdatedEvent( getEntitySet( entitySetId ) ) );
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -987,6 +987,28 @@ public void updateEntitySetMetadata( UUID entitySetId, MetadataUpdate update ) {
...
+        /* If an entity set is being moved across organizations, its audit entity sets should also be moved to the new organization */
+        if ( update.getOrganizationId().isPresent() ) {
+
+            AclKey aclKey = new AclKey( entitySetId );
+
+            Set<UUID> auditEntitySetIds = Sets.union( aresManager.getAuditRecordEntitySets( aclKey ),
+                    aresManager.getAuditEdgeEntitySets( aclKey ) );
+
+            entitySets.executeOnKeys( auditEntitySetIds,
+                    new UpdateEntitySetMetadataProcessor( new MetadataUpdate( Optional.empty(),
```
geekbeast: lulz we need to convert this to kotlin and use @JvmOverloads

PR #532, opened 2019-06-20
step 1

makosblade:
Gradles


build.gradle
```diff
@@ -105,11 +105,11 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
 }
 
 dependencies {
-    spotbugsPlugins 'com.h3xstream.findsecbugs:findsecbugs-plugin:1.8.0'
-    compileOnly 'net.jcip:jcip-annotations:1.0'
-    compileOnly 'com.github.spotbugs:spotbugs-annotations:3.1.9'
-    testCompileOnly 'net.jcip:jcip-annotations:1.0'
-    testCompileOnly 'com.github.spotbugs:spotbugs-annotations:3.1.9'
+    spotbugsPlugins "com.h3xstream.findsecbugs:findsecbugs-plugin:${findsecbugs_version}"
```
katherinebernstein: some of these versions are going backwards -- is that intentional / should gradles be updated for the latest versions we use?

makosblade: Downgrade Findsecbugs `1.8.0` => `1.7.1`
Would like to bump the `openlattice.gradle` version after this

build.gradle
```diff
@@ -105,11 +105,11 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
...
 
 dependencies {
-    spotbugsPlugins 'com.h3xstream.findsecbugs:findsecbugs-plugin:1.8.0'
-    compileOnly 'net.jcip:jcip-annotations:1.0'
-    compileOnly 'com.github.spotbugs:spotbugs-annotations:3.1.9'
-    testCompileOnly 'net.jcip:jcip-annotations:1.0'
-    testCompileOnly 'com.github.spotbugs:spotbugs-annotations:3.1.9'
+    spotbugsPlugins "com.h3xstream.findsecbugs:findsecbugs-plugin:${findsecbugs_version}"
+    compileOnly "net.jcip:jcip-annotations:${jcip_version}"
+    compileOnly "com.github.spotbugs:spotbugs-annotations:${spotbugs_annotations_version}"
```
makosblade: Downgrade spotbugs-annotations `3.1.9` => `3.1.3`

build.gradle
makosblade: I would very much like for us to be able to push our gradle versions to the most recent we can. So I figure that'll be something of a follow-up eventually

PR #533, opened 2019-06-24
LATTICE-1927: implement entity type property metadata

jokedurnez:


jokedurnez:
Like in datastore: not sure how to make tests work since dependencies are defined in other PR's.

src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -547,6 +535,19 @@ private void setupDefaultEntitySetPropertyMetadata( UUID entitySetId, UUID entit
         } );
     }
 
+    private void setupDefaultEntityTypePropertyMetadata( UUID entityTypeId ) {
+        final var et = getEntityType( entityTypeId );
+        et.getProperties().forEach( propertyTypeId -> {
+            EntityTypePropertyKey key = new EntityTypePropertyKey( entityTypeId, propertyTypeId );
+            PropertyType property = getPropertyType( propertyTypeId );
```
katherinebernstein: Reads from hazelcast are expensive and `getPropertyType` reads from hazelcast, so it would be faster to bulk the hazelcast calls by calling `propertyTypes.getAll(et.getProperties())` before the forEach loop, and then referencing that map

src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -1628,6 +1629,28 @@ public void updateEntitySetPropertyMetadata( UUID entitySetId, UUID propertyType
         entitySetPropertyMetadata.executeOnKey( key, new UpdateEntitySetPropertyMetadataProcessor( update ) );
     }
 
+    @Override
+    public Map<UUID, EntityTypePropertyMetadata> getAllEntityTypePropertyMetadata( UUID entityTypeId ) {
+        return getEntityType( entityTypeId ).getProperties().stream()
```
katherinebernstein: Same comment here on the hazelcast load, since `getEntityTypePropertyMetadata` reads from hazelcast -- it would be faster to use a Predicate on `entityTypeId` (and add a map config in the mapstore)

src/test/java/com/openlattice/hazelcast/serializers/EntityTypePropertyKeyStreamSerializerTest.java
```diff
@@ -0,0 +1,24 @@
...
+private static final long serialVersionUID = -4933403371497497344L;
+
+@Override
+protected EntityTypePropertyKeyStreamSerializer createSerializer() {
+        return new EntityTypePropertyKeyStreamSerializer();
+        }
+
+@Override
+protected EntityTypePropertyKey createInput() {
+        return new EntityTypePropertyKey( UUID.randomUUID(), UUID.randomUUID() );
```
katherinebernstein: I like to throw these in TestDataFactory since it can be useful to have a mock object easily available

src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -124,6 +109,7 @@
     private final IMap<UUID, String>                                    names;
     private final IMap<UUID, AssociationType>                           associationTypes;
     private final IMap<EntitySetPropertyKey, EntitySetPropertyMetadata> entitySetPropertyMetadata;
+    private final IMap<EntityTypePropertyKey, EntityTypePropertyMetadata> entityTypePropertyMetadata;
```
katherinebernstein: does this mapstore exist? /a postgres table where this data will live?

jokedurnez: That part I really don't understand.  It seems to work, but I don't know where in the database this data is stored... :-\

katherinebernstein: You'll want to add a table in PostgresTable which specifies all the columns you need, and the data will be stored there. Then, you should add a mapstore which explains how to write data into the table (you can look at EntitySetPropertyMetadataMapstore for reference), and add it as a bean in MapstoresPod.

jokedurnez:
I think I added everything from the review.  Also this was very incomplete, but I think it should be done now.
This is passing: https://github.com/openlattice/rehearsal/pull/51

jokedurnez:
This is done ! :-)

src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -311,6 +311,24 @@ public void createEntityType( EntityType entityTypeRaw ) {
...
+        et.getProperties().forEach( propertyTypeId -> {
+            EntityTypePropertyKey key = new EntityTypePropertyKey( entityTypeId, propertyTypeId );
+            PropertyType property = propertyTypeMap.get( propertyTypeId );
+            EntityTypePropertyMetadata metadata = new EntityTypePropertyMetadata(
+                    property.getTitle(),
+                    property.getDescription(),
+                    new LinkedHashSet(propertyTags.get(propertyTypeId)),
+                    true
+                    );
+            entityTypePropertyMetadata.put( key, metadata );
```
katherinebernstein: This isn't going to be that expensive since we're not creating entity types all the time / they typically don't have a huge number of property types, but this call can also be bulked by instantiating a `Map<EntityTypePropertyKey, EntityTypePropertyMetadata>` of size (# property types in entity set), adding the values to that map, and then after exiting the `.forEach`, calling `entityTypePropertyMetadata.putAll( <yourMap> )` -- not a huge deal but would speed up entity type creation a little bit, and is a little bit safer in case something goes wrong / throws an exception halfway through the forEach (so that you don't end up in a weird halfway state)

src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -1214,6 +1231,29 @@ public EntityDataModelDiff getEntityDataModelDiff( EntityDataModel edm ) {
         return Lists.newArrayList( result, conflicts );
     }
 
+    @Override
+    public Map<UUID, EntityTypePropertyMetadata> getAllEntityTypePropertyMetadata( UUID entityTypeId ) {
+        return entityTypePropertyMetadata
+                .entrySet( Predicates.equal( EntityTypePropertyMetadataMapstore.ENTITY_TYPE_INDEX, entityTypeId ) )
+                .stream()
+                .collect( Collectors.toMap( entry -> entry.getKey().getPropertyTypeId(), entry -> entry.getValue() ) );
```
katherinebernstein: I generally prefer the `Entry::getValue` syntax over `entry -> entry.getValue()` syntax, but not a big deal!

src/test/java/com/openlattice/hazelcast/serializers/EntityTypePropertyMetadataStreamSerializerTest.java
```diff
@@ -0,0 +1,28 @@
...
+    private static final long serialVersionUID = 5114029297563838101L;
+
+    @Override
+    protected EntityTypePropertyMetadataStreamSerializer createSerializer() {
+        return new EntityTypePropertyMetadataStreamSerializer();
+    }
+
+    @Override
+    protected EntityTypePropertyMetadata createInput() {
+        return new EntityTypePropertyMetadata( "title", "description" );
```
katherinebernstein: same TestDataFactory comment as above

src/main/java/com/openlattice/edm/types/processors/UpdateEntityTypePropertyMetadataProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+    @SuppressFBWarnings( value = "SE_BAD_FIELD", justification = "Custom Stream Serializer is implemented" )
+
+    private final        MetadataUpdate update;
+
+    public UpdateEntityTypePropertyMetadataProcessor( MetadataUpdate update ) {
+        this.update = update;
+    }
+
+    @Override
+    public Object process( Map.Entry<EntityTypePropertyKey, EntityTypePropertyMetadata> entry ) {
```
katherinebernstein: this would be a lot simpler in kotlin :)

PR #535, opened 2019-06-25
Bugfix/lattice 2082 add members enrty processor

you-leee:


you-leee:
Contains fixes for both https://jira.openlattice.com/browse/LATTICE-2082 and https://jira.openlattice.com/browse/LATTICE-2085

src/main/java/com/openlattice/organizations/HazelcastOrganizationService.java
```diff
@@ -320,25 +323,36 @@ private void addMembers( AclKey orgAclKey, Set<Principal> members ) {
...
+                        new SecurablePrincipalList( securablePrincipals )
+                )
+        );
+    }
 
-        eventBus.post( new MembersRemovedFromOrganizationEvent( organizationId, new PrincipalSet( members ) ) );
+    private void removeRolesFromMembers( Stream<AclKey> roles, Collection<SecurablePrincipal> members ) {
+        members.forEach( securablePrincipal ->
+                roles.forEach( role ->
+                        securePrincipalsManager.removePrincipalFromPrincipal( role, securablePrincipal.getAclKey() )
```
katherinebernstein: We should probably allow for a set of AclKeys to be passed to securePrincipalsManager so that we can bulk the calls to hazelcast instead of making individual calls for each principal/role pair.

Also, removePrincipalFromPrincipal ends up constructing a singleton set for each securable principal which it passes to an entry processor, so we should probably just support passing the collection through directly instead of iterating and calling individually.

src/main/java/com/openlattice/organizations/roles/HazelcastPrincipalService.java
```diff
@@ -375,6 +366,10 @@ private static Predicate findPrincipal( Principal p ) {
         return Predicates.equal( "principal", p );
     }
 
+    private static Predicate findPrincipals( Collection<Principal> principals ) {
+        return Predicates.in( "principal", principals.toArray( new Principal[] {} ) );
```
katherinebernstein: Better to use the `PrincipalMapstore.PRINCIPAL_INDEX` constant

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -183,28 +183,32 @@ class AssemblerConnectionManager(
...
-                            dbName,
-                            buildPostgresUsername(securePrincipalsManager.getPrincipal(principal.id))
-                    )
-                }
+    fun configureNewUsersInDatabase(
+            dbName: String,
+            dataSource: HikariDataSource,
+            principals: Collection<SecurablePrincipal>) {
+        principals.forEach {
+            configureUserInDatabase(dataSource, dbName, buildPostgresUsername(it))
```
katherinebernstein: any reason not to batch these too?

src/main/java/com/openlattice/organizations/HazelcastOrganizationService.java
you-leee: Gotcha, will do!

src/main/java/com/openlattice/organizations/roles/HazelcastPrincipalService.java
you-leee: :thumbsup:

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
you-leee: Nope, will do!

src/main/java/com/openlattice/organizations/HazelcastOrganizationService.java
```diff
@@ -320,25 +323,37 @@ private void addMembers( AclKey orgAclKey, Set<Principal> members ) {
...
+                new MembersRemovedFromOrganizationEvent(
+                        organizationId,
+                        new SecurablePrincipalList( securablePrincipals )
+                )
+        );
+    }
 
-        eventBus.post( new MembersRemovedFromOrganizationEvent( organizationId, new PrincipalSet( members ) ) );
+    private void removeRolesFromMembers( Set<AclKey> roles, Collection<SecurablePrincipal> members ) {
+        members.forEach( securablePrincipal ->
```
katherinebernstein: I think this can still be bulkier -- this will still be calling `principalTrees.executeOnKey` for every member, but you can just pass all the member aclKeys through and call `principalTrees.executeOnKeys` for them all at once

src/main/java/com/openlattice/organizations/HazelcastOrganizationService.java
```diff
@@ -320,25 +323,37 @@ private void addMembers( AclKey orgAclKey, Set<Principal> members ) {
...
+                securePrincipalsManager.removePrincipalsFromPrincipal( roles, securablePrincipal.getAclKey() )
+        );
     }
 
-    private void removeRolesFromMembers( Stream<AclKey> roles, Stream<AclKey> members ) {
-        members.forEach( member -> roles
-                .forEach( role -> securePrincipalsManager.removePrincipalFromPrincipal( role, member ) ) );
+    private void removeOrganizationFromMembers( AclKey organization, Collection<SecurablePrincipal> members ) {
+        members.forEach( securablePrincipal ->
+                securePrincipalsManager.removePrincipalFromPrincipal( organization, securablePrincipal.getAclKey() )
```
katherinebernstein: ^same comment

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -183,28 +183,31 @@ class AssemblerConnectionManager(
...
+        configureUsersInDatabase(dataSource, dbName, userNames)
+    }
+
+    fun removeMembersFromOrganization(
+            dbName: String,
+            dataSource: HikariDataSource,
+            members: Collection<SecurablePrincipal>
+    ) {
+        members.forEach { securablePrincipal ->
+            revokeConnectAndSchemaUsage(
```
katherinebernstein: Same thing looks possible here too

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -622,26 +625,29 @@ class AssemblerConnectionManager(
...
+                statement.execute("GRANT USAGE ON SCHEMA $MATERIALIZED_VIEWS_SCHEMA TO $userIdsSql")
                 //Don't allow users to access public schema which will contain foreign data wrapper tables.
-                logger.info("Revoking $PUBLIC_SCHEMA schema right from user: {}", userId)
-                statement.execute("REVOKE USAGE ON SCHEMA $PUBLIC_SCHEMA FROM $dbUser")
+                logger.info("Revoking $PUBLIC_SCHEMA schema right from user: {}", userIds)
+                statement.execute("REVOKE USAGE ON SCHEMA $PUBLIC_SCHEMA FROM $userIdsSql")
                 //Set the search path for the user
-                statement.execute("ALTER USER $dbUser set search_path TO $MATERIALIZED_VIEWS_SCHEMA")
+                userIds.forEach { userId ->
+                    statement.execute("ALTER USER $userId SET search_path TO $MATERIALIZED_VIEWS_SCHEMA")
```
katherinebernstein: also batchable?

you-leee: ALTER USER is not batchable unfortunately

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
you-leee: Right, thanks!

src/main/java/com/openlattice/organizations/HazelcastOrganizationService.java
you-leee: Yeah, right, thanks!

PR #547, opened 2019-07-05
LATTICE-2102 add case for bidirectional checks of allowed src/dst ent…

you-leee:
…ity types


src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -601,12 +594,51 @@ open class DataGraphService(
...
+
+    private fun checkAllowedEntityTypesOfAssociation(
+            edgeAssociationType: AssociationType,
+            edgeEntitySetId: UUID,
+            srcEntityTypes: Collection<UUID>,
+            srcEntitySetIds: Set<UUID>,
+            dstEntityTypes: Collection<UUID>,
+            dstEntitySetIds: Set<UUID>
+    ) {
+        val isSrcMNotAllowed = !edgeAssociationType.src.containsAll(srcEntityTypes)
```
katherinebernstein: typo

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -601,12 +594,51 @@ open class DataGraphService(
...
+        val isDstNotAllowed = !edgeAssociationType.dst.containsAll(dstEntityTypes)
+
+        if (edgeAssociationType.isBidirectional) {
+            val isSrcNotAllowedInDst = !edgeAssociationType.dst.containsAll(srcEntityTypes)
+            val isDstNotAllowedInSrc = !edgeAssociationType.src.containsAll(dstEntityTypes)
```
katherinebernstein: I'm not sure this will do the right thing -- if the src (or dst) entity types that are used are actually divided between the src and dst types on a bidirectional association type, it looks like this will still fail

you-leee: Oh, I see, because it's checking in a batch.. I guess bidirectional cases need to handled individually for each association.

you-leee:
Can you pls elaborate on that? I mean, what is the problem of throwing the error in DataGraphService?

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -559,39 +586,61 @@ open class DataGraphService(
...
+     * @param dstAssociationEntitySetIds The entity set ids of the dst entities in associations mapped by their association entity set id.
+     * @param bidirectionalEntitySetIds The src and dst entity set id pairs of bidirectional associations mapped by the association entity set id.
+     * @param associationTypes Map of association entity set ids and their association types.
+     */
     private fun checkAssociationEntityTypes(
-            srcAssociationEntitySetIds: Map<UUID, Set<UUID>>, dstAssociationEntitySetIds: Map<UUID, Set<UUID>>
+            srcAssociationEntitySetIds: Map<UUID, Set<UUID>>,
+            dstAssociationEntitySetIds: Map<UUID, Set<UUID>>,
+            bidirectionalEntitySetIds: Map<UUID, Set<Pair<UUID, UUID>>>, // edge - (src, dst)
+            associationTypes: Map<UUID, AssociationType> // entityset id - associationtype
```
katherinebernstein: It's probably also better to load all association types + entity types in a few bulk calls here instead of having `integrateAssociations` handle part of it while iterating and having `checkAllowedEntityTypesOfAssociation` handle the rest:
1) load all the entity type ids from the combined keySets and values of `srcAssociationEntitySetIds` and `dstAssociationEntitySetIds`
2) Load all the association types in bulk from the entity type ids of the association entity sets
3) Load all the src+dst entity types in bulk from the entity type ids of the src+dst entity sets

You can also avoid iterating over bidirectional and non-bidirectional (unidirectional? idk) association types separately / writing separate functions to check their allowed entity types. Bidirectional is basically shorthand for saying that `src == src + dst` and `dst == src + dst`, so you can just combine their src and dst entityTypeId sets and use that combined set for both src and dst. i.e. something like

```
val entityTypeIdsByEntitySetId = ...
val associationTypesByAssociationTypeId = ...

(srcAssociationEntitySetIds.keys + dstAssociationEntitySetIds.keys).forEach { edgeEntitySetId -> 
  val associationTypeId = entityTypeIdsByEntitySetId.getValue(edgeEntitySetId)
  val associationType = associationTypesByAssociationTypeId.getValue(associationTypeId)

  val requestedSrcEntityTypeIds = srcAssociationEntitySetIds.getValue(edgeEntitySetId).map{ entityTypeIdsByEntitySetId.getValue(it)
  val requestedDstEntityTypeIds = dstAssociationEntitySetIds.getValue(edgeEntitySetId).map{ entityTypeIdsByEntitySetId.getValue(it)

  val allowedSrcEntityTypeIds = associationType.src
  val allowedDstEntityTypeIds = associationType.dst

  if (associationType.isBidirectional) {
    allowedSrcEntityTypeIds.addAll(associationType.dst)
    allowedDstEntityTypeIds.addAll(associationType.src)
  }

  if (!allowedSrcEntityTypeIds.containsAll(requestedSrcEntityTypeIds) || !allowedDstEntityTypeIds.containsAll(requestedDstEntityTypeIds)) {
    // much sadness
  }

}
```

^btw no guarantees on the compilability of my kotlin

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -468,22 +476,41 @@ open class DataGraphService(
...
                 .asSequence()
                 .forEach { association ->
                     val srcEsId = association.src.entitySetId
                     val dstEsId = association.dst.entitySetId
                     val edgeEsId = association.key.entitySetId
 
-                    if (srcAssociationEntitySetIds.putIfAbsent(edgeEsId, mutableSetOf(srcEsId)) != null) {
-                        srcAssociationEntitySetIds.getValue(edgeEsId).add(srcEsId)
-                    }
+                    val edgeAssociationType = edmManager.getAssociationTypeByEntitySetId(edgeEsId)
```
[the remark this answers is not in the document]
you-leee: 👍

geekbeast:
> Can you pls elaborate on that? I mean, what is the problem of throwing the error in DataGraphService?

This is more of a separation of concerns-- DataGraphService won't always be invoked by a controller. While using custom exceptions will probably be handled fine by clients it does create this weird coupling where *at least* BaseExceptionHandler needs to be present in order to correctly return errors. 

It also is additional code smell that we're creating multiple levels of validation instead of failing fast. Code execution shouldn't be getting down to DataGraphService and still have to do access control and validation checks (that belongs in the controller). There's another reasonable pattern here (push validation to the greatest common denominator), but this means that there is a lot of execution that can happen before validation kicks in and increases the likelihood of something breaking if layers are inserted in between down the road.

you-leee:
Thanks a lot of the explanation!

src/main/kotlin/com/openlattice/data/DataGraphService.kt
you-leee: "Bidirectional is basically shorthand for saying that src == src + dst and dst == src + dst": I don't think, that is enough to check, because in this case you could still mix and match. For example, allowed src types are `A, B` and allowed dst types are `C,D`, that check would allow `src = A` and `dst = B`

katherinebernstein: I think that might still be correct, although that's a good point and isn't well-defined anywhere.

PR #548, opened 2019-07-09
LATTICE-2106 use ids table to clear entities, small fixes

you-leee:


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -327,20 +328,20 @@ internal val upsertEntitiesSql = "UPDATE ${PostgresTable.ENTITY_KEY_IDS.name} SE
...
  * 2. partitions
  */
-internal val lockEntitiesSql = "SELECT 1 FROM ${PostgresTable.IDS.name} " +
+internal val lockEntitiesSql = "SELECT 1 FROM ${IDS.name} " +
         "WHERE ${ID_VALUE.name} = ANY(?) AND ${PARTITION.name} = ANY(?) " +
         "FOR UPDATE"
 
-
-fun upsertEntities(entitySetId: UUID, idsClause: String, version: Long): String {
+@Deprecated("Old version of upsert")
```
katherinebernstein: Looks like it's safe to delete this all the way through to the replace calls in PEDQSOld (since the PEDQSOld.upsertEntities call is only used by unused functions)

you-leee: Deleted

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryServiceOld.kt
```diff
@@ -376,149 +376,6 @@ class PostgresEntityDataQueryServiceOld(
         }, adapter)
     }
 
-    fun upsertEntities(
```
katherinebernstein: ❤️

PR #549, opened 2019-07-10
LATTICE-2103 filter out non-existing principals when granting select

you-leee:


src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -391,12 +391,19 @@ class AssemblerConnectionManager(
...
+        val permissionsToCheck = EdmAuthorizationHelper.READ_PERMISSION
         // collect all principals of type user, role, which have read access on entityset
         val authorizedPrincipals = securePrincipalsManager
-                .getAuthorizedPrincipalsOnSecurableObject(AclKey(entitySet.id), EdmAuthorizationHelper.READ_PERMISSION)
+                .getAuthorizedPrincipalsOnSecurableObject(AclKey(entitySet.id), permissionsToCheck)
                 .filter { it.type == PrincipalType.USER || it.type == PrincipalType.ROLE }
-                .toSet()
+                .filter {
+                    //There are some non-existing principals with permissions
+                    val principalExists = securePrincipalsManager.principalExists(it)
```
katherinebernstein: This is a relatively heavy call since it executes a hazelcast predicate -- it might be easier to just try/catch on the lookup call that fails (or turn this into a bulk principalExists call)

geekbeast: :+1:

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -465,6 +478,7 @@ class AssemblerConnectionManager(
      * Build grant select sql statement for a given table and user with column level security.
      * If properties (columns) are left empty, it will grant select on whole table.
      */
+    @Throws(NoSuchElementException::class)
```
makosblade: Where exactly does this `NoSuchElementException` get thrown? I had thought it came from `stmt.addBatch` 🤔

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -430,32 +430,45 @@ class AssemblerConnectionManager(
...
+                    logger.error("Principal $principal does not exists but has $permissionsToCheck permission on entity set ${entitySet.id}")
+                }
             }
             stmt.executeBatch()
         }
     }
 
     fun grantSelectForEdges(stmt: Statement, tableName: String, entitySetIds: Set<UUID>): IntArray {
-        val permissions = EnumSet.of(Permission.READ)
+        val permissionsToCheck = EdmAuthorizationHelper.READ_PERMISSION
```
makosblade: Not necessary for this PR but I think this block(450-461) can be moved all the way up to `MaterializedEdgesProcessor` since its logic is independent of the work being done in this class to build and execute this SQL. That way we also don't have to wait for network calls + exceptions for each principal during the `.foreach` and can just pass `authorizedPrincipals` in from `MaterializedEdgesProcessor`

makosblade: -- though I don't remember the particular restrictions on what can/can't be done inside of `EntryProcessor`s

you-leee: Good idea! Will do!

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
you-leee: It's coming from the HazelcastPrincipalService.lookupRole -> Optional.get()
But at this point, the principal should exist and should be a Role

katherinebernstein: Shouldn't it be catching these exceptions now though? So theoretically they won't ever get thrown

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
geekbeast: Should be fine as long as it is offloadable and read only iirc.

PR #558, opened 2019-07-11
LATTICE-2110 fix delete entities queries

you-leee:


src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -549,42 +549,65 @@ class PostgresEntityDataQueryService(
...
+                }
+            }
 
-                        val ps = it.prepareStatement(deletePropertiesOfEntities(entitySetId, property.key))
-                        val arr = PostgresArrays.createUuidArray(it, entityKeyIds)
-                        ps.setArray(1, arr)
+            connection.autoCommit = false
+            val partitions = partitionsInfo.partitions.toList()
+            val propertyTypesArr = PostgresArrays.createUuidArray(connection, authorizedPropertyTypes.keys)
+            entityKeyIds
```
katherinebernstein: it might be worth moving the group by outside of the connection just to avoid keeping it open for longer than absolutely necessary, but it would only matter in the case where there's a ton of entities being passed in

geekbeast: :+1: please file ticket going to merge for now

PR #563, opened 2019-07-11
LATTICE-2113 group outside of connection

you-leee:


src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -557,43 +556,56 @@ class PostgresEntityDataQueryService(
...
-                        val count = ps.executeUpdate()
-                        connection.commit()
-                        count
-                    }.sum()
-        }
 
-        return WriteEvent(System.currentTimeMillis(), numUpdates)
+            val idsArr = PostgresArrays.createUuidArray(connection, entities)
+
+            // Acquire entity key id locks
```
makosblade: Just curious about postgres things here: we don't want to turn this into a batch call because we need to lock all these rows right?

you-leee: I think, we also had the same problem , if it were a batch call, because of the connection timeout possibility

PR #579, opened 2019-07-16
Bugfix/lattice 2115 id constants entity set

you-leee:


src/main/kotlin/com/openlattice/assembler/AssemblerQueryService.kt
```diff
@@ -47,7 +47,7 @@ class AssemblerQueryService(private val edmService: EdmManager) {
                           srcAggregates: Map<String, List<AggregationType>>, edgeAggregates: Map<String, List<AggregationType>>, dstAggregates: Map<String, List<AggregationType>>,
                           calculations: Set<Calculation>,
                           srcFilters: Map<String, List<Filter>>, edgeFilters: Map<String, List<Filter>>, dstFilters: Map<String, List<Filter>>
-    ): Iterable<Map<String, Any>> {
+    ): Iterable<Map<String, Any?>> {
```
katherinebernstein: Does this ever return null?

you-leee: Grouping (and selected) values can be null, if I remember right. There is a section in the mapper of the PostgresIterable returned where if the grouping column has no value, the getArray will return null

katherinebernstein: okay -- I think this will require datastore updates then

you-leee: Yeah, you'r right. I will make a branch there to to get it compile

PR #580, opened 2019-07-17
"order" -John Bercow

katherinebernstein:


src/main/kotlin/com/openlattice/auditing/AuditInitializationTask.kt
```diff
@@ -56,7 +57,8 @@ class AuditInitializationTask(
                 UsersAndRolesInitializationTask::class.java,
                 OrganizationsInitializationTask::class.java,
                 Auth0SyncInitializationTask::class.java,
-                EdmSyncInitializerTask::class.java
+                EdmSyncInitializerTask::class.java,
+                IdConstantsReservationTask::class.java
```
you-leee: IdConstantsReservationTask is running already after EdmSyncInitializerTask, so I think  EdmSyncInitializerTask can be removed (unless we want list all for clarity)

katherinebernstein: probably safe to remove but there seems to be some badness happening with lazy evaluation that we should look into at some point so I'd rather leave it for now just to be safe

katherinebernstein: *evaluation of task order

you-leee: Ah, okay! 👍

PR #583, opened 2019-07-19
LATTICE-2126 move and fix buildWithClause to use ids table

you-leee:


you-leee:
Cool, I just moved the old version here. Will make the fix!

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
```diff
@@ -37,103 +86,121 @@ class IndexingMetadataManager(private val hds: HikariDataSource) {
...
-fun updateLastIndexSql(linking: Boolean): String {
-    val idColumn = if (linking) LINKING_ID.name else ID.name
-    val filterLinkingIds = if (linking) " AND ${LINKING_ID.name} IS NOT NULL " else ""
-    val entitiesClause = " AND ${ENTITY_SET_ID.name} = ? AND  $idColumn = ? $filterLinkingIds"
-    val withClause = buildWithClause(linking, entitiesClause)
-    // rather use id than linking_id in linking join
-    val joinClause = joinClause(IDS.name, FILTERED_ENTITY_KEY_IDS, entityKeyIdColumnsList)
-    val updateColumn = if (linking) LAST_LINK_INDEX.name else LAST_INDEX.name
+private val updateLastIndexSql = "UPDATE ${IDS.name} SET ${LAST_INDEX.name} = ? " +
+        "WHERE ${VERSION.name} > 0 AND ${ENTITY_SET_ID.name} = ? AND ${ID.name} = ANY(?) AND ${PARTITION.name} = ?"
```
katherinebernstein: I don't think the version > 0 filter is necessary -- could potentially cause weirdness if the entity is deleted after it's indexed and before it's marked as indexed

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
```diff
@@ -1,34 +1,83 @@
...
+        return count
     }
 
-    fun markAsIndexed(entityKeyIdsWithLastWrite: Map<UUID, Map<UUID, OffsetDateTime>>, linking: Boolean): Int {
+    private fun markAsIndexed(
+            connection: Connection,
+            entitySetId: UUID,
+            entityKeyIds: List<UUID>,
+            partition: Int,
+            lastWrite: OffsetDateTime
```
katherinebernstein: Just for clarity it would be good to rename this `lastIndex`

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
```diff
@@ -1,34 +1,83 @@
...
+
+            connection.prepareStatement(updateLastLinkingIndexSql).use { stmt ->
+                linkingEntityKeyIdsWithLastWrite.forEach { (entitySetId, linkingEntities) ->
+                    val partitions = entitySetPartitions.getValue(entitySetId).partitions
+                    val partitionsArray = PostgresArrays.createIntArray(connection, partitions)
+
+                    linkingEntities.entries
+                            .groupBy { it.value }
+                            .mapValues { it.value.map { it.key } }
+                            .forEach { (lastWrite, linkingIds) ->
```
katherinebernstein: Is this param actually lastWrite or is it lastIndex?

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
you-leee: Yeah, you're right!

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
you-leee: We update last_index with the value of last_write. So here I would still use lastWrite, since it's the value of that

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
you-leee: 👍

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
katherinebernstein: ohhh that makes sense

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
```diff
@@ -37,103 +86,121 @@ class IndexingMetadataManager(private val hds: HikariDataSource) {
...
-fun markLinkingIdsAsNeedToBeIndexedSql(): String {
-    val linkingEntitiesClause = " AND ${LINKING_ID.name} IS NOT NULL AND ${LINKING_ID.name} = ANY(?) "
-    val withClause = buildWithClause(true, linkingEntitiesClause)
-    // rather use id than linking_id in join
-    val joinClause = joinClause(IDS.name, FILTERED_ENTITY_KEY_IDS, entityKeyIdColumnsList)
-
-    return "$withClause UPDATE ${IDS.name} SET ${LAST_LINK_INDEX.name} = '-infinity()' " +
-            "FROM $FILTERED_ENTITY_KEY_IDS WHERE ($joinClause) "
-}
+private val markLinkingIdsAsNeedToBeIndexedSql = "UPDATE ${IDS.name} SET ${LAST_LINK_INDEX.name} = '-infinity()' " +
```
geekbeast: Comment: version > 0 isn't necessary since marking a deleted entity as needing indexed won't have any effect and once if anything update the version it would qualify for needing indexed.

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
you-leee: We agreed on using last write instead of now.. just don't remembering why exactly.. "Load last write during indexing and once indexing is done, set last index to last write instead of index => now()"
Can we derive the partition from linking ids as well?

you-leee: FYI: we only use this call for a single linkingId right now

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
```diff
@@ -37,103 +86,121 @@ class IndexingMetadataManager(private val hds: HikariDataSource) {
     }
 
     fun markEntitySetsAsNeedsToBeIndexed(entitySetIds: Set<UUID>, linking: Boolean): Int {
+        val entitySetPartitions = partitionManager.getEntitySetsPartitionsInfo(entitySetIds).values
+                .flatMap { it.partitions }
+                .toSet()
+
         return hds.connection.use { connection ->
             val updateSql = markEntitySetsAsNeedsToBeIndexedSql(linking)
             connection.prepareStatement(updateSql).use { stmt ->
```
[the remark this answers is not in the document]
you-leee: Probably there will be a reasonable amount of partitions shared amongst the entity sets in the update, so the overall number of tasks will be smaller compared to do it 1 by 1

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
you-leee: Ahh.. totally confused old entity_key_ids table with new ids table with separate rows for linking entities. Will make the change!

PR #588, opened 2019-07-23
write ids to data

katherinebernstein:


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -65,15 +65,12 @@ fun buildPreparableFiltersClauseForLinkedEntities(
 ): Pair<String, Set<SqlBinder>> {
     val filtersClauses = buildPreparableFiltersClause(startIndex, propertyTypes, propertyTypeFilters)
     val filtersClause = filtersClauses.first
+
     val innerSql = selectEntitiesGroupedByIdAndPropertyTypeId(idsPresent = idsPresent, partitionsPresent = partitionsPresent)
 
-    val sql = joinPreparableFiltersSelectToIdsTable(
-            "$innerSql AND ${ORIGIN_ID.name} IS NOT NULL AND ${VERSION.name} > 0" +
+    val sql = "$innerSql AND ${ORIGIN_ID.name} IS NOT NULL AND ${VERSION.name} > 0" +
```
geekbeast: origin_id is not null seems redundant for entity sets that are not linking entity sets.

katherinebernstein: this is for linking entity sets

PR #589, opened 2019-07-23
get rid of ENTITY_ID

makosblade:
This PR:
- Removes all uses of `ENTITY_ID` except uses in SYNC_IDS table and mechanic migrations


makosblade:
Note: Still need to drop that column from the tables on staging/prod

makosblade:
And get rid of that index

PR #593, opened 2019-07-23
more commit

katherinebernstein:


katherinebernstein:
Based on the bug @you-leee  found it seems like autocommit hadn't been set to false somewhere so I assume there's places where autocommit isn't set back to true after

katherinebernstein:
If it's important that it's false and stays false we could update it more layers up though

makosblade:
👍  Would be nice to have an annotation that can check such things at compile time :(

katherinebernstein:
yeah :) or documentation :)

PR #595, opened 2019-07-24
Bugfix/lattice 2140 entity set sizes

you-leee:
This PR basically wraps the join on entity_sets (for linking entity set counts).
It would be much easier to query the data table by using the following query, but we don't have linking id changes ready on data table yet. So this can be merged in the meantime or we wait until those changes are finished.
`select entity_set_id,count(distinct id) from data where version > 0  group by entity_set_id;`


katherinebernstein:
for now we might want to rename the materialized view so that we can deploy this on staging without getting into a competition with prod

you-leee:
Ah, okay! 👍

PR #596, opened 2019-07-25
Bugfix/lattice 2142 update get linking ids by entity set ids

you-leee:


src/main/java/com/openlattice/edm/PostgresEdmManager.java
```diff
@@ -88,35 +92,53 @@ public PostgresEdmManager( HikariDataSource hds, PostgresTableManager ptm, Hazel
...
-                        " GROUP BY " + ENTITY_SET_ID.getName();
+    /**
+     * Queries the linking ids belonging to the entities in the requested entity sets.
+     *
+     * @param entitySetIds Set of ids of regular entity sets.
+     * @return Set of linking ids mapped by the requested entity set ids.
+     */
+    public PostgresIterable<Pair<UUID, Set<UUID>>> getLinkingIdsByEntitySetIds( Set<UUID> entitySetIds ) {
+        final var entitySetPartitions = partitionManager.getEntitySetsPartitionsInfo( entitySetIds ).values().stream()
+                .flatMap( partitionInfo -> partitionInfo.component1().stream() )
```
katherinebernstein: component1?

src/main/java/com/openlattice/edm/PostgresEdmManager.java
```diff
@@ -88,35 +92,53 @@ public PostgresEdmManager( HikariDataSource hds, PostgresTableManager ptm, Hazel
...
-                        " WHERE " + LINKING_ID.getName() + " IS NOT NULL AND " + ENTITY_SET_ID.getName()
-                        + " = ANY( ? ) " +
-                        " GROUP BY " + ENTITY_SET_ID.getName();
+    /**
+     * Queries the linking ids belonging to the entities in the requested entity sets.
+     *
+     * @param entitySetIds Set of ids of regular entity sets.
+     * @return Set of linking ids mapped by the requested entity set ids.
+     */
+    public PostgresIterable<Pair<UUID, Set<UUID>>> getLinkingIdsByEntitySetIds( Set<UUID> entitySetIds ) {
```
katherinebernstein: why do we need to load all linking ids for an entity set as a set? These sets could potentially be huge

src/main/java/com/openlattice/edm/PostgresEdmManager.java
you-leee: Yeah, getPartitions() would be more readable. https://kotlinlang.org/docs/reference/multi-declarations.html

src/main/java/com/openlattice/edm/PostgresEdmManager.java
you-leee: It was for indexing linking entities in elasticsearch, when we remove normal entity sets from linking entity set, but back then we indexed by entity set and needed to remove linking ids, which were no longer part of linking entity set and re-index linking ids, which were, just got these entities "removed" from it.

you-leee: So, since now linking entities are indexed without any data on their linking entity sets, only the normal entity set, where the data is coming from, this is probably not needed anymore. Will make a separate ticket (since this should have been changed earlier) for it, but push the changes to this branch.

katherinebernstein: Okay! In any case it would probably be safe to either do this in batches or iterate over the entityKeyIds instead of collecting them to a set so that memory doesn't explode

PR #599, opened 2019-07-26
Feature/lattice 2123 option to audit to postgres

katherinebernstein:


src/main/kotlin/com/openlattice/auditing/LocalAuditingService.kt
```diff
@@ -0,0 +1,104 @@
...
+
+        val auditingConfiguration = ares.auditingTypes
+
+        return if (auditingConfiguration.isAuditingInitialized()) {
+            events
+                    .groupBy { ares.getActiveAuditEntitySetIds(it.aclKey, it.eventType) }
+                    .filter { (auditEntitySetConfiguration, _) ->
+                        auditEntitySetConfiguration.auditRecordEntitySet != null
+                    }
+                    .map { (auditEntitySetConfiguration, entities) ->
```
you-leee: Really minor thing, but "entities" is a bit confusing for me, since they are filtered events (and they have a field, that is called entities)

katherinebernstein: +1, will update!

src/main/kotlin/com/openlattice/auditing/AuditingProfiles.kt
```diff
@@ -0,0 +1,13 @@
...
+
+class AuditingProfiles {
+
+    companion object {
+
+        @JvmField
+        val LOCAL_AUDITING_PROFILE = "auditlocal"
+
+        @JvmField
+        val LOCAL_AWS_AUDITING_PROFILE = "auditlocalaws"
```
makosblade: Am confused about what this would be auditing; AWS? or local machine?

src/main/kotlin/com/openlattice/auditing/LocalAuditingService.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    fun recordEvent(event: AuditableEvent): Int {
+        return recordEvents(listOf(event))
+    }
+
+    override fun recordEvents(events: List<AuditableEvent>): Int {
+
+        val auditingConfiguration = ares.auditingTypes
+
+        return if (auditingConfiguration.isAuditingInitialized()) {
```
makosblade: Any chance you'd be willing to this instead:
```
if ( !auditingConfiguration.isAuditingInitialized() ){
    return 0
}
// CODES
// return things
```

src/main/kotlin/com/openlattice/auditing/AuditingProfiles.kt
katherinebernstein: it's for running locally while auditing to the aws test bucket (instead of to the local machine) -- I was following the `medialocal`, `medialocalaws` pattern but can change it if it's confusing!

src/main/kotlin/com/openlattice/auditing/LocalAuditingService.kt
katherinebernstein: there is
a chance

makosblade: phew

src/main/kotlin/com/openlattice/auditing/AuditingProfiles.kt
makosblade: I think it's fine for now but I find the `*local` vs `*localaws` pattern to be unintuitive. I think it makes more sense to call them `*local` and `*aws` for running against a `local` target and an `aws` target respectively

PR #606, opened 2019-08-06
IdCatchupEntryProcessor should use IDS table

katherinebernstein:


src/main/java/com/openlattice/ids/IdCatchupEntryProcessor.java
```diff
@@ -73,7 +73,7 @@ public Void process( Entry<Integer, Range> entry ) {
 
     public PreparedStatement prepareExistQuery( Connection connection ) throws SQLException {
         return connection.prepareStatement(
-                "SELECT count(*) from " + PostgresTable.ENTITY_KEY_IDS.getName() +
+                "SELECT count(*) from " + PostgresTable.IDS.getName() +
                         " WHERE " + PostgresColumn.ID_VALUE.getName() + " = ?" );
```
you-leee: Maybe add also filter for partition?

katherinebernstein: 👍👍👍

katherinebernstein: Although actually we'd need to look up the entity set ids in order to load the partitions, which would require doing this same query anyway

you-leee: Yeah, right!

katherinebernstein:
Yeah, I didn't touch it since whoever wrote it left that comment but kept the checkNotNull anyway and I'm not sure why...

PR #612, opened 2019-08-15
Bugfix/lattice 2154 materialized entityset org update

you-leee:


src/main/kotlin/com/openlattice/assembler/Assembler.kt
```diff
@@ -495,15 +520,39 @@ class Assembler(
...
+    private fun isEntitySetMaterialized(entitySetId: UUID): Boolean {
+        return materializedEntitySets
+                .keySet(entitySetIdPredicate(entitySetId))
+                .isNotEmpty()
+    }
+
+    /**
+     * Returns true, if the entity set is materialized in the organization
+     */
+    private fun isEntitySetMaterialized(organizationId: UUID, entitySetId: UUID): Boolean {
```
katherinebernstein: Is this one ever used?

you-leee: Nope, will delete it!

PR #616, opened 2019-08-27
set default ORIGIN_ID to IdConstants.EMPTY_UUID

makosblade:
This PR:
- Sets ORIGIN_ID to have a default value from IdConstants
- Is the code that gets deployed after https://github.com/openlattice/mechanic/pull/39 and before https://github.com/openlattice/mechanic/pull/38

### DIFF
--- a/src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
+++ b/src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
@@ -69,7 +69,7 @@ fun buildPreparableFiltersSqlForLinkedEntities(
 
     val innerSql = selectEntitiesGroupedByIdAndPropertyTypeId(
             idsPresent = idsPresent, partitionsPresent = partitionsPresent, selectOriginIds = selectOriginIds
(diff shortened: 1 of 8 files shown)


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -68,7 +69,7 @@ fun buildPreparableFiltersSqlForLinkedEntities(
 
     val innerSql = selectEntitiesGroupedByIdAndPropertyTypeId(
             idsPresent = idsPresent, partitionsPresent = partitionsPresent, selectOriginIds = selectOriginIds
-    ) + " AND ${ORIGIN_ID.name} IS NOT NULL " + filtersClause + GROUP_BY_ESID_EKID_PART_PTID
+    ) + " AND ${ORIGIN_ID.name} IS NOT NULL AND ${ORIGIN_ID.name} IS NOT '${IdConstants.EMPTY_ORIGIN_ID.id}' " + filtersClause + GROUP_BY_ESID_EKID_PART_PTID
```
katherinebernstein: `AND ${ORIGIN_ID.name} IS NOT NULL` seems removable

makosblade: Yes, given the doing-migration-to-non-null-and-setting-empty-uuid beforehand 👍

makosblade:
Depends on https://github.com/openlattice/api/pull/348

PR #617, opened 2019-08-29
Feature/lattice 1913 organization principal permission change

you-leee:
This PR:

- keeps materialized entityset in synch with materialization permission changes by marking them unsynch and after picking them up with a background processor
- only calls refresh on materialized view, when entity data changed (and does not re-create materialized view with authorized properties for org principal)


src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -416,6 +406,26 @@ class AssemblerConnectionManager(
...
+            entitySetId: UUID,
+            materializablePropertyTypes: Map<UUID, PropertyType>) {
+        val selectColumns = getSelectColumnsForMaterializedView(materializablePropertyTypes.values)
+                .joinToString(",")
+
+        val sql = "SELECT $selectColumns FROM $PRODUCTION_FOREIGN_SCHEMA.${entitySetIdTableName(entitySetId)}"
+
+        val dropMaterializedEntitySet = "DROP MATERIALIZED VIEW IF EXISTS $tableName"
+        val createMaterializedViewSql = "CREATE MATERIALIZED VIEW IF NOT EXISTS $tableName AS $sql"
+
```
pckeyes: is the "IF NOT EXISTS" necessary in createMaterializedViewSql? if the table did exist before it gets dropped and no longer exists. if it didn't exist before, it still doesn't exist.

you-leee: Nope, nice catch!

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -521,6 +541,8 @@ class AssemblerConnectionManager(
                 stmt.executeUpdate("ALTER MATERIALIZED VIEW IF EXISTS $oldTableName RENAME TO $newTableName")
             }
         }
+        logger.info("Renamed materialized view of entity set with old name $oldName to name name $newName in " +
+                "organization $organizationId")
```
pckeyes: think there's an extra "name" in here

you-leee: yeah, that should be "new", thanks!

src/main/kotlin/com/openlattice/assembler/processors/RefreshMaterializedEntitySetProcessor.kt
```diff
@@ -54,9 +46,8 @@ data class RefreshMaterializedEntitySetProcessor(
             throw IllegalStateException("Encountered null materialized entity set while trying to refresh data in " +
                     "materialized view for entity set ${entitySet.id} in organization $organizationId.")
         } else {
-            acm?.refreshEntitySet(
-                    organizationId, entitySet, materializablePropertyTypes, authorizedPropertyTypesOfPrincipals
-            ) ?: throw IllegalStateException(NOT_INITIALIZED)
+            acm?.refreshEntitySet(organizationId, entitySet)
+                    ?: throw IllegalStateException(AssemblerConnectionManagerDependent.NOT_INITIALIZED)
```
pckeyes: each time I see this replace the repeatedly used constant I feel happy :)

src/main/kotlin/com/openlattice/assembler/tasks/MaterializePermissionSyncTask.kt
```diff
@@ -0,0 +1,118 @@
...
+                "AND array[${ENTITY_SET_ID.name}] = ${ACL_KEY.name})"
+
+class MaterializePermissionSyncTask : HazelcastFixedRateTask<MaterializedEntitySetsDependencies> {
+
+    override fun getInitialDelay(): Long {
+        return 30000L // wait until AssemblerConnectionManager can be initialized
+    }
+
+    override fun getPeriod(): Long {
+        return 3600L // 1 min
```
pckeyes: 60000 is a minute in ms, 3600 is 3.6 seconds, i think you did 60* 60 instead of 60 * 1000 lol

you-leee: 🤦‍♀ Nice catch!

src/main/java/com/openlattice/authorization/HazelcastAuthorizationService.java
```diff
@@ -83,6 +84,39 @@ public void setSecurableObjectType( AclKey aclKey, SecurableObjectType objectTyp
...
+    }
+
+    private void signalMaterializationPermissionChangeBulk(
+            Set<AceKey> aceKeys,
+            EnumSet<Permission> permissions,
+            SecurableObjectType securableObjectType
+    ) {
+        // if there is a change in materialization permission for a property type for an organization principal,
+        // we need to flag it
+        if ( permissions.contains( Permission.MATERIALIZE )
```
katherinebernstein: Do you know if this is normally called on a huge number of aceKeys at once? It would be nice for this to just call through to `signalMaterializationPermissionChange` instead of duplicating the logic here

src/main/kotlin/com/openlattice/assembler/processors/UpdateMaterializedEntitySetProcessor.kt
```diff
@@ -0,0 +1,82 @@
...
+    @Transient
+    private var acm: AssemblerConnectionManager? = null
+
+    override fun process(entry: MutableMap.MutableEntry<EntitySetAssemblyKey, MaterializedEntitySet?>): Void? {
+        val organizationId = entry.key.organizationId
+        val materializedEntitySet = entry.value
+        if (materializedEntitySet == null) {
+            throw IllegalStateException("Encountered null materialized entity set while trying to update " +
+                    "materialized view for entity set ${entitySet.id} in organization $organizationId.")
+        } else {
```
katherinebernstein: You can unnest all this from the `else`

src/main/kotlin/com/openlattice/assembler/tasks/MaterializePermissionSyncTask.kt
```diff
@@ -0,0 +1,118 @@
...
+import com.openlattice.tasks.Task
+import java.sql.ResultSet
+import java.util.*
+import java.util.concurrent.TimeUnit
+import java.util.function.Function
+import java.util.function.Supplier
+
+
+private val selectUnauthorizedEntitySetAssemblies =
+        "SELECT ${ENTITY_SET_ID.name}, ${ORGANIZATION_ID.name} FROM ${MATERIALIZED_ENTITY_SETS.name} " +
```
katherinebernstein: Does this not need to be computed for each organization? e.g. if an entity set is materialized by two organizations, and one of those organizations has lost its MATERIALIZE permissions, it should still be marked as needing removal from one organization's database? (or does this delete not affect individual organization databases?)

you-leee: For entity set permission changes in materialization, we don't need to mark them, because: 

- if permissions get's added to an organizations, there is no action required (it just means, that owners can materialize the entity set if they choose so) as opposed to, when materialization permission get's added to a property type, that is part of a materialized entity set, we need to re-create the materialized view with the new column
- if permission get's removed, we don't need to mark it for entity sets, because they can be picked up with this query and get deleted from that organizations database (and from mapstore)

src/main/kotlin/com/openlattice/assembler/processors/UpdateMaterializedEntitySetProcessor.kt
you-leee: oh, yeah, thanks!

src/main/kotlin/com/openlattice/assembler/tasks/MaterializePermissionSyncTask.kt
katherinebernstein: Right, but if an entity set is materialized in two organizations' databases, and only one of them loses its MATERIALIZE permissions, then it won't get picked up by this query, right?

you-leee: Ohh, you're right! There would need to be a join on organization too, but that's not possible from these 2 tables. I will mark entity sets as well then.

src/main/java/com/openlattice/authorization/HazelcastAuthorizationService.java
you-leee: I'm not sure, there are not metrics yet on `PermissionsController.updateAcls`.

src/main/kotlin/com/openlattice/assembler/tasks/MaterializePermissionSyncTask.kt
```diff
@@ -0,0 +1,102 @@
+/*
```
makosblade: This doesn't appear to be used 🤔

src/main/kotlin/com/openlattice/assembler/Assembler.kt
```diff
@@ -330,6 +362,33 @@ class Assembler(
...
+                    entitySetAssemblyKey,
+                    UpdateMaterializedEntitySetProcessor(entitySets.getValue(entitySetId), materializablePropertyTypes)
+                            .init(acm)
+            )
+        }
+
+    }
+
     private fun materializeEdges(organizationId: UUID) {
         // materialize edges by including all the materialized entity sets in organization
```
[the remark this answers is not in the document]
you-leee: As far as I remember, we will have a postgres role for membership in an organization and basically that will be used to grant select on edges. That would simplify things a lot :)

src/main/kotlin/com/openlattice/assembler/tasks/MaterializePermissionSyncTask.kt
you-leee: It is used here: https://github.com/openlattice/conductor/pull/86/files

src/main/java/com/openlattice/authorization/HazelcastAuthorizationService.java
geekbeast: This should only get called for # property types x # of organizations changed, which should generally be one. I think Katherine's comment is valid and we should consider not firing off a bunch of events, especially if they can bulked more effectively (each event blocks a thread for the duration of processing).

you-leee: I think Katherine is implying, that we should reuse the code if this call is used mostly with single (or just few) permission updates. And I guess, you would change the call rather to go through the `updates` (`SetMultimap<AceValue, AceKey>`) one more time and only fire 1 event per 1 organization principal, right?

PR #611, opened 2019-08-08
Feature/lattice 2159 linker should write entities into the data table

makosblade:
This PR:
- Updates write path to
  - propagate data table property changes to their equivalent linking rows
  - tombstone linked rows when regular rows are tombstoned
  - adds queries to support ^
  - adds queries used by linking query service to add and tombstone linking rows
  - fixes incorrect preparedstatement types where possible

PR #619, opened 2019-09-05
Add ReadOnly, EntryProcessors for getting EntityType from EntitySet and Properties from EntityType

makosblade:
This PR:
- Creates ReadOnly EntryProcessors for reducing the overhead of authorization calls


makosblade:
Depends on https://github.com/openlattice/conductor-client/pull/611

makosblade:
Curious about your thoughts on this approach vs using a Projection @geekbeast
Since they're `@ReadOnly` EntryProcessors, they don't do any partition locking and are isolated to only the target partition vs projection having to run a `Predicates.equal("__key", key)` predicate across all partitions

geekbeast:
> Curious about your thoughts on this approach vs using a Projection @geekbeast
> Since they're `@ReadOnly` EntryProcessors, they don't do any partition locking and are isolated to only the target partition vs projection having to run a `Predicates.equal("__key", key)` predicate across all partitions

I like it!

src/main/kotlin/com/openlattice/edm/processors/GetEntityTypeFromEntitySetEntryProcessor.kt
```diff
@@ -0,0 +1,13 @@
+package com.openlattice.edm.processors
+
+import com.hazelcast.core.ReadOnly
+import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor
+import com.openlattice.edm.EntitySet
+import java.util.*
+
+class GetEntityTypeFromEntitySetEntryProcessor : AbstractRhizomeEntryProcessor<UUID, EntitySet, Optional<UUID>>(), ReadOnly {
+    override fun process(entry: MutableMap.MutableEntry<UUID, EntitySet>?): Optional<UUID> {
```
katherinebernstein: Pretty sure the generated signature is wrong and entry will never be null

makosblade: It can't be null given the way we've used it but used in other contexts, it could be null :'(

geekbeast: Only the value can be null. The entry itself will never be null (at least according Hazelcast Docs). The signature be MutableEntry<UUID, EntitySet?> but I'm not sure there's a way for kotlin to know that.

makosblade: EntryProcessor:88
"     * The entry's value will be {@code null}, if the entry for the key doesn't exist."
Which, given the way we use it, should not happen but totally could happen

makosblade: Wait what
😕

makosblade: Okay, I see. Am fixing

makosblade: Related: https://github.com/geekbeast/rhizome/pull/282
and further related: #628

makosblade:
Hmm so super doesn't depend on #611 
Not sure why I put that there. . . 
In any case, this is ready to merge once the various data expiration PRs merge and fix the build

src/main/kotlin/com/openlattice/hazelcast/serializers/NoOpSelfRegisteringStreamSerializer.kt
```diff
@@ -0,0 +1,16 @@
+package com.openlattice.hazelcast.serializers
+
+import com.hazelcast.nio.ObjectDataOutput
+import com.kryptnostic.rhizome.pods.hazelcast.SelfRegisteringStreamSerializer
+import com.openlattice.hazelcast.StreamSerializerTypeIds
+
+abstract class NoOpSelfRegisteringStreamSerializer<T>: SelfRegisteringStreamSerializer<T> {
+    override fun getTypeId(): Int {
+        return StreamSerializerTypeIds.NO_SERIALIZATION_ENTRY_PROCESSOR.ordinal
```
katherinebernstein: looks like this is making hazelcast unhappy (since multiple stream serializers are getting registered for the same type id)

geekbeast: was this fixed?

PR #621, opened 2019-09-05
Feature/lattice 1756 data expiration

pckeyes:
New/updated stream serializers and other Hazelcast additions for new data expiration field in entity sets class, as well as added columns to entity sets postgres table. Also, logic for using metadata update api for setting/changing data expiration policy on an entity set.


src/main/java/com/openlattice/postgres/PostgresTable.java
```diff
@@ -144,7 +144,10 @@
                             ORGANIZATION_ID,
                             ENTITY_SET_FLAGS,
                             PARTITIONS,
-                            PARTITIONS_VERSION );
+                            PARTITIONS_VERSION,
```
makosblade: will need a mechanic PR for migration 👍

src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -990,7 +991,7 @@ public void updateEntitySetMetadata( UUID entitySetId, MetadataUpdate update ) {
             }
         }
 
-        final var newEntitySet = ( EntitySet ) entitySets.executeOnKey(
```
makosblade: 😬

src/main/java/com/openlattice/postgres/PostgresTable.java
pckeyes: Thanks I will add one!

src/main/java/com/openlattice/datastore/services/EdmService.java
pckeyes: Is this a bad face?

makosblade: haha sorry. We don't have clear formatting standards but I prefer the extra spacing style. Super don't need to change this at all but I wanted to flag it cause we should settle on some formatting preferences as a team

pckeyes: Oh I see! I think I submitted this before Katherine showed me the formatting style we use. I will update and push that

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -874,6 +882,35 @@ class PostgresEntityDataQueryService(
 
         return WriteEvent(tombstoneVersion, numUpdated)
     }
+
+    fun getExpiringEntitiesFromEntitySet(entitySetId: UUID, sqlParams: Triple<String, Any, Int>, deleteType: DeleteType) : BasePostgresIterable<UUID> {
```
you-leee: Just a minor thing, but I would put the `sqlParams` into separate parameters

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -656,6 +669,26 @@ public static int partitionVersions( ResultSet rs ) throws SQLException {
         return PostgresArrays.getIntArray( rs, PARTITIONS_FIELD );
     }
 
+    public static Long timeToExpiration( ResultSet rs ) throws SQLException {
+        return (Long) rs.getObject( TIME_TO_EXPIRATION_FIELD );
```
katherinebernstein: you should be able to do `rs.getLong( TIME_TO_EXPIRATION_FIELD )`

src/main/java/com/openlattice/postgres/mapstores/EntitySetMapstore.java
```diff
@@ -49,18 +46,48 @@ public EntitySetMapstore( HikariDataSource hds ) {
         ps.setArray( 9, flags );
         ps.setArray( 10, partitions );
         ps.setInt( 11, value.getPartitionsVersion() );
+        if ( value.getExpiration() == null ) {
+            ps.setNull( 12, Types.NULL );
+            ps.setNull( 13, Types.NULL );
+            ps.setNull( 14, Types.NULL );
+            ps.setNull( 15, Types.NULL );
+        } else {
+            ps.setObject( 12, value.getExpiration().getTimeToExpiration() );
```
katherinebernstein: should be `ps.setLong`

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -635,6 +636,17 @@ public static EntitySet entitySet( ResultSet rs ) throws SQLException {
         final var flags = entitySetFlags( rs );
         final var partitions = partitions( rs );
         final var partitionVersion = partitionVersions( rs );
+        final var timeToExpiration = timeToExpiration( rs );
```
katherinebernstein: probably good to break this out into a `public static DataExpiration dataExpiration( ResultSet rs` that's called from `entitySet` for readability / standardization / in case we ever use this object anywhere else

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -874,6 +883,35 @@ class PostgresEntityDataQueryService(
...
+
+    private fun getExpiringEntitiesQuery(entitySetId: UUID, expirationBaseColumn: String, deleteType: DeleteType, partitions: String) : String {
+        var ignoredClearedEntitiesClause = ""
+        if (deleteType == DeleteType.Soft) {
+            ignoredClearedEntitiesClause = "AND ${VERSION.name} >= 0 "
+        }
+        return "SELECT ${ID.name} FROM ${PostgresTable.DATA.name} " +
+                "WHERE ${ENTITY_SET_ID.name} = '$entitySetId' " +
+                "AND ${PARTITION.name} IN $partitions " +
+                "AND ${PARTITIONS_VERSION.name} = ? " +
```
katherinebernstein: just wondering why some of these are parameterized and some values are passed in and written to the query string?

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
pckeyes: rs.getLong will return 0 if the value in postgres is null, which it may be if there is no expiration policy. getObject just returns a null. Is it better to use getLong and do a check for 0? teeechnically the time to expiration is allowed to be 0.

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
pckeyes: The last variable is the only one that is of type Any and so I have to pass with a sql type. I can make them all parameters instead if that's preferable

geekbeast: Use of parameters is preferred if it's a prepared statement. SQL injection isn't a risk here, but you lose a big advantage of prepared statements due to high cardinality of possible values for entitySetId and partitions (that is each combination of passed in values generates a different prepared statement and execution plan).

pckeyes: gotcha, will fix

src/main/java/com/openlattice/hazelcast/serializers/EntitySetStreamSerializer.java
```diff
@@ -69,6 +67,12 @@ public static void serialize( ObjectDataOutput out, EntitySet object ) throws IO
 
         out.writeInt( object.getPartitionsVersion() );
 
+        if ( object.getExpiration() != null ) {
+            out.writeBoolean( true );
+            DataExpirationStreamSerializer.serialize( out, object.getExpiration() );
+        } else {
+            out.writeBoolean( false );
+        }
```
makosblade: Can use `KotlinOptionalStreamSerializers.serializeMaybeValue` here

src/main/java/com/openlattice/hazelcast/serializers/EntitySetStreamSerializer.java
```diff
@@ -101,6 +105,14 @@ public static EntitySet deserialize( ObjectDataInput in ) throws IOException {
 
         int partitionsVersion = in.readInt();
 
+        DataExpiration expiration;
+        boolean hasExpiration = in.readBoolean();
+        if ( hasExpiration ) {
+            expiration = DataExpirationStreamSerializer.deserialize( in );
+        } else {
+            expiration = null;
+        }
```
makosblade: can use `KotlinOptionalStreamSerializers.deserializeMaybeValue` here

src/main/java/com/openlattice/hazelcast/serializers/MetadataUpdateStreamSerializer.java
```diff
@@ -59,6 +62,13 @@ public static void serialize( ObjectDataOutput out, MetadataUpdate object ) thro
         OptionalStreamSerializers.serialize( out,
                 object.getPartitions(),
                 ( output, elem ) -> output.writeIntArray( elem.stream().mapToInt( e -> e ).toArray() ) );
+        if ( object.getDataExpiration().isPresent() ) {
+            out.writeBoolean( true );
+            OptionalStreamSerializers
+                    .serialize( out, object.getDataExpiration(), DataExpirationStreamSerializer::serialize );
+        } else {
+            out.writeBoolean( false );
+        }
```
makosblade: Can use `KotlinOptionalStreamSerializers.serializeOptionalValue` here

src/main/java/com/openlattice/hazelcast/serializers/MetadataUpdateStreamSerializer.java
```diff
@@ -77,6 +87,14 @@ public static MetadataUpdate deserialize( ObjectDataInput in ) throws IOExceptio
                 .deserialize( in, UUIDStreamSerializer::deserialize );
         Optional<LinkedHashSet<Integer>> partitions = OptionalStreamSerializers
                 .deserialize( in, input -> toLinkedHashSet( input.readIntArray() ) );
+        Optional<DataExpiration> dataExpiration;
+        boolean hasExpiration = in.readBoolean();
+        if ( hasExpiration ) {
+            dataExpiration = OptionalStreamSerializers.deserialize( in, DataExpirationStreamSerializer::deserialize );
+        } else {
+            dataExpiration = Optional.empty();
+        }
```
makosblade: Can use `KotlinOptionalStreamSerializers.deserializeOptional` here

src/main/kotlin/com/openlattice/hazelcast/processors/RemoveDataExpirationPolicyProcessor.kt
```diff
@@ -0,0 +1,31 @@
...
+import java.util.UUID
+
+class RemoveDataExpirationPolicyProcessor() :
+        AbstractRhizomeEntryProcessor<UUID, EntitySet, EntitySet>() {
+
+    companion object {
+        private const val serialVersionUID = -6602384557982347L //TODO ????
+    }
+
+    override fun process(entry: MutableMap.MutableEntry<UUID, EntitySet>?): EntitySet {
```
makosblade: Update signature to:
`override fun process(entry: MutableMap.MutableEntry<UUID, EntitySet?>): EntitySet {`
specifically: 
`Entry<UUID, EntitySet?>` specifically

src/main/kotlin/com/openlattice/hazelcast/processors/RemoveDataExpirationPolicyProcessor.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor
+import com.openlattice.edm.EntitySet
+import java.util.UUID
+
+class RemoveDataExpirationPolicyProcessor() :
+        AbstractRhizomeEntryProcessor<UUID, EntitySet, EntitySet>() {
+
+    companion object {
+        private const val serialVersionUID = -6602384557982347L //TODO ????
```
makosblade: `TODO` ?

src/main/kotlin/com/openlattice/hazelcast/serializers/DataExpirationStreamSerializer.kt
```diff
@@ -0,0 +1,55 @@
...
+    companion object {
+        private val expirationTypes = ExpirationBase.values()
+        private val deleteTypes = DeleteType.values()
+
+        @JvmStatic
+        fun serialize(out: ObjectDataOutput, obj: DataExpiration) {
+            out.writeLong(obj.timeToExpiration)
+            out.writeInt(obj.expirationBase.ordinal)
+            out.writeInt(obj.deleteType.ordinal)
+            OptionalStreamSerializers.serialize(out, obj.startDateProperty, UUIDStreamSerializer::serialize)
```
makosblade: Can use the Kotlin ones mentioned above here as well. Unsure if one is better but the Kotlin ones use Kotlin (duh 😆)

src/main/java/com/openlattice/postgres/mapstores/EntitySetMapstore.java
```diff
@@ -49,18 +46,48 @@ public EntitySetMapstore( HikariDataSource hds ) {
...
+        ps.setInt( 25, value.getPartitionsVersion() );
+        if ( value.getExpiration() == null ) {
+            ps.setNull( 26, Types.NULL );
+            ps.setNull( 27, Types.NULL );
+            ps.setNull( 28, Types.NULL );
+            ps.setNull( 29, Types.NULL );
+        } else {
+            ps.setObject( 26, value.getExpiration().getTimeToExpiration() );
+            ps.setString( 27, value.getExpiration().getExpirationBase().toString() );
+            ps.setString( 28, value.getExpiration().getDeleteType().toString() );
```
makosblade: Probably not a big deal but I feel like there's some more optimal way of storing enums in Postgres. I'm also unsure of this, just something to think about as Strings are large to store relative to limited number sets

src/main/java/com/openlattice/postgres/mapstores/EntitySetMapstore.java
```diff
@@ -49,18 +46,48 @@ public EntitySetMapstore( HikariDataSource hds ) {
...
+        ps.setArray( 23, flags );
+        ps.setArray( 24, partitions );
+        ps.setInt( 25, value.getPartitionsVersion() );
+        if ( value.getExpiration() == null ) {
+            ps.setNull( 26, Types.NULL );
+            ps.setNull( 27, Types.NULL );
+            ps.setNull( 28, Types.NULL );
+            ps.setNull( 29, Types.NULL );
+        } else {
+            ps.setObject( 26, value.getExpiration().getTimeToExpiration() );
```
makosblade: Should be 
`ps.setLong(`

src/main/java/com/openlattice/postgres/mapstores/EntitySetMapstore.java
makosblade: Should be `ps.setLong(`

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -615,4 +624,48 @@ open class DataGraphService(
 //        }.stream()
         return Stream.empty()
     }
+
+    override fun getExpiringEntitiesFromEntitySet(entitySetId: UUID, expirationPolicy: DataExpiration,
+                                                  dateTime: OffsetDateTime, deleteType: DeleteType,
+                                                  expirationPT: Optional<PropertyType>): BasePostgresIterable<UUID> {
```
makosblade: either this `expirationPT` or `expirationPropertyType` in `DataGraphManager` should be renamed so they are both the same

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -615,4 +624,48 @@ open class DataGraphService(
...
+        val sqlParams = getSqlParameters(expirationPolicy, dateTime, expirationPT)
+        val expirationBaseColumn = sqlParams.first
+        val formattedDateMinusTTE = sqlParams.second
+        val sqlFormat = sqlParams.third
+        return eds.getExpiringEntitiesFromEntitySet(entitySetId, expirationBaseColumn, formattedDateMinusTTE,
```
makosblade: I would consider just updating this call to take a triple `¯\_(ツ)_/¯`

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -615,4 +624,48 @@ open class DataGraphService(
...
+                sqlFormat, deleteType)
+    }
+
+    private fun getSqlParameters(expirationPolicy: DataExpiration, dateTime: OffsetDateTime, expirationPT: Optional<PropertyType>): Triple<String, Any, Int> {
+        val expirationBaseColumn: String
+        val formattedDateMinusTTE: Any
```
makosblade: This type frightens me 😬

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -928,6 +937,41 @@ class PostgresEntityDataQueryService(
...
+                    stmt.setArray(2, partitions)
+                    stmt.setInt(3, partitionVersion)
+                    stmt.setObject(4, IdConstants.ID_ID.id)
+                    stmt.setObject(5, formattedDateMinusTTE, sqlFormat)
+                }
+        ) { rs -> ResultSetAdapters.id(rs)}
+    }
+
+    private fun getExpiringEntitiesQuery(expirationBaseColumn: String, deleteType: DeleteType) : String {
+        var ignoredClearedEntitiesClause = ""
```
makosblade: can `val` this by setting a type (pretty sure that will default it to `""`)

src/main/java/com/openlattice/hazelcast/serializers/EntitySetStreamSerializer.java
pckeyes: oooooh fancy new thing?

makosblade: eehhh fancy, not-so-much. New, yes!

src/main/java/com/openlattice/postgres/mapstores/EntitySetMapstore.java
pckeyes: oh interesting. I think I've seen other enums stored in this way, but doesn't mean we have to do it that way! Should i store the ordinal instead?

src/main/kotlin/com/openlattice/data/DataGraphService.kt
pckeyes: lol @you-leee asked me to change it to separate params to make it more readable. it originally was just the triple.

who wins this debate??

src/main/kotlin/com/openlattice/data/DataGraphService.kt
pckeyes: Its why we have the sql type as well below tho!

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
pckeyes: oh nice

pckeyes: actually seems like not :(

src/main/kotlin/com/openlattice/data/DataGraphService.kt
you-leee: I think the triple is less readable afterwards used in the code in `getExpiringEntitiesFromEntitySet`

src/main/kotlin/com/openlattice/hazelcast/processors/RemoveDataExpirationPolicyProcessor.kt
pckeyes: i don't remember what that was for... i'll delete it

src/main/kotlin/com/openlattice/data/DataGraphService.kt
makosblade: Cool, no worries! Just wanted to make sure that it had moved this way intentionally 😄

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
makosblade: `val ignoredCleardEntitiesClause: String`
Should work? Or does it not end up being empty-string?

pckeyes: it does not, it ends up being an uninitialized variable

makosblade: Oooooofff. That . . . is . . . . what. But nullability whyyyyy 🤦‍♂

src/main/java/com/openlattice/hazelcast/serializers/EntitySetStreamSerializer.java
makosblade: Oof nvm, the semantics aren't great if invoked from Java 😢

src/main/java/com/openlattice/hazelcast/serializers/EntitySetStreamSerializer.java
makosblade: Oof nvm, the semantics aren't great if invoked from Java 😢

src/main/java/com/openlattice/hazelcast/serializers/EntitySetStreamSerializer.java
pckeyes: bummer :(

src/main/kotlin/com/openlattice/data/DataGraphService.kt
makosblade: Yeah the code is definitely more elegant looking the way its currently written
But you could move the `return`s into each branch, have a `val dateMinusTTEAsOffsetDateTime = . . . ` at the top that gets used in two of the cases and then gets turned into a `.toLocalDate()` in one and the `.toEpochMilli()` in the other?
Little bit ugly. I leave it up to you, it's a pretty minuscule optimization ¯\_(ツ)_/¯

PR #624, opened 2019-09-12
LATTICE-2196-deleteEntities does not delete row containing only the entity key id in data table

pckeyes:
Currently, when deleting an entity, deletion from the postgres data table using property types of the entity set as a query parameter. This leaves behind a row in the data table that contains the entity key id with no property values.

This PR changes the delete statement to not use property types as a condition for selecting rows. I don't think this will break anything, as there logically shouldn't be a case where not all of the properties of an entity are deleted when that entity is deleted.


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -487,7 +487,7 @@ internal val deleteEntitySetEntityKeys = "DELETE FROM ${IDS.name} WHERE ${ENTITY
  * 5. property type ids
  */
 internal val deletePropertiesOfEntitiesInEntitySet = "DELETE FROM ${DATA.name} " +
-        "WHERE ${ENTITY_SET_ID.name} = ? AND ${ID_VALUE.name} = ANY(?) AND ${PARTITION.name} = ? AND ${PARTITIONS_VERSION.name} = ? AND ${PROPERTY_TYPE_ID.name} = ANY(?) "
+        "WHERE ${ENTITY_SET_ID.name} = ? AND ${ID_VALUE.name} = ANY(?) AND ${PARTITION.name} = ? AND ${PARTITIONS_VERSION.name} = ? "
```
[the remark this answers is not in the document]
pckeyes: This should be fixed now.

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -576,13 +576,37 @@ class PostgresEntityDataQueryService(
...
                 .map { (partition, entities) ->
-                    deleteEntities(entitySetId, entities, authorizedPropertyTypes, partition, partitionVersion)
+                    deletePropertiesFromEntities(entitySetId, entities, authorizedPropertyTypes, partition, partitionVersion)
                 }.sum()
 
         return WriteEvent(System.currentTimeMillis(), numUpdates)
     }
 
-    private fun deleteEntities(
+    fun deleteEntityDataAndEntity(
```
katherinebernstein: Minor, but for clarity can you change this to `entities`? (since it deletes multiple)

geekbeast: +1

pckeyes: fixed :)

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -616,6 +640,36 @@ class PostgresEntityDataQueryService(
...
     }
 
+    private fun deleteEntities(
+            entitySetId: UUID,
+            entities: Collection<UUID>,
+            partition: Int,
+            partitionVersion: Int
+    ): Int {
+        return hds.connection.use { connection ->
+            connection.autoCommit = false
```
you-leee: Shoudn't be autocommit set to true after committing?

pckeyes: since the connection doesn't get closed I'm pretty sure you're right, but either way it seems safer to just reset it to true.

makosblade: We've found out that apparently when `.close()` is called on `connection` (by hikari in the case of `.use {`, the connection will actually be reset to its default state automatically (true)

you-leee: Good to know! Thanks!

PR #627, opened 2019-09-25
Bugfix/naming consistency

jokedurnez:
Found this a bit a weird function name with feedbacks when returning an iterable and feedback when it's singular. It's grammatically incorrect.


src/main/java/com/openlattice/hazelcast/HazelcastMap.java
```diff
@@ -67,7 +67,7 @@
     LINKING_EDGES,
     LINKING_ENTITIES,
     LINKING_ENTITY_VERTICES,
-    LINKING_FEEDBACKS,
+    LINKING_FEEDBACK,
```
katherinebernstein: Is there a mpastore for this? If there is the name needs to be updated there too

you-leee: +1

you-leee: `LinkingFeedbackMapstore`

jokedurnez: ok done... I think...  Is the commit below enough?

katherinebernstein: If you're using intellij to do the refactor it should update all usages for you so you don't have to search through all the projects or worry about missing any (and you can also find all usages using intellij to double check)

PR #629, opened 2019-09-27
Feature/lattice 1959 new member materialized view grants

you-leee:


src/main/kotlin/com/openlattice/assembler/Assembler.kt
```diff
@@ -274,14 +276,51 @@ class Assembler(
 
     @Subscribe
     fun handleAddMembersToOrganization(event: MembersAddedToOrganizationEvent) {
+        // check if organization is initialized
+        ensureAssemblyInitialized(event.organizationId)
```
makosblade: I added an EntryProcessor you can use in this method to avoid serializing the whole `OrganizationAssembly` just for a boolean check: https://github.com/openlattice/conductor-client/pull/619/files#diff-d008436a30debc36dd15f70fe40cfb74

src/main/kotlin/com/openlattice/assembler/Assembler.kt
```diff
@@ -274,14 +276,51 @@ class Assembler(
...
+            )
+            val authorizedPropertyTypeAcls = authorizationManager.getAuthorizedObjectsOfType(
+                    principal.principal,
+                    SecurableObjectType.PropertyTypeInEntitySet,
+                    EdmAuthorizationHelper.READ_PERMISSION
+            )
+
+            val allEntitySets = entitySets
+                    .getAll(authorizedEntitySets.map { it[0] }.collect(Collectors.toSet()))
+                    .filter { materializedEntitySets.containsKey(EntitySetAssemblyKey(it.key, event.organizationId)) }
```
makosblade: This is probably fine but if there's a way to create a `ContainsKeyPredicate` or a similar idea, it would definitely be preferable over filtering the results like this

src/main/kotlin/com/openlattice/assembler/Assembler.kt
you-leee: Nice, thanks! Will do after merge!

src/main/kotlin/com/openlattice/assembler/Assembler.kt
you-leee: Doh, nice catch!

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -470,21 +502,34 @@ class AssemblerConnectionManager(
...
     ): String {
-        val postgresUserName = if (principal.type == PrincipalType.USER) {
-            buildPostgresUsername(securePrincipalsManager.getPrincipal(principal.id))
-        } else {
-            buildPostgresRoleName(securePrincipalsManager.lookupRole(principal))
-        }
+        val postgresUserName =
+                if (principal.type == PrincipalType.USER) {
+                    buildPostgresUsername(securePrincipalsManager.getPrincipal(principal.id))
+                } else {
```
katherinebernstein: just double checking -- `grantSelectSql` is only ever called on users and roles, right? To be safe / clear, you could make this an `} else if (principal.type === PrincipalType.ROLE) {` and add an `else { throw SomeException }`

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -470,21 +502,34 @@ class AssemblerConnectionManager(
...
 
+    /**
+     * Build grant select sql statement for a given table and user with column level security.
+     * If properties (columns) are left empty, it will grant select on whole table.
+     */
+    private fun grantSelectSql(
+            tableName: String,
+            postgresUserName: String,
+            columns: List<String>
+    ): String {
```
katherinebernstein: (same comment)

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
you-leee: At this point, there shouldn't be any other type of principals, because we filter them in `Assembler.getReadAuthorizedUsersAndRolesOnEntitySet`, but yeah, you're right, I also it's a good idea to explicitly show that in the code too

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
you-leee: what do you mean by same comment here?

geekbeast: confused as well

katherinebernstein: same comment as the one above (re grantSelectSql only ever being called on users/roles)

you-leee: at this point, the parameter is already the postgres username/rolename, only a string

katherinebernstein: lol somehow I read that as a check on principal type -- ignore me

PR #631, opened 2019-10-03
Bugfix/lattice 2086 cleanup linking indexing

you-leee:
- implement ORIGIN_ID and LAST_WRITE metadata options for `PEDQS.getEntitySetIterable`
- add origin id to linked inidices in elasticsearch (meaning there will be 1 entry in entity array for each source entity)
- add entity set id and last_write to all indices in elasticsearch
- mark as indexed, when indexing directly (https://jira.openlattice.com/browse/LATTICE-2224)
- update last_write on partial clear/deletes (https://jira.openlattice.com/browse/LATTICE-2226)
- cleanups around PEDQS


src/main/java/com/openlattice/conductor/rpc/ConductorElasticsearchApi.java
```diff
@@ -143,7 +143,8 @@ boolean createEntityData(
 
     boolean createBulkLinkedData(
             UUID entityTypeId,
-            Map<UUID, Map<UUID, Map<UUID, Set<Object>>>> entitiesByLinkingId );
+            Map<UUID, Map<UUID, Map<UUID, Map<UUID, Set<Object>>>>> entitiesByLinkingId
```
katherinebernstein: can you document somewhere what all these UUIDs represent?

src/main/java/com/openlattice/postgres/PostgresTable.java
```diff
@@ -113,6 +113,7 @@
                             VERSIONS )
                     .primaryKey( ID, EDGE_COMP_1, EDGE_COMP_2, COMPONENT_TYPES )
                     .distributionColumn( ID_VALUE );
+    @Deprecated
```
katherinebernstein: 🙌

src/main/java/com/openlattice/conductor/rpc/ConductorElasticsearchApi.java
you-leee: sure!

src/main/java/com/openlattice/search/SearchService.java
```diff
@@ -226,6 +228,7 @@ public void deleteEntities( EntitiesDeletedEvent event ) {
     @Timed
     @Subscribe
     public void deleteLinkedEntities( LinkedEntitiesDeletedEvent event ) {
+        // TODO figure how we handle deletion of entities which belong to a linking entity
```
katherinebernstein: do we have a ticket for this?

you-leee: Yes! https://jira.openlattice.com/browse/LATTICE-2225
This ticket will handle both normal and linked entity deletions in elasticsearch

src/main/kotlin/com/openlattice/data/events/EntitiesUpsertedEvent.kt
```diff
@@ -18,33 +16,10 @@
...
-    public UUID getEntitySetId() {
-        return entitySetId;
-    }
+import java.util.UUID
 
-    public Map<UUID, Map<UUID, Set<Object>>> getEntities() {
-        return entities;
-    }
-}
+data class EntitiesUpsertedEvent(val entitySetId: UUID, val entities: Map<UUID, Map<UUID, Set<Any>>>)
```
katherinebernstein: 😍

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -56,64 +47,95 @@ val jsonValueColumnsSql = PostgresDataTables.dataColumns.entries
...
-        partitionsPresent: Boolean,
-        selectOriginIds: Boolean = false
+        partitionsPresent: Boolean
 ): Pair<String, Set<SqlBinder>> {
     val filtersClauses = buildPreparableFiltersClause(startIndex, propertyTypes, propertyTypeFilters)
     val filtersClause = if (filtersClauses.first.isNotEmpty()) " AND ${filtersClauses.first} " else ""
 
+    val metadataOptionColumns = metadataOptions.associateWith(::mapMetaDataToColumnSql)
+    val nonAggregatedMetadataSql = metadataOptionColumns
+            .filter { !isMetaDataAggregated(it.key) }.values.joinToString("")
```
katherinebernstein: do these not need to be comma-separated?

src/main/java/com/openlattice/conductor/rpc/ConductorElasticsearchApi.java
geekbeast: JavaDoc or KDoc please

src/main/kotlin/com/openlattice/hazelcast/serializers/SearchResultStreamSerializer.kt
```diff
@@ -0,0 +1,54 @@
...
+        return StreamSerializerTypeIds.SEARCH_RESULT.ordinal
+    }
+
+    override fun getClazz(): Class<out SearchResult> {
+        return SearchResult::class.java
+    }
+
+    override fun write(output: ObjectDataOutput, obj: SearchResult) {
+        output.writeLong(obj.numHits)
+        output.writeByteArray(ObjectMappers.getSmileMapper().writeValueAsBytes(obj.hits))
```
geekbeast: It is my understanding that you don't need to use typereference in Kotlin as the type system will figure it out automatically for you.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
you-leee: Nope, the `mapMetaDataToColumnSql` has already the comma prefix.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -330,11 +263,11 @@ internal fun selectLinkingEntitySetSql(linkingEntitySetId: UUID): String {
...
-            "${DataTables.LAST_WRITE.name} = now(), " +
-            "${VERSION.name} = CASE " +
-                "WHEN abs(${IDS.name}.${VERSION.name}) < abs(?) THEN ? " +
-                "ELSE ${IDS.name}.${VERSION.name} " +
-            "END " +
+        "${DataTables.LAST_WRITE.name} = now(), " +
+        "${VERSION.name} = CASE " +
+        "WHEN abs(${IDS.name}.${VERSION.name}) < abs(?) THEN ? " +
+        "ELSE ${IDS.name}.${VERSION.name} " +
+        "END " +
```
katherinebernstein: not of the utmost importance but I think the old formatting here was a bit easier to read

you-leee: Oh, yeah, that was autoformat. Will change it back

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -649,6 +700,54 @@ class PostgresEntityDataQueryService(
...
+    private fun updateLastWrite(
+            connection: Connection,
+            entitySetId: UUID,
+            entities: Collection<UUID>,
+            partitionsInfo: PartitionsInfo = partitionManager.getEntitySetPartitionsInfo(entitySetId)
+    ): Int {
+        val idsArr = PostgresArrays.createUuidArray(connection, entities)
+        val partitionsArr = PostgresArrays.createIntArray(
+                connection,
+                entities.map { getPartition(it, partitionsInfo.partitions.toList()) }
```
katherinebernstein: It's probably better to pull the `partitionsInfo.partitions.toList()` out of the `.map` in case there's a very large number of entities since toList() requires iterating over all partitions

you-leee: 👍

src/main/kotlin/com/openlattice/postgres/PostgresResultSetAdapters.kt
```diff
@@ -26,71 +27,94 @@ internal class PostgresResultSetAdapters
...
+        metadataOptions: Set<MetadataOption>,
         byteBlobDataManager: ByteBlobDataManager
 ): Pair<UUID, MutableMap<FullQualifiedName, MutableSet<Any>>> {
     val id = id(rs)
     val entitySetId = entitySetId(rs)
     val propertyTypes = authorizedPropertyTypes.getValue(entitySetId)
+
+    val entity = readJsonDataColumns(rs, propertyTypes, byteBlobDataManager)
+
+    val entityByFqn = entity.mapKeys { propertyTypes.getValue(it.key).type }.toMutableMap()
```
katherinebernstein: Are we guaranteed that only authorized property types will be present in the results here?

katherinebernstein: ohh I see -- readJsonDataColumns handles that

you-leee: Yep :)

src/main/java/com/openlattice/search/SearchService.java
```diff
@@ -292,25 +295,23 @@ public void deleteOrganization( OrganizationDeletedEvent event ) {
...
-        elasticsearchApi.createBulkEntityData( entityTypeId, event.getEntitySetId(), event.getEntities() );
-    }
-
-    private void indexLinkedEntities(
-            UUID linkingEntitySetId, Map<UUID, Set<UUID>> linkingIds, Map<UUID, PropertyType> propertyTypes ) {
-        if ( !linkingIds.isEmpty() ) {
-            UUID entityTypeId = dataModelService.getEntityTypeByEntitySetId( linkingEntitySetId ).getId();
+        if ( elasticsearchApi.createBulkEntityData( entityTypeId, event.getEntitySetId(), event.getEntities() ) ) {
+            // mark them as indexed
+            indexingMetadataManager.markAsIndexed(
```
geekbeast: Wrap this in a timer. We need visibility to make sure thread pool doesn't get backlogged by users making small writes.

It should be distinct from the timer as markAsIndexed as it will skew results from background indexing service unless we break out aggregation of all markAsIndexed. @UnsungHero97

src/main/kotlin/com/openlattice/data/events/EntitiesDeletedEvent.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018. OpenLattice, Inc.
+ * Copyright (C) 2019. OpenLattice, Inc.
```
katherinebernstein: Copyright is mysterious to me -- can we change these dates?

katherinebernstein: jk this is a different file

src/main/kotlin/com/openlattice/hazelcast/serializers/BulkLinkedDataLambdasStreamSerializer.kt
```diff
@@ -65,25 +60,31 @@ class BulkLinkedDataLambdasStreamSerializer(
...
                 val entitySetId = readUUID(input)
 
-                val numBytes = input.readInt()
-                var entityData: HashMultimap<UUID, Any>
-                try {
-                    entityData = mapper.readValue(input.readBytes(numBytes), ref)
-                    entitiesByEntitySetId[entitySetId] = entityData
-                } catch (e: IOException) {
-                    logger.debug("Unable to deserialize entities for linking id: {}", linkingId)
+                val originIdsSize = input.readInt()
```
makosblade: I don't see any equivalent updates to the `write()` method for this

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -534,13 +578,19 @@ class PostgresEntityDataQueryService(
...
+     * @param partitionsInfo Contains the partition information for the requested entity set.
      */
     fun clearEntities(
-            entitySetId: UUID, entityKeyIds: Set<UUID>, authorizedPropertyTypes: Map<UUID, PropertyType>
+            entitySetId: UUID,
+            entityKeyIds: Set<UUID>,
+            authorizedPropertyTypes: Map<UUID, PropertyType>,
+            partitionsInfo: PartitionsInfo = partitionManager.getEntitySetPartitionsInfo(entitySetId)
     ): WriteEvent {
+        val tombstoneVersion = -System.currentTimeMillis()
```
makosblade: I'm confused. I was liking the pattern of `val version = System.currentTimMillis()` and then using it as `-version` but now this is the other way around. What's the rationale here?

src/main/kotlin/com/openlattice/hazelcast/serializers/BulkLinkedDataLambdasStreamSerializer.kt
you-leee: Ah, nice catch!

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
you-leee: Yeah, probably it's more readable that way. I was using the same name `tombstoneVersion` here, just put it 1 call up, because both of `tombstone` functions are using the -version. Will change to -version then.

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
```diff
@@ -8,59 +8,47 @@ import com.openlattice.postgres.PostgresArrays
...
+                                        .groupBy { it.value }
+                                        .mapValues { it.value.map { it.key } }
+                                        .forEach { (lastWrite, entities) ->
+
+                                            val idsArray = PostgresArrays.createUuidArray(connection, entities)
+                                            stmt.setObject(1, lastWrite)
+                                            stmt.setObject(2, entitySetId)
+                                            stmt.setArray(3, idsArray)
+                                            stmt.setInt(4, partition)
+                                            stmt.addBatch()
```
katherinebernstein: I think we also want to be setting partitions_version as a constraint here

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -482,35 +526,35 @@ class PostgresEntityDataQueryService(
...
-            ps.setLong(3, tombstoneVersion)
+            ps.setLong(1, -version)
+            ps.setLong(2, -version)
+            ps.setLong(3, -version)
             ps.setObject(4, entitySetId)
             ps.executeUpdate()
         }
 
-        return WriteEvent(tombstoneVersion, numUpdated)
+        return WriteEvent(-version, numUpdated)
```
katherinebernstein: I think the WriteEvent should return the nonnegative version -- these events are used for auditing, and the version for auditing should effectively just be the timestamp (so that when you query for events in a given time range, the tombstone events will also show up). I realize we're using negative versions for the WriteEvents in all the tombstone calls, but I think that's a bug

katherinebernstein: Alternately we could force AuditingComponent to take the absolute value of the version when converting it to an OffsetDateTime

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
you-leee: 👍

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
you-leee: Okay, that wasn't really clear to me, but makes sense. Will do both!

PR #632, opened 2019-10-04
update EPs that are readonly abstractEPs to use the direct type

makosblade:
This PR:
- Uses new ReadOnly type so that the implementer doesn't have to pass a boolean to an implemented constructor

### DIFF
--- a/src/main/java/com/openlattice/datastore/services/EdmService.java
+++ b/src/main/java/com/openlattice/datastore/services/EdmService.java
@@ -684,19 +684,16 @@ public class EdmService implements EdmManager {
     @Override
     public Map<UUID, PropertyType> getPropertyTypesForEntitySet( UUID entitySetId ) {
 
-        Optional<UUID> maybeEtId = (Optional<UUID>) entitySets.executeOnKey( entitySetId, new GetEntityTypeFromEntitySetEntryProcessor() );
-        if ( maybeEtId.isEmpty() ) {
+        UUID maybeEtId = (UUID) entitySets.executeOnKey( entitySetId, new GetEntityTypeFromEntitySetEntryProcessor());
+        if ( maybeEtId == null ) {
             throw new ResourceNotFoundException( "Entity set " + entitySetId.toString() + " does not exist." );
         }
-
-        UUID entityTypeId = maybeEtId.get();
-        
-        Optional<Set<UUID>> maybeEtProps = (Optional<Set<UUID>>) entityTypes.executeOnKey( entityTypeId, new GetPropertiesFromEntityTypeEntryProcessor() );
-        if ( maybeEtProps.isEmpty() ) {
-            throw new ResourceNotFoundException( "Entity type " + entityTypeId.toString() + " does not exist." );
+        Set<UUID> maybeEtProps = (Set<UUID>) entityTypes.executeOnKey( maybeEtId, new GetPropertiesFromEntityTypeEntryProcessor());
+        if ( maybeEtProps == null ) {
+            throw new ResourceNotFoundException( "Entity type " + maybeEtId.toString() + " does not exist." );
         }
 
-        return propertyTypes.getAll( maybeEtProps.get() );
+        return propertyTypes.getAll( maybeEtProps );
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


makosblade:
Depends on https://github.com/geekbeast/rhizome/pull/283

src/main/kotlin/com/openlattice/edm/processors/GetEntityTypeFromEntitySetEntryProcessor.kt
```diff
@@ -3,9 +3,10 @@ package com.openlattice.edm.processors
 import com.hazelcast.core.ReadOnly
 import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor
 import com.openlattice.edm.EntitySet
+import com.openlattice.rhizome.hazelcast.entryprocessors.AbstractReadOnlyRhizomeEntryProcessor
 import java.util.*
 
-class GetEntityTypeFromEntitySetEntryProcessor : AbstractRhizomeEntryProcessor<UUID, EntitySet, Optional<UUID>>(), ReadOnly {
+class GetEntityTypeFromEntitySetEntryProcessor : AbstractReadOnlyRhizomeEntryProcessor<UUID, EntitySet, Optional<UUID>>() {
     override fun process(entry: MutableMap.MutableEntry<UUID, EntitySet?>): Optional<UUID> {
```
you-leee: Got this error when calling `EdmService.getPropertyTypesForEntitySet` :(
```2019-10-04T15:14:19,333 ERROR [qtp1475028309-91] com.openlattice.controllers.util.BaseExceptionHandler - 
com.hazelcast.nio.serialization.HazelcastSerializationException: Failed to serialize 'java.util.Optional'```

Comment (makosblade) [reply src/main/kotlin/com/openlattice/edm/processors/GetEntityTypeFromEntitySetEntryProcessor.kt]: Uhhhhh. What. Do we not have stream serializers for java.util.Optional?

Comment (you-leee) [reply src/main/kotlin/com/openlattice/edm/processors/GetEntityTypeFromEntitySetEntryProcessor.kt]: Well, we have a custom serializer for Optionals `OptionalStreamSerializers`, but " "Shouldn't Optional be Serializable?"  The short answer is that the Java Lambda (JSR-335) expert group considered and rejected it."

File: src/main/java/com/openlattice/datastore/services/EdmService.java [hunk 2]
```diff
@@ -684,19 +684,16 @@ public PropertyType getPropertyType( FullQualifiedName propertyType ) {
     @Override
     public Map<UUID, PropertyType> getPropertyTypesForEntitySet( UUID entitySetId ) {
 
-        Optional<UUID> maybeEtId = (Optional<UUID>) entitySets.executeOnKey( entitySetId, new GetEntityTypeFromEntitySetEntryProcessor() );
-        if ( maybeEtId.isEmpty() ) {
+        UUID maybeEtId = (UUID) entitySets.executeOnKey( entitySetId, new GetEntityTypeFromEntitySetEntryProcessor());
```
you-leee: I think the cast should be after the null check

src/main/java/com/openlattice/datastore/services/EdmService.java
```diff
@@ -684,19 +684,16 @@ public PropertyType getPropertyType( FullQualifiedName propertyType ) {
...
+        if ( maybeEtId == null ) {
             throw new ResourceNotFoundException( "Entity set " + entitySetId.toString() + " does not exist." );
         }
-
-        UUID entityTypeId = maybeEtId.get();
-        
-        Optional<Set<UUID>> maybeEtProps = (Optional<Set<UUID>>) entityTypes.executeOnKey( entityTypeId, new GetPropertiesFromEntityTypeEntryProcessor() );
-        if ( maybeEtProps.isEmpty() ) {
-            throw new ResourceNotFoundException( "Entity type " + entityTypeId.toString() + " does not exist." );
+        Set<UUID> maybeEtProps = (Set<UUID>) entityTypes.executeOnKey( maybeEtId, new GetPropertiesFromEntityTypeEntryProcessor());
```
you-leee: Same comment here

src/main/java/com/openlattice/datastore/services/EdmService.java
makosblade: Doesn't it not matter since Set<UUID> and UUID are both able to safely be null?

src/main/kotlin/com/openlattice/edm/processors/GetEntityTypeFromEntitySetEntryProcessor.kt
makosblade: wow. That is my least favorite news all year

src/main/java/com/openlattice/datastore/services/EdmService.java
you-leee: Ah, yeah, you're right! I thought it will nullpointer

PR #646, opened 2019-10-08
Feature/lattice 2228 atlas external access

pckeyes:
Creation and handling of securable object types that represent tables and columns of an organization's atlas database. These new securable objects will allow for creation of acl keys to enable granting permissions at the table or column level within atlas.


geekbeast:
Don't call it atlas.

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -920,6 +922,27 @@ public static CollectionTemplateKey collectionTemplateKey( ResultSet rs ) throws
         return new CollectionTemplateKey( entitySetCollectionId, templateTypeId );
     }
 
+    public static OrganizationExternalDatabaseTable organizationAtlasTable( ResultSet rs ) throws SQLException {
+        UUID id = id(rs);
+        String name = name(rs);
+        String title = title(rs);
+        Optional<String> description = Optional.ofNullable( description( (rs) ) );
+        UUID organizationId = id(rs);
```
katherinebernstein: I think this should be `organizationId(rs)`? (assuming that exists?)

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
```diff
@@ -920,6 +922,27 @@ public static CollectionTemplateKey collectionTemplateKey( ResultSet rs ) throws
...
+        return new OrganizationExternalDatabaseTable( id, name, title, description, organizationId );
+    }
+
+    public static OrganizationExternalDatabaseColumn organizationAtlasColumn( ResultSet rs ) throws SQLException {
+        UUID id = id(rs);
+        String name = name(rs);
+        String title = title(rs);
+        Optional<String> description = Optional.ofNullable( description( (rs) ) );
+        UUID tableId = id(rs);
+        UUID organizationId = id(rs);
```
katherinebernstein: same comment -- and `tableId` probably needs its own too

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,361 @@
...
+        val columnAclKey = AclKey(orgId, column.tableId, column.id)
+        authorizationManager.setSecurableObjectType(columnAclKey, SecurableObjectType.OrganizationAtlasColumn)
+        authorizationManager.addPermission(columnAclKey, principal, EnumSet.allOf(Permission::class.java))
+
+        return column.id
+    }
+
+    fun createNewOrganizationExternalDatabaseTable(orgId: UUID, tableName: String, columnNameToSqlType: LinkedHashMap<String, String>) {
+        //TODO figure out title stuff
+        val table = OrganizationExternalDatabaseTable(Optional.empty(), tableName, "title", Optional.empty(), orgId)
```
katherinebernstein: What does "title" do?

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
pckeyes: Will check

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: nothing useful currently. It's included in the OrganizationExternalDatabase object simply because it's a child class of AbstractSecurableObject. I'm not sure if it is necessary / if the title field can just serve as the name field of the column or table?

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
pckeyes: It exists! Will fix

src/main/java/com/openlattice/postgres/ResultSetAdapters.java
pckeyes: Will do

src/main/java/com/openlattice/postgres/PostgresTable.java
```diff
@@ -309,6 +309,28 @@
...
+                            ORDINAL_POSITION);
+
+    public static final PostgresTableDefinition ORGANIZATION_EXTERNAL_DATABASE_TABLE =
+            new PostgresTableDefinition( "organization_external_database_tables" )
+                    .addColumns(
+                            ID,
+                            NAME,
+                            TITLE,
+                            DESCRIPTION,
+                            ORGANIZATION_ID );
```
makosblade: I feel like these need primary keys 🤔

you-leee: `ID` is already a primary key

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+    fun deleteOrganizationExternalDatabaseTables(orgId: UUID, tableIds: Set<UUID>) {
+        tableIds.forEach { deleteOrganizationExternalDatabaseTable(orgId, it) }
+    }
+
+    fun deleteOrganizationExternalDatabaseTable(orgId: UUID, tableId: UUID) {
+        organizationExternalDatabaseTables.remove(tableId) //TODO make this a set so we can batch delete, entry processor?
+        aclKeyReservations.release(tableId)
+
+        //delete columns that belonged to this table
+        val belongsToDeletedTable = Predicates.equal("tableId", tableId)
```
you-leee: Also, there should be an index for this probably too + would use the constant for it (you can move it to the companion class as `@JvmStatic val ...` to be able to use it as `OrganizationExternalDatabaseTableMapstore.TABLE_ID_INDEX`)

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    /**
+     * Deletes an organization's entire database.
+     * Is called when an organization is deleted.
+     */
+    //TODO break into further helper methods to reduce reuse of code
+    fun deleteOrganizationExternalDatabase(orgId: UUID) {
+        //remove all tables/columns within org
+        val belongsToDeletedDB = Predicates.equal("organizationId", orgId)
```
you-leee: You should use the constant `ORGANIZATION_ID_INDEX` here

src/main/java/com/openlattice/postgres/PostgresTable.java
```diff
@@ -309,6 +309,28 @@
                     .addColumns( ORGANIZATION_ID, INITIALIZED )
                     .primaryKey( ORGANIZATION_ID );
 
+    public static final PostgresTableDefinition ORGANIZATION_EXTERNAL_DATABASE_COLUMN =
```
you-leee: This should probably be ..._TABLE and not ..._COLUMN

src/main/kotlin/com/openlattice/organizations/processors/UpdateOrganizationExternalDatabaseTable.kt
```diff
@@ -0,0 +1,48 @@
+package com.openlattice.organizations.processors
+
+import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor
+import com.openlattice.edm.requests.MetadataUpdate
+import com.openlattice.organization.OrganizationExternalDatabaseTable
+import java.util.*
+
+class UpdateOrganizationExternalDatabaseTable(val update: MetadataUpdate) :
```
you-leee: You can make it into a data class, so you don't need to implement `equals `, `hashCode` and `toString`

src/main/kotlin/com/openlattice/organizations/processors/UpdateOrganizationExternalDatabaseColumn.kt
```diff
@@ -0,0 +1,49 @@
+package com.openlattice.organizations.processors
+
+import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor
+import com.openlattice.edm.requests.MetadataUpdate
+import com.openlattice.organization.OrganizationExternalDatabaseColumn
+import java.util.*
+
+class UpdateOrganizationExternalDatabaseColumn(val update: MetadataUpdate) :
```
you-leee: You can make it into a data class, so you don't need to implement `equals `, `hashCode` and `toString`

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+                                "LEFT OUTER JOIN information_schema.constraint_column_usage on " +
+                                "information_schema.columns.column_name = information_schema.constraint_column_usage.column_name " +
+                                "LEFT OUTER JOIN information_schema.table_constraints " +
+                                "on information_schema.constraint_column_usage.constraint_name = " +
+                                "information_schema.table_constraints.constraint_name " +
+                                "WHERE information_schema.columns.table_name = '$tableName' " +
+                                "$columnCondition"
+                )
+                while (rs.next()) {
+                    val columnName = rs.getString("column_name")
```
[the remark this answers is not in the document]
you-leee: ➕

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+                    val dataType = rs.getString("data_type")
+                    val position = rs.getInt("ordinal_position")
+                    val isPrimaryKey = rs.getString("constraint_type") == "PRIMARY KEY"
```
you-leee: This should be in a dedicated field (`"PRIMARY KEY"`) or in an enum `ConstraintType`

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+            val organizationAtlasTable = organizationExternalDatabaseTables[securableObjectId]!!
+            tableName = organizationAtlasTable.name
+            columnName = ""
+        }
+        //add checks in here for map indexing??
+        return Pair(tableName, columnName)
+    }
+
+    private fun getDBUser(principalId: String): String {
+        val securePrincipal = securePrincipalsManager.getPrincipal(principalId)
```
you-leee: Either here or 1 call back ( in `addTrustedUser`) there should be a check, if the principals type is a user and if not, fail the call.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+        val dbName = PostgresDatabases.buildOrganizationDatabaseName(orgId)
+        assemblerConnectionManager.connect(dbName).use {
+            it.connection.createStatement().use { stmt ->
+                stmt.execute("DROP DATABASE $dbName")
+            }
+        }
+    }
+
+    /*PERMISSIONS*/
+    fun addTrustedUser(orgId: UUID, userPrincipal: Principal, ipAdresses: Set<String>) {
```
you-leee: Shouldn't there be a call for adding trusted role? Or it's not in the requirements for this feature? (We are creating automaticcaly a postgres role for every openlattice role btw)

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+
+        return table.id
+    }
+
+    fun createOrganizationExternalDatabaseColumn(orgId: UUID, column: OrganizationExternalDatabaseColumn): UUID {
+        val principal = Principals.getCurrentUser()
+        Principals.ensureUser(principal)
+
+        checkState(organizationExternalDatabaseTables[column.tableId] != null,
+                "OrganizationExternalDatabaseColumn ${column.name} belongs to a table that does not exist")
```
you-leee: Minor thing: I would also log the table name here

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+        val columnFQN = FullQualifiedName(column.tableId.toString(), column.name)
+        checkState(organizationExternalDatabaseColumns.putIfAbsent(column.id, column) == null,
+                "OrganizationExternalDatabaseColumn ${columnFQN.fullQualifiedNameAsString} already exists")
```
you-leee: Really minor: No need to call `fullQualifiedNameAsString`. If you look at the `FullQualifiedName` implementation, `toString()` returns the same

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+
+    /*GET*/
+    fun getOrganizationIds(): Set<UUID> {
+        val orgIds = mutableSetOf<UUID>()
+        hds.connection.createStatement().use {
+            val rs = it.executeQuery("SELECT id FROM organizations")
+            while (rs.next()) {
+                orgIds.add(organizationId(rs))
+            }
+        }
```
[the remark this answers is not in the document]
you-leee: BTW: we already have MapStores, that store organization ids as their keys: `OrganizationAssemblyMapstore`, `OrganizationAppsMapstore`, `OrganizationDescriptionsMapstore`, `OrganizationEmailDomainsMapstore`..

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+        return organizationExternalDatabaseColumns[columnId]!!
+    }
+
+    /*DELETE*/
+    fun deleteOrganizationExternalDatabaseTables(orgId: UUID, tableIds: Set<UUID>) {
+        tableIds.forEach { deleteOrganizationExternalDatabaseTable(orgId, it) }
+    }
+
+    fun deleteOrganizationExternalDatabaseTable(orgId: UUID, tableId: UUID) {
+        organizationExternalDatabaseTables.remove(tableId) //TODO make this a set so we can batch delete, entry processor?
```
you-leee: Like the TODO here :)

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+        deleteOrganizationExternalDatabaseColumns(orgId, columnIds)
+
+    }
+
+    fun deleteOrganizationExternalDatabaseColumns(orgId: UUID, columnIds: Set<UUID>) {
+        columnIds.forEach { deleteOrganizationExternalDatabaseColumn(orgId, it) }
+    }
+
+    fun deleteOrganizationExternalDatabaseColumn(orgId: UUID, columnId: UUID) {
+        organizationExternalDatabaseColumns.remove(columnId) //TODO make this a set so we can batch delete, entry processor?
```
you-leee: 👍 on TODO

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+     * Deletes an organization's entire database.
+     * Is called when an organization is deleted.
+     */
+    //TODO break into further helper methods to reduce reuse of code
+    fun deleteOrganizationExternalDatabase(orgId: UUID) {
+        //remove all tables/columns within org
+        val belongsToDeletedDB = Predicates.equal("organizationId", orgId)
+        val tablesToDelete = organizationExternalDatabaseTables.values(belongsToDeletedDB)
+        tablesToDelete.forEach { table ->
+            organizationExternalDatabaseTables.remove(table.id)
```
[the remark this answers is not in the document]
you-leee: ➕

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+        return newColumns
+    }
+
+    /*PRIVATE FUNCTIONS*/
+    private fun createPrivilegesSql(action: Action, privileges: List<String>, tableName: String, columnName: String, dbUser: String): String {
+        val privilegesAsString = privileges.joinToString(separator = ", ")
+        val tableNamePath = "$PUBLIC_SCHEMA.$tableName"
+        return if (action == Action.REMOVE) {
+            "REVOKE $privilegesAsString $columnName ON $tableNamePath FROM $dbUser"
+        } else {
```
you-leee: I think there should be a check for `Action.ADD` here and not accept all other actions

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+                            } else {
+                                if (ace.permissions.contains(Permission.WRITE)) {
+                                    privileges.addAll(listOf("INSERT", "UPDATE"))
+                                }
+                                if (ace.permissions.contains(Permission.READ)) {
+                                    privileges.add("SELECT")
+                                }
+                            }
+                            val sql = createPrivilegesSql(action, privileges, tableAndColumnNames.first, tableAndColumnNames.second, dbUser)
+                            if (action == Action.SET) {
```
you-leee: I would use `Action.ADD` rather, since (in the call) you are not specifying what to set, but rather add all of the appropriate ones

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+            val dbName = PostgresDatabases.buildOrganizationDatabaseName(orgId)
+            assemblerConnectionManager.connect(dbName).use { dataSource ->
+                dataSource.connection.autoCommit = false
+                dataSource.connection.createStatement().use { stmt ->
+                    acls.forEach {
+                        val tableAndColumnNames = getTableAndColumnNames(AclKey(it.aclKey))
+                        it.aces.forEach { ace ->
+                            val dbUser = getDBUser(ace.principal.id)
+                            val privileges = mutableListOf<String>()
+                            if (ace.permissions.contains(Permission.OWNER)) {
```
you-leee: Use when instead here:
```        val result = with(ace.permissions) {
            when {
                contains(Permission.OWNER) -> listOf("ALL")
                contains(Permission.WRITE) -> listOf("INSERT", "UPDATE")
                contains(Permission.READ) -> listOf("SELECT")
                else -> throw error... or empty list
            }
        }```

File: src/main/kotlin/com/openlattice/hazelcast/serializers/OrganizationExternalDatabaseColumnStreamSerializer.kt [hunk 23]
```diff
@@ -0,0 +1,57 @@
...
+
+    companion object {
+        fun serialize(output: ObjectDataOutput, obj: OrganizationExternalDatabaseColumn) {
+            UUIDStreamSerializer.serialize(output, obj.id)
+            output.writeUTF(obj.name)
+            output.writeUTF(obj.title)
+            output.writeUTF(obj.description)
+            UUIDStreamSerializer.serialize(output, obj.tableId)
+            UUIDStreamSerializer.serialize(output, obj.organizationId)
+            output.writeUTF(obj.dataType)
```
katherinebernstein: what is the datatype of datatype? If it's a string, we should probably convert to an enum

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+                                privileges.add("ALL")
+                            } else {
+                                if (ace.permissions.contains(Permission.WRITE)) {
+                                    privileges.addAll(listOf("INSERT", "UPDATE"))
+                                }
+                                if (ace.permissions.contains(Permission.READ)) {
+                                    privileges.add("SELECT")
+                                }
+                            }
+                            val sql = createPrivilegesSql(action, privileges, tableAndColumnNames.first, tableAndColumnNames.second, dbUser)
```
you-leee: You should skip/throw error if `ace.permissions` do not contain any of the 3, because `privileges` will be empty. (or check if privileges are empty)

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+        authorizationManager.addPermission(columnAclKey, principal, EnumSet.allOf(Permission::class.java))
+
+        return column.id
+    }
+
+    /*GET*/
+    fun getOrganizationIds(): Set<UUID> {
+        val orgIds = mutableSetOf<UUID>()
+        hds.connection.createStatement().use {
+            val rs = it.executeQuery("SELECT id FROM organizations")
```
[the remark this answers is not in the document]
katherinebernstein: I think this should be O(#orgs)? You could use one of the org mapstores for this though, since they're all in memory already anyway.

Also unrelated comment (and irrelevant for mapstores) but it's preferable to use the PostgresColumn / PostgresTable constants for the column/table names instead of hardcoding them. Even if it's unlikely that a column or table name will change in the future, using the constant can be helpful even just for finding usages of that table, in case there's some change that requires finding/modifying all queries that touch a certain column or table.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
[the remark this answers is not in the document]
katherinebernstein: Oh ^^^ this about mapstores

But yeah, I think I prefer using the BasePGI, just so that we have consistency around any supports/metrics/internal logic we build for ourselves around querying postgres. I think the custom logic there is relatively minimal right now, but it does handle connection timeouts and retries and closing result sets on exceptions and all that stuff, and I can imagine that support getting built out further in the future.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+            while (rs.next()) {
+                orgIds.add(organizationId(rs))
+            }
+        }
+        return orgIds
+    }
+
+    fun getOrganizationExternalDatabaseTable(tableId: UUID): OrganizationExternalDatabaseTable {
+        return organizationExternalDatabaseTables[tableId]!!
```
katherinebernstein: Also super minor, but for consistency, I think we've been opting for the `a.getValue(b)` syntax over the `a[b]!!` syntax

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
katherinebernstein: +1000

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+                                if (ace.permissions.contains(Permission.WRITE)) {
+                                    privileges.addAll(listOf("INSERT", "UPDATE"))
+                                }
+                                if (ace.permissions.contains(Permission.READ)) {
+                                    privileges.add("SELECT")
+                                }
+                            }
+                            val sql = createPrivilegesSql(action, privileges, tableAndColumnNames.first, tableAndColumnNames.second, dbUser)
+                            if (action == Action.SET) {
+                                val revokeSql = createPrivilegesSql(Action.REMOVE, listOf("ALL"), tableAndColumnNames.first, tableAndColumnNames.second, dbUser)
```
katherinebernstein: Shouldn't the revoke be done before the add?

katherinebernstein: Okay nvm, I see the revoke is done first. So this is minor, but I'd move this if statement section to the top -- I feel like it's a little more intuitive to see
```
define revokeSql
addBatch(revokeSql)
define addSql
addBatch(addSql)
```

than

```
define addSql
define revokeSql
addBatch(revokeSql)
addBatch(addSql)
```

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+            aclKey = AclKey(orgId, tableId, columnId)
+        } else {
+            aclKey = AclKey(orgId, tableId)
+        }
+        assemblerConnectionManager.connect(dbName).use { dataSource ->
+            dataSource.connection.createStatement().use { stmt ->
+                val rs = stmt.executeQuery(
+                        "SELECT grantee, privilege_type " +
+                                "FROM information_schema.role_table_grants " +
+                                "WHERE table_name = '$tableName " +
```
katherinebernstein: is this missing a close quote?

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+        val columnCondition = if (columnName.isPresent) {
+            "AND information_schema.columns.column_name = '${columnName.get()}'"
+        } else {
+            ""
+        }
+        val newColumns = mutableSetOf<OrganizationExternalDatabaseColumn>()
+        assemblerConnectionManager.connect(dbName).use { dataSource ->
+            dataSource.connection.createStatement().use { stmt ->
+                val rs = stmt.executeQuery(
+                        "SELECT information_schema.tables.table_name, information_schema.columns.column_name, " +
```
katherinebernstein: This query looks like it has a lot of overlap with the query above -- is there a way to consolidate these?

src/main/kotlin/com/openlattice/organizations/processors/UpdateOrganizationExternalDatabaseTable.kt
```diff
@@ -0,0 +1,48 @@
...
+        AbstractRhizomeEntryProcessor<UUID, OrganizationExternalDatabaseTable, OrganizationExternalDatabaseTable?>() {
+
+    override fun process(entry: MutableMap.MutableEntry<UUID, OrganizationExternalDatabaseTable>): OrganizationExternalDatabaseTable? {
+        val table = entry.value
+
+        if (update.title.isPresent) {
```
katherinebernstein: Minor, but I think you can use ifPresent, which is just a little cleaner since you can avoid the .get() call 

i.e.
```
update.title.ifPresent( title -> table.title = title )
```
(or something like that, not 100% on the syntax)

src/main/kotlin/com/openlattice/hazelcast/serializers/OrganizationExternalDatabaseColumnStreamSerializer.kt
pckeyes: It is a string. It's whatever the datatype is in postgres. I can make it an enum class if that would be better.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: Oh good point!

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: Nice

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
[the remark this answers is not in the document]
pckeyes: 'Tis O(#orgs). And you're right I'll use the mapstore, wasn't thinking.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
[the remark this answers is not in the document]
pckeyes: Cool, I have a whole bunch of this logic sprinkled about, I'll convert it all to BPGI

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: Gotcha. Kotlin told me to change it but I'll change it back.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: :)

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: Ope, I do have that index in the mapstore and did not use it. Ty for the catch!

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: :)

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: again good catch :)

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: I honestly am not sure yet, I haven't gotten to this piece of the feature yet. I'll look into it though and get back to you!

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: gotcha ty!

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: for sure, I'll change that

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Revokes all privileges for a user on an organization's database
+     * when that user is removed from an organization.
+     */
+    fun revokeAllPrivilegesFromMember(orgId: UUID, userId: String) {
+        val dbName = PostgresDatabases.buildOrganizationDatabaseName(orgId)
+        val userName = getDBUser(userId)
+        assemblerConnectionManager.connect(dbName).use {
+            it.connection.createStatement().use { stmt ->
+                stmt.execute("REVOKE ALL ON DATABASE $dbName FROM $userName")
```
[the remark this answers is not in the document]
pckeyes: cool will do!

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: I'm not sure I understand. In the case of set, we need to remove existing privileges and only set the ones passed in, whereas with an add we just append the extra privileges, right?

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: YUP. Ty.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: Yea, all of these raw strings I'm planning on abstracting out, just while I'm still writing it helps me to see the queries directly as they still may need to be changed (I haven't tested them yet)

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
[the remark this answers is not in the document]
pckeyes: will do!

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: cool i'll change this

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: Earlier in the flow there is a filter that only lets through ADD, SET, and REMOVE I believe. I can still add the check though if it's still better to be safe.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: Oh for sure ty

src/main/kotlin/com/openlattice/organizations/processors/UpdateOrganizationExternalDatabaseTable.kt
pckeyes: nice will do

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
you-leee: Yeah, sorry! I thought it takes only 2 kinds of actions - remove and set - and would have changed set to add rather. Just ignore this comment.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
you-leee: Yeah, I would do the explicit check, since the filter is not used in `executePrivilegesUpdate`, and that is a public function.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,369 @@
...
+        val userName = getDBUser(userPrincipal.id)
+        //edit the pg_hba file through some magic. must. become. magician.
+    }
+
+    /**
+     * Sets privileges for a user on an organization's table or column
+     */
+    fun executePrivilegesUpdate(action: Action, acls: List<Acl>) {
+        val aclsByOrg = acls.groupBy { it.aclKey[0] }
+        aclsByOrg.forEach { (orgId, acls) ->
```
makosblade: Pretty sure `acls` here shadows `acls` method parameter

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
you-leee: Ignore the first part of the comment, missed, that TABLE_ID_INDEX was already added to the mapstore.. So just basically using the constant here is preferred

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: intellij isn't complaining, but to be safe I'll change it!

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: 👍

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -0,0 +1,378 @@
...
+            privilegesByUser.getOrPut(user) { mutableSetOf() }.add(privilege)
+        }
+        privilegesByUser.forEach { (user, privileges) ->
+            val principal = securePrincipalsManager.getPrincipal(user).principal
+            val aceKey = AceKey(aclKey, principal)
+            var permissions = EnumSet.noneOf(Permission::class.java)
+            if (privileges == OwnerPrivileges.values() as Set<String>) { //eek
+                permissions.add(Permission.OWNER)
+            }
+            if (privileges.contains("SELECT")) {
```
you-leee: You could use `OwnerPrivileges` also here, but maybe with renaming them as PostgresTablePrivileges or something like that. Otherwise, if the only purpose of that enum class is to check, if the user has all of those privileges, I would move it to this class as a set of string.

pckeyes: This makes so much more sense, I'll change that

src/main/kotlin/com/openlattice/organizations/mapstores/OrganizationExternalDatabaseColumnMapstore.kt
```diff
@@ -0,0 +1,79 @@
...
+import com.openlattice.organization.OrganizationExternalDatabaseColumn
+import com.openlattice.postgres.PostgresTable
+import com.openlattice.postgres.ResultSetAdapters
+import com.openlattice.postgres.mapstores.AbstractBasePostgresMapstore
+import com.zaxxer.hikari.HikariDataSource
+import java.sql.PreparedStatement
+import java.sql.ResultSet
+import java.util.*
+
+const val ID_INDEX = "__key"
```
you-leee: Just a minor thing, but you can use `QueryConstants.KEY_ATTRIBUTE_NAME.value` here. That enum has the 2 constant index names for hazelcast maps.

src/main/kotlin/com/openlattice/organizations/mapstores/OrganizationExternalDatabaseColumnMapstore.kt
```diff
@@ -0,0 +1,79 @@
...
+        return ResultSetAdapters.id(rs)
+    }
+
+    override fun mapToValue(rs: ResultSet?): OrganizationExternalDatabaseColumn {
+        return ResultSetAdapters.organizationExternalDatabaseColumn(rs)
+    }
+
+    override fun getMapConfig(): MapConfig {
+        return super.getMapConfig()
+                .addMapIndexConfig(MapIndexConfig(ID_INDEX, false))
```
you-leee: You don't need to add an index for the key of the map

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -737,8 +737,6 @@ class AssemblerConnectionManager(
             connection.createStatement().use { statement ->
                 logger.info("Granting usage on $MATERIALIZED_VIEWS_SCHEMA schema and revoking from $PUBLIC_SCHEMA schema for users: $userIds")
                 statement.execute("GRANT USAGE ON SCHEMA $MATERIALIZED_VIEWS_SCHEMA TO $userIdsSql")
-                //Don't allow users to access public schema which will contain foreign data wrapper tables.
-                statement.execute("REVOKE USAGE ON SCHEMA $PUBLIC_SCHEMA FROM $userIdsSql")
```
you-leee: You will need a mechanic job that grants usage on public schema for each member within the organization database, so that they will have access to public schema.

src/main/java/com/openlattice/authorization/HazelcastSecurableObjectResolveTypeService.java
```diff
@@ -59,4 +60,17 @@ public SecurableObjectType getSecurableObjectType( AclKey aclKey ) {
         return securableObjectTypes
                 .keySet( Predicates.equal( SecurableObjectTypeMapstore.SECURABLE_OBJECT_TYPE_INDEX, type ) );
     }
+
+    @Override
+    public Set<AclKey> getOrganizationExternalDatabaseAclKeys( List<Acl> acls ) {
+        return securableObjectTypes.keySet(
```
you-leee: I think this is missing to match with keys of the `acls`

pckeyes: Oh yes! Will fix.

pckeyes:
A step that (I'm pretty sure) I don't have authorization to do is to set the paths to the pg_hba.conf file  in pg_hba_config.yaml for the prod version of this yaml, as I don't know where that file lives on prod.

src/main/kotlin/com/openlattice/organizations/mapstores/OrganizationExternalDatabaseColumnMapstore.kt
pckeyes: I tried to do this and kotlin was unhappy. Seems to think that `QueryConstants.KEY_ATTRIBUTE_NAME.value()` is not a constant.

you-leee: You don't need to put it into a constant anymore I think. But otherwise, you can just use the QueryConstants anywhere instead of creating a new field that references it

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
pckeyes: pr opened in mechanic :)

src/main/kotlin/com/openlattice/organizations/mapstores/OrganizationExternalDatabaseColumnMapstore.kt
pckeyes: replaced other constant with QueryConstant!

PR #647, opened 2019-10-08
lock ids before updating entities, complete writes in single transaction

katherinebernstein:


src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -238,12 +238,21 @@ class PostgresEntityDataQueryService(
             entitySetId: UUID,
             entities: Map<UUID, Map<UUID, Set<Any>>>,
             authorizedPropertyTypes: Map<UUID, PropertyType>,
-            awsPassthrough: Boolean = false
+            awsPassthrough: Boolean = false,
+            idsAlreadyLocked: Boolean = false
```
you-leee: Can you pls update the doc with this param?

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -261,14 +270,13 @@ class PostgresEntityDataQueryService(
     private fun upsertEntities(
             connection: Connection,
             entitySetId: UUID,
+            partitionsInfo: PartitionsInfo,
```
you-leee: Same comment here

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -758,6 +768,29 @@ class PostgresEntityDataQueryService(
         return WriteEvent(System.currentTimeMillis(), numUpdates)
     }
 
+    private fun lockEntitiesForUpdate(connection: Connection, entityKeyIds: Set<UUID>, partitionsInfo: PartitionsInfo) {
+        // Acquire entity key id locks
+        lockEntitiesForUpdate(
+                connection,
+                PostgresArrays.createUuidArray(connection, entityKeyIds),
+                PostgresArrays.createIntArray(connection, partitionsInfo.partitions),
```
you-leee: Shouldn't we also filter partitions by entity key ids?
`entityKeyIds.map { getPartition(it, partitionsInfo .partitions) }`

katherinebernstein: +1 updated for this

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -240,17 +241,14 @@ class PostgresEntityDataQueryService(
...
             awsPassthrough: Boolean = false
     ): WriteEvent {
-        return hds.connection.use { connection ->
-            val writeEvent = upsertEntities(connection, entitySetId, entities, authorizedPropertyTypes, awsPassthrough)
-//            connection.commit()
-//            connection.autoCommit = true
-            return@use writeEvent
-        }
+
+        return upsertEntities(entitySetId, null, entities, authorizedPropertyTypes, awsPassthrough)
```
geekbeast: Don't use null use a no-op function, which will be elided by the compiler anyway and avoid the null check.

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -758,6 +762,19 @@ class PostgresEntityDataQueryService(
         return WriteEvent(System.currentTimeMillis(), numUpdates)
     }
 
+    private fun lockEntitiesForUpdate(
+            connection: Connection,
+            idsArr: java.sql.Array,
+            partition: Int,
+            partitionVersion: Int) {
+        // Acquire entity key id locks
+        val rowLocks = connection.prepareStatement(lockEntitiesSql)
```
geekbeast: As a safety check it might be a good idea to ensure that connection.autoCommit = false here, because locks are released at end of transaction.

geekbeast: This way we can catch unexpected invocations.

PR #652, opened 2019-10-12
- do json validation outside of connection

katherinebernstein:
- partition forEach in parallel (do we want this?)


katherinebernstein:
This parallelizes the upserts -- not sure if we want this due to connection limitations. Another option would be to keep the parallelization and move the semaphore from DataController to here.

katherinebernstein:
@geekbeast Right but since we're limited to 12 connections, if there are 10 requests that each write to 4 partitions, we'll likely get connection timeouts (even with the semaphore in DataController)

PR #653, opened 2019-10-16
Start range at one past max found id

geekbeast:


src/main/kotlin/com/openlattice/ids/IdCatchupEntryProcessor.kt
```diff
@@ -25,26 +24,26 @@ class IdCatchupEntryProcessor(hds: HikariDataSource) : AbstractRhizomeEntryProce
...
+                    ps.setObject(2, supremum)
 
-                    while (exists(ps, range.peek())) {
-                        range.nextId()
-                        counter++
-                        if ((counter % 10000) == 0) {
-                            logger.info("Incremented range with base {} by {}", range.base, counter)
-                        }
-                    }
+                    val id = id(ps.executeQuery())
```
you-leee: I think we need a ps.next() here to get the first result. 
Also, is it possible, that no id is in that range?

src/main/kotlin/com/openlattice/ids/IdCatchupEntryProcessor.kt
```diff
@@ -25,26 +24,26 @@ class IdCatchupEntryProcessor(hds: HikariDataSource) : AbstractRhizomeEntryProce
         private val logger = LoggerFactory.getLogger(IdCatchupEntryProcessor::class.java)
     }
 
-    override fun process(entry: MutableMap.MutableEntry<Int, Range>): Void? {
-        val range = checkNotNull(entry.value) //Range should never be null in the EP.
+    override fun process(entry: MutableMap.MutableEntry<Int, Range?>): Void? {
+        val range = entry.value!! //Range should never be null in the EP.
         var counter = 0
         val x = 1L shl 2
```
you-leee: This is not used anywhere

PR #655, opened 2019-10-21
LATTICE-2253 double max on last_write

you-leee:


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -105,7 +110,13 @@ internal fun selectEntitiesGroupedByIdAndPropertyTypeId(
  * Returns the correspondent column name used for the metadata option with a comma prefix.
  */
 private fun mapMetaDataToColumnSql(metadataOption: MetadataOption): String {
-    return ",${mapMetaDataToColumnName(metadataOption)}"
+    return when (metadataOption) {
```
katherinebernstein: does `mapMetaDataToColumnName` not need to be updated everywhere for this updated LAST_WRITE value?

you-leee: Nope, that is just mapping from MetadataOption -> PostgresColumn

PR #656, opened 2019-10-22
Bugfix/delete refactor and fixes

katherinebernstein:


src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -487,6 +488,49 @@ class Graph(
         return neighbors
     }
 
+    override fun getEdgeEntitySetsConnectedToEntities(entitySetId: UUID, entityKeyIds: Set<UUID>): Set<UUID> {
+        val query = "SELECT DISTINCT ${EDGE_ENTITY_SET_ID.name} " +
+                "FROM ${E.name} " +
+                "WHERE (${SRC_ENTITY_SET_ID.name} = ? AND ${SRC_ENTITY_KEY_ID.name} = ANY(?) " +
```
you-leee: Just a minor thing, but can we use `SRC_IDS_SQL` and `DST_IDS_SQL` here?

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -228,59 +196,6 @@ open class DataGraphService(
...
-
-        // delete entities
-        val entityWriteEvent = eds.deleteEntities(entitySetId, entityKeyIds, authorizedPropertyTypes)
-
-        logger.info("Deleted {} entities and {} vertices.", entityWriteEvent.numUpdates, verticesCount)
-
-        return entityWriteEvent
-    }
-
     override fun deleteAssociationsBatch(
```
you-leee: I don't think this is used anymore, right? (same for clearAssociationsBatch)

makosblade:
If these are public APIs then we probably should update the docs as well

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -487,6 +488,49 @@ class Graph(
...
+                "OR (${DST_ENTITY_SET_ID.name} = ? AND ${DST_ENTITY_KEY_ID.name} = ANY(?) )"
+
+        return PostgresIterable(
```
makosblade: I *think* pref BPGI b/c it's the Kotlin version ¯\_(ツ)_/¯

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -487,6 +488,49 @@ class Graph(
...
+        ).toSet()
+    }
+
+
+    override fun getEdgeEntitySetsConnectedToEntitySet(entitySetId: UUID): Set<UUID> {
+        val query = "SELECT DISTINCT ${EDGE_ENTITY_SET_ID.name} " +
+                "FROM ${E.name} " +
+                "WHERE ${SRC_ENTITY_SET_ID.name} = ? OR ${DST_ENTITY_SET_ID.name} = ?"
+
+        return PostgresIterable(
```
makosblade: BPGI

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
```diff
@@ -0,0 +1,434 @@
...
+            entitySetId: UUID,
+            entityKeyIds: Set<UUID>,
+            deleteType: DeleteType,
+            principals: Set<Principal>
+    ): WriteEvent {
+        return clearOrDeleteEntities(entitySetId, entityKeyIds, deleteType, principals, skipAuthChecks = false)
+    }
+
+
+    // TODO rewrite this from scratch.
```
makosblade: 😢

src/main/kotlin/com/openlattice/graph/Graph.kt
katherinebernstein: ooh true!

src/main/kotlin/com/openlattice/graph/Graph.kt
katherinebernstein: +yes

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
```diff
@@ -0,0 +1,434 @@
...
+                .collect(Collectors.toSet())
+
+        if (unauthorizedAclKeys.isNotEmpty()) {
+            throw ForbiddenException("Unable to delete from entity set $entitySetId: missing required permissions " +
+                    "on associations for AclKeys $unauthorizedAclKeys")
+        }
+
+    }
+
+    private fun getAuthorizedPropertyTypesForDelete(
```
you-leee: Just a minor thing, but shouldn't we do the authorization checks in the controllers?

katherinebernstein: Theoretically yes, although the auth logic is pretty deep into the delete logic in multiple spots, and moving it into the controller would involve repeating a ton of logic in multiple places (which led to copy/paste and maintenance errors in our current implementation) as well as duplicating calls unnecessarily, so it seemed safer to me to add an auth-aware delete function in DataDeletionService.

you-leee: 👍

PR #657, opened 2019-10-22
Debugging/oh lawg

makosblade:
This PR:
- Updates a frequently used call (getLinkingIdsOfEntityKeyIds) to be used in a more centralized manner
- Refactors use pattern to prevent re-calling when the information is already in memory
- Makes invocations use partition instead of only using ekid


makosblade:
The merging or rebasing is impossible lolol

PR #660, opened 2019-10-30
fix updateVersionsForPropertyTypesInEntitiesInEntitySet for linking case

you-leee:


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -401,9 +401,10 @@ internal val updateVersionsForPropertyTypesInEntitySet = "$updateVersionsForProp
  * 8. partition version
  */
 internal fun updateVersionsForPropertyTypesInEntitiesInEntitySet( linking: Boolean = false ): String {
-    return "$updateVersionsForPropertyTypesInEntitySet " +
-            if( linking ) "AND ${ORIGIN_ID.name} = ANY(?) " else "AND ${ID_VALUE.name} = ANY(?) " +
-            "AND ${PARTITION.name} = ANY(?) AND ${PARTITIONS_VERSION.name} = ? "
+    val idsSql = if( linking ) "AND ${ORIGIN_ID.name} = ANY(?)" else "AND ${ID_VALUE.name} = ANY(?)"
```
katherinebernstein: I think it's easier to read if/else statements on multiple lines

you-leee: Will do!

PR #665, opened 2019-11-05
Refine reading and writing of data.

geekbeast:
- Simplified query for reading data.
- Auto-commit is enabled for writing with proper transactional mechanics.
- Finalizing a write transaction propagates writes to linked entities. This code will also be useful in the future when needing to propagate changes from items that are linked together.

Due to changes in aggregation we know longer need to accumulate properties in datastore as this handled by Postgres. I left in the group by origin_id for backwards compatibility, but this is a short term fix as it is not ideal to have a read path just for linking indexing which doesn't reflect data this is loaded. Long term, I think linking indexing should be able to use aggregation over detail objects in order to compute individual components for indexing. Not sure why this is necessary.


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -281,6 +279,71 @@ fun optionalWhereClausesSingleEdk(
...
+            PARTITIONS_VERSION.name
+    ).joinToString(",")
+
+    val conflictClause = (metadataColumns + dataTableValueColumns).joinToString(
+            ","
+    ) { "${it.name} = EXCLUDED.${it.name}" }
+
+    return "WITH linking_map as ($upsertEntitiesSql) INSERT INTO ${DATA.name} ($metadataColumnsSql,$insertColumns) " +
+            "SELECT $metadataReadColumnsSql,$insertColumns FROM ${DATA.name} INNER JOIN " +
+            "linking_map USING(${ENTITY_SET_ID.name},${ID.name},${PARTITION.name}) " +
```
geekbeast: Should consider caching a filtered CTE here, if perf becomes an issue.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -58,51 +64,43 @@ fun buildPreparableFiltersSql(
...
-        " AND ${ORIGIN_ID.name} IS NOT NULL AND ${ORIGIN_ID.name} != '${IdConstants.EMPTY_ORIGIN_ID.id}' "
-    } else {
-        ""
-    }
+    val metadataOptionColumnsSql = metadataOptionColumns.values.joinToString("")
+    val innerGroupBy = groupBy(ESID_EKID_PART_PTID)
+    val outerGroupBy = if (metadataOptions.contains(MetadataOption.ORIGIN_IDS)) {
+        groupBy(ESID_EKID_PART)
+    } else groupBy("$ESID_EKID_PART,${ORIGIN_ID.name}")
+    val linkingClause = if (linking) " AND ${ORIGIN_ID.name} != '${ID_VALUE.name}' " else ""
```
you-leee: Maybe I'm missing here something, but right now, this will return true for both linking and non-linking entities + `upsertPropertyValueSql` did not change in a way, that it would insert the normal entity key id to the `origin_id` field.

src/main/kotlin/com/openlattice/postgres/PostgresDataTables.kt
```diff
@@ -142,11 +142,15 @@ class PostgresDataTables {
                     .name(quote(prefix + "id_version_idx"))
                     .ifNotExists()
                     .desc()
-
-            val originIdNotNullIndex = PostgresExpressionIndexDefinition(tableDefinition, "(${ORIGIN_ID.name} != '${IdConstants.EMPTY_ORIGIN_ID.id}')")
-                    .name(quote(prefix + "origin_id_not_equal_empty_idx"))
+          
+            val readDataIndex = PostgresExpressionIndexDefinition(tableDefinition, "(${ORIGIN_ID.name} != '${IdConstants.EMPTY_ORIGIN_ID.id}')" )
```
you-leee: These two indices are the same.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -58,51 +64,43 @@ fun buildPreparableFiltersSql(
...
-    val innerGroupBy = groupBy(ESID_EKID_PART_PTID + nonAggregatedMetadataSql)
-    // TODO: remove IS NOT NULL post-migration
-    val linkingClause = if (linking) {
-        " AND ${ORIGIN_ID.name} IS NOT NULL AND ${ORIGIN_ID.name} != '${IdConstants.EMPTY_ORIGIN_ID.id}' "
-    } else {
-        ""
-    }
+    val metadataOptionColumnsSql = metadataOptionColumns.values.joinToString("")
+    val innerGroupBy = groupBy(ESID_EKID_PART_PTID)
+    val outerGroupBy = if (metadataOptions.contains(MetadataOption.ORIGIN_IDS)) {
```
you-leee: This should be negated

you-leee: `PROPERTIES` should be in group by or aggregated in outer query

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
geekbeast: This needs a migration and I did forget to insert normal entity key is upsertEntities

src/main/kotlin/com/openlattice/postgres/PostgresDataTables.kt
geekbeast: bad merge

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
geekbeast: +1

geekbeast: Missed a jsonb_object_agg

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -340,112 +345,68 @@ class PostgresEntityDataQueryService(
...
+
+            //Make data visible by marking new version in ids table.
+            val upsertEntities = connection.prepareStatement(buildUpsertEntitiesAndLinkedData())
+
+            upsertEntities.setObject(1, versionsArrays)
+            upsertEntities.setObject(2, version)
+            upsertEntities.setObject(3, version)
+            upsertEntities.setObject(4, entitySetId)
+            upsertEntities.setArray(5, entityKeyIdsArr)
+            upsertEntities.setInt(6, partition)
```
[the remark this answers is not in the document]
geekbeast: I'll fix this after get back from Sacramento. Will require a second CTE query to compute the partitions of all the linking ids.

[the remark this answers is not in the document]
geekbeast: Never mind. This should have been done this way from the beginning. I vaguely recall discussing this and wanting to use the same partition since it co-locates linking related operations.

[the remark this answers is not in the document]
geekbeast: (a) This was easier than expected and actually ended up removing that weird read before write. (b) delete doesn't have an implementation that removes linked entities, should be a separate PR/bugfix. (c) We're still doing two statements tombstones, we should switch to one statement tombstones but that is a separate ticket.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -58,51 +65,44 @@ fun buildPreparableFiltersSql(
...
-    val metadataOptionColumnsSql = metadataOptionColumns
-            .values.joinToString("")
-    // TODO don't filter out LAST_WRITE from group by after empty rows are eliminated  https://jira.openlattice.com/browse/LATTICE-2254
-    val metadataOptionOuterGroupBySql = metadataOptionColumns
-            .filter { it.key != MetadataOption.LAST_WRITE }
-            .values.joinToString("")
-    val outerGroupBy = groupBy(ESID_EKID_PART + metadataOptionOuterGroupBySql)
-    val sql = "SELECT ${ENTITY_SET_ID.name},${ID_VALUE.name},${PARTITION.name}$metadataOptionColumnsSql,$jsonValueColumnsSql " +
+    val sql = "SELECT ${ENTITY_SET_ID.name},${ID_VALUE.name},${PARTITION.name}$metadataOptionColumnsSql," +
+            "jsonb_object_agg($PROPERTY_TYPE_ID,PROPERTIES) " +
```
you-leee: `PROPERTY_TYPE_ID.name` should be here I think

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -58,51 +65,44 @@ fun buildPreparableFiltersSql(
...
-        " AND ${ORIGIN_ID.name} IS NOT NULL AND ${ORIGIN_ID.name} != '${IdConstants.EMPTY_ORIGIN_ID.id}' "
-    } else {
-        ""
-    }
+    val metadataOptionColumnsSql = metadataOptionColumns.values.joinToString("")
+    val innerGroupBy = groupBy(ESID_EKID_PART_PTID)
+    val outerGroupBy = if (metadataOptions.contains(MetadataOption.ORIGIN_IDS)) {
+        groupBy("$ESID_EKID_PART,${ORIGIN_ID.name}")
+    } else groupBy(ESID_EKID_PART)
+    val linkingClause = if (linking) " AND ${ORIGIN_ID.name} != '${IdConstants.EMPTY_ORIGIN_ID}' " else ""
```
you-leee: It should be `IdConstants.EMPTY_ORIGIN_ID.id`

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -58,51 +65,44 @@ fun buildPreparableFiltersSql(
...
-            .filter { !isMetaDataAggregated(it.key) }.values.joinToString("")
-    val innerGroupBy = groupBy(ESID_EKID_PART_PTID + nonAggregatedMetadataSql)
-    // TODO: remove IS NOT NULL post-migration
-    val linkingClause = if (linking) {
-        " AND ${ORIGIN_ID.name} IS NOT NULL AND ${ORIGIN_ID.name} != '${IdConstants.EMPTY_ORIGIN_ID.id}' "
-    } else {
-        ""
-    }
+    val metadataOptionColumnsSql = metadataOptionColumns.values.joinToString("")
+    val innerGroupBy = groupBy(ESID_EKID_PART_PTID)
```
you-leee: If ORIGIN_ID metadata option is true, it should be either in this group by or aggregated in the inner query

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -58,51 +65,44 @@ fun buildPreparableFiltersSql(
...
+    val outerGroupBy = if (metadataOptions.contains(MetadataOption.ORIGIN_IDS)) {
```
you-leee: If ENTITY_KEY_IDS metadata option is set, it should be either in this group by or aggregated in the outer query.
It is used for materialized views btw, to collect the origin ids for linking entities. Which will probably need another discussion because of the re-creation of linked entities shouldn't be possible, but otherwise neighbor queries are not possible in materialized views.

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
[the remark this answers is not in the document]
makosblade: (c) https://jira.openlattice.com/browse/LATTICE-2281

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -23,14 +24,20 @@ import java.util.*
...
 const val VALUES = "values"
+const val PROPERTIES = "properties"
 
 val dataTableColumnsSql = PostgresDataTables.dataTableColumns.joinToString(",") { it.name }
 
-val valuesColumnsSql = PostgresDataTables.dataTableValueColumns.joinToString(",") {
-    "array_agg(${it.name}) FILTER (where ${it.name} IS NOT NULL) as ${it.name}"
-}
+val detailedValueColumnsSql = dataTableValueColumns.joinToString("||") {
+    "jsonb_agg(json_build_object('value',${it.name},'${ENTITY_SET_ID.name}', entity_set_id, '${ID_VALUE.name}', origin_id)) FILTER (where ${it.name} IS NOT NULL)"
```
katherinebernstein: Seems safer/clearer to use the PostgresColumn names everywhere instead of mixing them with hardcoded values. Also I think this would be a bit easier to read if it were written as 
```
"jsonb_agg(json_build_object(
  'value', ${it.name},
  '${ENTITY_SET_ID.name}', ${ENTITY_SET_ID.name},
  '${ID_VALUE.name}', ${ORIGIN_ID.name}
)) FILTER (
  WHERE ${it.name} IS NOT NULL
)"
```

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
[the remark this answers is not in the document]
katherinebernstein: This will also require changes in PostgresLinkingQueryService

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
katherinebernstein: +1, and also `$PROPERTIES`

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
geekbeast: ORIGIN_ID makes no sense outside of linking and shouldn't really be a metadata option. Will probably switch this if statement to be based on linking.

Long term we need to fix linking indexing to handle the new read format and not read data in the current format (which I'm not changing in this PR).

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
geekbeast: As much as I would love to fix all the issues, this wasn't implemented previously (at least not correctly) and reading linked entities is currently not implemented correctly either. You can see these in the code to read linked entities which has to re-shuffle all the data for every single read call in memory after read each component entry individually. 

I have aggregated origin_ids in and aliased the column in the out call, but I'm not sure this will fix the complete problem for materialized views.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
geekbeast: We're missing tons of validation here. This is extending a hack that will have to be removed.

geekbeast: For example, ORIGIN_IDS should not be specifiable metadata independent of linking, and it makes no sense that ORIGIN_IDS are separate concept from the detailed view of reading an object.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
makosblade: Second

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -114,7 +114,7 @@ private fun mapMetaDataToColumnSql(metadataOption: MetadataOption): String {
         MetadataOption.ORIGIN_IDS -> ",${ORIGIN_ID.name}"
         // TODO should be just last_write with comma prefix after empty rows are eliminated https://jira.openlattice.com/browse/LATTICE-2254
         MetadataOption.LAST_WRITE -> ",max(${LAST_WRITE.name}) AS ${mapMetaDataToColumnName(metadataOption)}"
-        MetadataOption.ENTITY_KEY_IDS -> ",${ENTITY_KEY_IDS_COL.name}"
+        MetadataOption.ENTITY_KEY_IDS -> ",array_agg(${ORIGIN_ID.name}) as ${ENTITY_KEY_IDS_COL.name})"
```
you-leee: This won't work unless we always select origin_id in inner query. I think it should be ",array_agg(${ENTITY_KEY_IDS_COL.name}) as ${ENTITY_KEY_IDS_COL.name})"

geekbeast: It will work since it will be in inner query if metadata option is provided.

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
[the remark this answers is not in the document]
geekbeast: +1

PR #666, opened 2019-11-07
refresh materialized view instead of drop/recreate

katherinebernstein:


geekbeast:
However it should be less of a problem now that we're never changing the query.

PR #669, opened 2019-11-09
 bugfix/lattice 2284 retry s3 uploads

pckeyes:
This pr adds retry configurations to the s3 client and implements aws's TransferManager for multipart uploads on large (> 16MB) uploads


src/main/kotlin/com/openlattice/data/storage/aws/AwsBlobDataService.kt
```diff
@@ -43,8 +53,10 @@ class AwsBlobDataService(
         metadata.contentLength = dataInputStream.available().toLong()
         metadata.contentType = contentType
         val putRequest = PutObjectRequest(datastoreConfiguration.bucketName, s3Key, dataInputStream, metadata)
-        s3.putObject(putRequest)
-
+        val transferManager = TransferManagerBuilder.standard().withS3Client(s3).build()
```
makosblade: I think you want a `.withMultiPartUploadThreshold(16_000_000L)` here

pckeyes: I believe TransferManager defaults to do multipart upload at 16MB so it doesn't need to be explicitly set

makosblade: Ah yeah they don't put that in the comments anywhere in TransferManager or TMBuilder 🤦‍♂

geekbeast:
This is good to go-- just waiting on benchmarks and will merge.

PR #683, opened 2019-11-20
Bugfix/lattice 2284 s3 upload retries

pckeyes:
uses TransferManager for retrying uploads and multipart uploads for large objects
implements benchmark testing for puts and gets


src/main/kotlin/com/openlattice/data/storage/aws/AwsBlobDataService.kt
```diff
@@ -43,8 +57,10 @@ class AwsBlobDataService(
         metadata.contentLength = dataInputStream.available().toLong()
         metadata.contentType = contentType
         val putRequest = PutObjectRequest(datastoreConfiguration.bucketName, s3Key, dataInputStream, metadata)
-        s3.putObject(putRequest)
-
+        val transferManager = TransferManagerBuilder.standard().withS3Client(s3).build()
```
geekbeast: This seems like it should be a class variable and not instantiated on every single upload call.

pckeyes: It does seem like that. But you have to close the transfer manager after every use (or at least that is recommended) so there can't just be one instance of it.

geekbeast: Yuck.

PR #687, opened 2019-11-26
Miscellaneous fixes.

geekbeast:


src/main/kotlin/com/openlattice/conductor/rpc/ConductorConfiguration.kt
```diff
@@ -0,0 +1,36 @@
...
+private val SEARCH_CONFIGURATION_FIELD = "searchConfiguration"
+private val key = SimpleConfigurationKey("conductor.yaml")
+
+/**
+ *
+ * @author Matthew Tamayo-Rios &lt;matthew@openlattice.com&gt;
+ */
+@ReloadableConfiguration(uri = "conductor.yaml")
+class ConductorConfiguration(
+        @JsonProperty("reportEmailAddress") val reportEmailAddress: String,
```
katherinebernstein: should this be using the constants defined above? also do we need to update existing conductor.yamls for the new connection param?

src/main/kotlin/com/openlattice/conductor/rpc/ConductorConfiguration.kt
```diff
@@ -0,0 +1,36 @@
...
+) : Configuration {
+    @JsonIgnore
+    override fun getKey(): ConfigurationKey {
+        return key
+    }
+
+    companion object {
+        @JvmStatic
+        fun key(): ConfigurationKey {
+            return key
```
you-leee: Won't this go into an infinite loop?

you-leee: I just remembering having this issue in IndexerConfiguration and that's why I renamed it there to configKey

katherinebernstein: ah yeah I remember this fun kotlin experience

geekbeast: I am not having this problem? I wonder if this has been fixed in the latest version of kotlin.

src/main/kotlin/com/openlattice/conductor/rpc/ConductorConfiguration.kt
geekbeast: I think changed this to optional. If you do specify, it will add all the users to the global org. I think we found a better solution than this.

geekbeast: Also for kotlin data files defining constants inline for non-reused fields is perfectly fine. You don't have to worry about synchronization with getters and setters. Obviously, for re-used fields they should be in SerializatonConstants.

katherinebernstein: Oh my bad -- I thought I saw unused constants defined for these values right above the constructor. I must have been looking at the java version or something.

PR #689, opened 2019-11-26
Serializer upd8

makosblade:
This PR:
- Add an OrganizationStreamSerializer that serializes everything independently instead of using jackson
- Add a serializer for GrantType
- Add an AbstractEnumSerializer that uses ORDINAL (so don't change the ORDer of enum entries using this)
- Add a Grant Serializer
Depends on ab8a658b388e554ca232df0cffc1e475f57f467e in Rhizome for some more serializers


src/main/kotlin/com/openlattice/hazelcast/serializers/AbstractEnumSerializer.kt
```diff
@@ -0,0 +1,36 @@
+package com.openlattice.hazelcast.serializers
+
+import com.google.common.cache.CacheBuilder
+import com.google.common.cache.CacheLoader
+import com.google.common.cache.LoadingCache
+import com.hazelcast.nio.ObjectDataInput
+import com.hazelcast.nio.ObjectDataOutput
+import com.kryptnostic.rhizome.pods.hazelcast.SelfRegisteringStreamSerializer
+
+abstract class AbstractEnumSerializer<T : Enum<T>> : SelfRegisteringStreamSerializer<Enum<T>> {
```
katherinebernstein: :D

makosblade:
> it would probably be good to have stream serializer tests for these

Huh, didn't we used to have `TestableSelfRegisteringStreamSerializer` that would atuo-test 🤔

katherinebernstein:
> TestableSelfRegisteringStreamSerializer

maaaybe? We have `AbstractStreamSerializerTest` which can be extended very easily, at least

makosblade:
I added a `TestableSelfRegisteringStreamSerializer` tester and guess what, all of this is broken 😆

geekbeast:
Only mapstores have auto-test.

Also, we had an organization stream serializer, I deleted it when I made organization object more complex, because of the deep map nesting. Glad to have it back!

PR #691, opened 2019-11-27
Feature/lattice 2255 partitions version is dead long live partitions version

katherinebernstein:


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -378,7 +376,7 @@ val upsertEntitiesSql = "UPDATE ${IDS.name} " +
  * 3. partition version
```
you-leee: Doc needs to be updated

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -78,9 +78,9 @@ class Graph(
     /* Create */
 
     override fun createEdges(keys: MutableSet<DataEdgeKey>): WriteEvent {
-        val partitionsInfoByEntitySet = partitionManager.getEntitySetsPartitionsInfo(keys.flatMap { listOf(it.src, it.dst, it.edge) }
+        val partitionsInfoByEntitySet = partitionManager.getEntitySetsPartitions(keys.flatMap { listOf(it.src, it.dst, it.edge) }
```
pckeyes: Super trivial, but since `PartitionsInfo` is no longer a thing, maybe just call this `partitionsByEntitySet`?

PR #692, opened 2019-11-28
Fix/warnings

makosblade:
This PR:
- Fixes a bunch of mechanical warnings
- Removing unused imports
- Removing unnecessary boxing/unboxing
- Uses singletonList for one-element lists
- Uses Java classes instead of Guava for functional constructs
- Fixes compiler warnings for upgrading from Java 5 through Java 8
- Fixes several performance oriented warnings
- Removes unnecessary wrapper types
- Removes unnecessary null checking


makosblade:
Depends on https://github.com/geekbeast/rhizome/pull/288

PR #698, opened 2019-12-04
Read offsetdatetime from postgresql correctly

geekbeast:


src/main/kotlin/com/openlattice/postgres/PostgresResultSetAdapters.kt
```diff
@@ -177,7 +177,7 @@ fun readJsonDataColumnsWithId(
                 }
 
                 if (metadataOptions.contains(MetadataOption.LAST_WRITE)) {
-                    val lastWrite = rs.getObject(PostgresColumn.LAST_WRITE_FIELD) as OffsetDateTime
+                    val lastWrite = lastWriteTyped(rs)
```
you-leee: Actually, if we get the last_write by row,  =this can be placed before the forEach loop

PR #704, opened 2019-12-06
remove remaining locking logic in PEDQS

katherinebernstein:


makosblade:
Though I should point out this:
- delete properties call X,Y,Z
- add properties X
- add properties in postgres X
- delete properties in postgres X, Y, Z
Is possible if someone does a delete during an integration (I think?)

geekbeast:
That is far more likely to arise during integrations, but integration tombstones before performing a write over properties. So not that big of a deal when it happens then (also shouldn't be integrated into user editable datasets, but that's another work item).

The long term fix here is to do the tombstone and update in a single transaction. This should be doable via a CTE which tombstones + primary update query.

PR #706, opened 2019-12-09
clearEntities throws errors because it expects autoCommit to be off

katherinebernstein:


makosblade:
Don't we definitely need to combine these into one query?
You could have a bit of a wait between the two queries depending on the scheduler

katherinebernstein:
> Don't we definitely need to combine these into one query?
> You could have a bit of a wait between the two queries depending on the scheduler

Combine what into one query? This looks like it was just leftover from the locking code, but since locking was removed autoCommit will never be false here, so soft deletes are just always broken

makosblade:
> > Don't we definitely need to combine these into one query?
> > You could have a bit of a wait between the two queries depending on the scheduler
> 
> Combine what into one query? This looks like it was just leftover from the locking code, but since locking was removed autoCommit will never be false here, so soft deletes are just always broken

The two Postgres queries should be combined because the thread running them can block in-between the queries resulting in potential badness/inconsistencies

makosblade:
https://jira.openlattice.com/browse/LATTICE-2338

PR #707, opened 2019-12-09
remove code iterating over PostgresIterable multiple times + option t…

katherinebernstein:
…o filter by normal entity set ids when loading normal entity key ids of linking ids


src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
```diff
@@ -291,16 +291,22 @@ class PostgresLinkingQueryService(private val hds: HikariDataSource, private val
...
+            normalEntitySetIds: Set<UUID>?
+    ): PostgresIterable<Pair<UUID, Set<UUID>>> {
         return PostgresIterable(
                 Supplier {
                     val connection = hds.connection
-                    val ps = connection.prepareStatement(ENTITY_KEY_IDS_OF_LINKING_IDS_SQL)
+                    val ps = connection.prepareStatement(buildEntityKeyIdsOfLinkingIdsSql(normalEntitySetIds != null))
                     val linkingIdsArray = PostgresArrays.createUuidArray(connection, linkingIds)
 
                     val partitions = PostgresArrays.createIntArray( connection, getPartitionsInfo( linkingIds, partitionManager.getAllPartitions() ))
```
you-leee: This should be probably a separate ticket (if it's not already existing) but since we changed partitions for linked entities to be mapped by their normal entity key ids, this won't give any results.
In case, when entity set ids are non-null, I think it's safe to use all partitions for those entity sets (since the actual partitions of the linking entities will be a subset of those)

you-leee: Btw I couldn't find a ticket for this

katherinebernstein: now there is one https://jira.openlattice.com/browse/LATTICE-2337

src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
```diff
@@ -291,16 +291,20 @@ class PostgresLinkingQueryService(private val hds: HikariDataSource, private val
...
+    ): PostgresIterable<Pair<UUID, Set<UUID>>> {
         return PostgresIterable(
                 Supplier {
                     val connection = hds.connection
-                    val ps = connection.prepareStatement(ENTITY_KEY_IDS_OF_LINKING_IDS_SQL)
+                    val ps = connection.prepareStatement(buildEntityKeyIdsOfLinkingIdsSql(normalEntitySetIds != null))
                     val linkingIdsArray = PostgresArrays.createUuidArray(connection, linkingIds)
 
-                    val partitions = PostgresArrays.createIntArray( connection, getPartitionsInfo( linkingIds, partitionManager.getAllPartitions() ))
-                    ps.setArray(1, partitions)
```
dbt: is there a reason why we don't need/can't use partition id here anymore?

src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
```diff
@@ -357,6 +361,24 @@ internal fun buildFilterEntityKeyPairs(entityKeyPairs: Collection<EntityKeyPair>
...
+ * Returns SQL to select normal entity key ids of linking ids. Bind order is as folows:
+ *
+ * 1. linkingIds
+ * 2. (only if filterEntitySetIds is true) normalEntitySetIds
+ */
+internal fun buildEntityKeyIdsOfLinkingIdsSql(filterEntitySetIds: Boolean): String {
+    val maybeEntitySetIdsClause = if (filterEntitySetIds) "AND ${ENTITY_SET_ID.name} = ANY(?) " else ""
+
+    return "SELECT ${LINKING_ID.name}, array_agg(${ID.name}) AS ${ENTITY_KEY_IDS_COL.name} " +
+            "FROM ${IDS.name} " +
```
makosblade: Yeah I think your `WHERE` escaped here

src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
katherinebernstein: We used to compute the partition for linked data based on the linking id, but we recently switched to using the origin entity key id to compute the partition. In order to accurately compute the partition for the linking data, we'd need to look up the normal entity key id (origin id) for the linking ids, which is what this function is doing.

The one case where we *could* use the partition is in the case where all the normalEntitySetIds all have a singleton partition list where the partition is the same between all of them, but that seems like a rare/specific enough case that it's not worth special casing.

src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
katherinebernstein: :rip: indeed

src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
makosblade: 👍 I think we probably want it since it's part of the primary key for `IDS` table

you-leee: We can use all the partitions of the normal entity set ids, since the actual partitions are a subset of that (and it's still better than going through all partitions)

you-leee: We had a similar discussion with Matthew about how to handle `PostgresEntityDataQueryService.getEntitySetIterable` for linking = true cases and it seemed the most efficient to use all entity set partitions.
https://jira.openlattice.com/browse/LATTICE-2319

katherinebernstein: Yeah, it depends how this is being called i.e. many normal entity sets are passed here / how many partitions they belong to -- if they end up spanning pretty much all the partitions anyway the extra partitions lookup will just be extra overhead, but if not then it might speed up the query. I can add it in, but I imagine it won't be strictly an improvement on perf.

katherinebernstein: (I added it, but we may want to keep an eye on typical usage here)

PR #708, opened 2019-12-09
Bugfix/lattice 2336 organizations map

pckeyes:
Change usage of OrganizationTitles to Organizations IMap


you-leee:
That was quick :)

src/main/kotlin/com/openlattice/auditing/AuditRecordEntitySetsManager.kt
```diff
@@ -65,7 +66,7 @@ class AuditRecordEntitySetsManager(
             HazelcastMap.AUDIT_RECORD_ENTITY_SETS.name
     )
 
-    private val organizationTitles = hazelcastInstance.getMap<UUID, String>(
```
katherinebernstein: I think the map name here still need to be changed to `ORGANIATIONS`

katherinebernstein: *ORGANIZATIONS. if you change it to ORGANIATIONS then it will not compile. lol

pckeyes: WHoops

PR #710, opened 2019-12-10
Remove partition filtering logic from PLQS.getEntityKeyIdsOfLinkingIds

katherinebernstein:
I thought about it a little more, and I'm worried about the scalability of the partition filtering addition. As linking begins to run in production and we get more and more data, it seems likely that linking ids will be assigned across more and more entity sets. Over time, it seems possible that this may be called with thousands of normal entity set ids, which will basically span all partitions (so we get no query perf improvements) and be a potentially very large hazelcast read. Excluding the partition filter from the query will also be no slower than when normal entity set ids are not specified.

That said, when this is called on a specific linking entity set, the number of normal entity set ids may be relatively small, so in that case the hazelcast read wouldn't be too bad and the query improvement would be worth it. 

Anyway, leaving this PR here for now...


geekbeast:
While in theory it could get to thousand of entity sets it seems like an unlikely case. An individual is unlikely to appear in thousands of person entity sets (unless something is really funky with linking). In fact, while our jurisdictions are sparse people are likely to be concentrated in a small number of partitions corresponding to a single geography/jurisdiction.

geekbeast:
@katherinebernstein Any objections to closing this without merging based on above?

katherinebernstein:
> @katherinebernstein Any objections to closing this without merging based on above?

Nope, as long as the use case will always be filtered by normal entity sets (which I think julia made a requirement for this call). I'd imagine linked entities (without entity set filters) will not reliably be in a small set of partitions (especially as we get more and more data) since we have no way of marking different organizations within the same jurisdiction to use the same partitions, but that set will probably be largely cut down when filtering on a subset of the normal entity sets.

Anyway, I'll close this

PR #716, opened 2019-12-11
Bugfix/lattice 2341 switch to near cache

geekbeast:
Hazelcast has an eventually consistent near cache that has a built in mechanism for cluster wide invalidation. This PR switches the per node guava caches to Hazelcast Near Caches.


src/main/java/com/openlattice/authorization/Principals.java
```diff
@@ -43,49 +41,26 @@
...
+            users = hazelcastInstance.<String, SecurablePrincipal>getMap( HazelcastMap.PRINCIPALS.name() );
+            principals = hazelcastInstance.<String, NavigableSet<Principal>>getMap( HazelcastMap.PRINCIPALS.name() );
         } else {
             logger.error( "Principals security processing can only be initialized once." );
             throw new IllegalStateException( "Principals context already initialized." );
         }
     }
 
-    public static void requireOrganization( Principal principal ) {
+    public static void ensureOrganization( Principal principal ) {
```
you-leee: I don't think this is used anywhere right now, but could be from `Organization.init`

src/main/java/com/openlattice/hazelcast/HazelcastReliableTopic.kt
```diff
@@ -0,0 +1,9 @@
+package com.openlattice.hazelcast
+
+/**
+ *
+ * @author Matthew Tamayo-Rios &lt;matthew@openlattice.com&gt;
+ */
+enum class HazelcastReliableTopic {
```
you-leee: What is the purpose of this?

src/main/kotlin/com/openlattice/authorization/mapstores/ResolvedPrincipalTreesMapLoader.kt
```diff
@@ -0,0 +1,101 @@
...
+        return currentPrincipals
+    }
+
+    override fun delete(key: String) {
+        throw NotImplementedException("This is a read only map loader.")
+    }
+
+    @Inject
+    fun initPrincipalsMapstore(hazelcastInstance: HazelcastInstance) {
+        this.principals = hazelcastInstance.getMap(HazelcastMap.SECURABLE_PRINCIPALS.name)
```
makosblade: I'm a little confused as to why we have this whole class that wraps an IMap. Is it just that there's no way to configure hazelcast to configure `HazelcastMap.SECURABLE_PRINCIPALS` as a `NearCache`?

src/main/kotlin/com/openlattice/hazelcast/serializers/PrincipalTreeInvalidatedEventStreamSerializer.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.hazelcast.nio.ObjectDataInput
+import com.hazelcast.nio.ObjectDataOutput
+import com.openlattice.hazelcast.StreamSerializerTypeIds
+import com.openlattice.principals.PrincipalTreeInvalidatedEvent
+import org.apache.commons.lang3.RandomStringUtils
+import org.springframework.stereotype.Component
+
+@Component
+class PrincipalTreeInvalidatedEventStreamSerializer : TestableSelfRegisteringStreamSerializer<PrincipalTreeInvalidatedEvent> {
+    override fun generateTestValue(): PrincipalTreeInvalidatedEvent {
```
you-leee: Where is the implementation of `PrincipalTreeInvalidatedEvent`?

src/main/java/com/openlattice/authorization/Principals.java
geekbeast: Was not used.

src/main/kotlin/com/openlattice/authorization/mapstores/ResolvedPrincipalTreesMapLoader.kt
geekbeast: I think this is out of date now. This class specifically also does more and uses SecurablePrincipalManager to walk tree of all children into a sorted set of SecurablePrincipals.

src/main/kotlin/com/openlattice/hazelcast/serializers/PrincipalTreeInvalidatedEventStreamSerializer.kt
geekbeast: This was removed.

src/main/java/com/openlattice/hazelcast/HazelcastReliableTopic.kt
geekbeast: Is an enum for reliable topics. I created one and then removed it so it has no reason to exist anymore, but left it in just in case.

src/main/kotlin/com/openlattice/authorization/mapstores/ResolvedPrincipalTreesMapLoader.kt
geekbeast: i.e it leverages multiple underlying maps and caches the computation (PrincipalTrees and Principals) based on id lookup.

src/main/kotlin/com/openlattice/authorization/mapstores/PrincipalTreesMapstore.kt
```diff
@@ -125,9 +125,10 @@ class PrincipalTreesMapstore(val hds: HikariDataSource) : TestableSelfRegisterin
                     ResultSetAdapters.aclKey(it) to ResultSetAdapters.principalOfAclKey(it)
                 }
         )
-        val map: MutableMap<AclKey, AclKeySet> = mutableMapOf()
+        val map = mutableMapOf<AclKey, AclKeySet>()
         data.forEach { map.getOrPut(it.first) { AclKeySet() }.add(it.second) }
-        (map.keys - keys).forEach { map.putIfAbsent(it, AclKeySet()) }
+        ( keys - map.keys ).forEach { map[it] = AclKeySet() }
```
katherinebernstein: this seems very important :D

src/main/kotlin/com/openlattice/authorization/mapstores/ResolvedPrincipalTreesMapLoader.kt
```diff
@@ -0,0 +1,99 @@
...
+        throw NotImplementedException("This is a read only map loader.")
+    }
+
+    override fun storeAll(map: Map<String, NavigableSet<Principal>>) {
+        throw NotImplementedException("This is a read only map loader.")
+    }
+
+    override fun getMapConfig(): MapConfig {
+        return MapConfig(mapName)
+                .setMaxIdleSeconds(300)
```
dbt: The old loading cache used a 30 second eviction time. Is this ok because we're relying on eviction when underlying data changes?

geekbeast: That's why I switched to five minutes, because if someone's permission changes hazelcast should in theory eventually invalidate caches instead of need to constantly refresh as user ever 30 seconds.

geekbeast: So, exactly!

src/main/kotlin/com/openlattice/authorization/mapstores/PrincipalTreesMapstore.kt
geekbeast: I know... this bug has been here for a while.

PR #718, opened 2019-12-12
add scripts to setup new citus, kill citus and docker-compose (from c…

makosblade:
This PR:
- Adds citus' docker-compose.yml (may want to nix that, I could go either way)
- Adds an initCitus script that starts citus and runs the appropriate postgres commands
- Adds a killCitus command that will tear-down citus
- Encapsulates database setup commands into sql files to be used with said script

### DIFF
--- a/src/main/resources/initCitus.sh
+++ b/src/main/resources/initCitus.sh
@@ -14,23 +14,16 @@ then
   exit 1
 fi
 
-function runPsqlOnNode {
-  docker exec "$1" psql -U postgres -c "$2"
-}
-
-function runPsqlFileOnNodeWithOLDB {
-  docker exec "$1" psql -U postgres -d openlattice -f "$2"
-}
-
-function runPsqlFileOnNode {
-  docker exec "$1" psql -U postgres -f "$2"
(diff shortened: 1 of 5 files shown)


src/main/resources/initCitus.sh
```diff
@@ -0,0 +1,67 @@
...
+function runPsqlFileOnNode {
+  sudo docker exec $1 psql -U postgres -f $2
+}
+
+MASTER_EXTERNAL_PORT=5433 COMPOSE_PROJECT_NAME=citus docker-compose up --scale worker=2 -d
+
+sleep 5
+
+# sudo docker exec citus_master psql -U postgres -c "SELECT * FROM master_get_active_worker_nodes();"
+NODES=`runPsqlOnNode citus_master "SELECT * FROM master_get_active_worker_nodes();"  | grep worker | cut -d ' ' -f 2`
```
dbt: I pretty much always prefer $() to `` due to nesting issues.

makosblade: This is great. I've never ever gotten feedback on a bash script

PR #719, opened 2019-12-12
Bugfix/lattice 2343 update audit entity set partitions

katherinebernstein:


src/main/kotlin/com/openlattice/data/storage/partitions/PartitionManager.kt
```diff
@@ -122,15 +122,13 @@ class PartitionManager @JvmOverloads constructor(
         setEntitySetPartitions(entitySetId, allocatedPartitions)
     }
 
-    private fun computePartitions(entitySet: EntitySet, partitionCount: Int): List<Int> {
+    private fun computePartitions(entitySet: EntitySet, partitionCount: Int?): List<Int> {
```
geekbeast: Suggest zero default instead of null with + JvmOverloads

PR #729, opened 2019-12-13
LATTICE-2315 AssemblerConnectionManager churns connection pool

dbt:
cache connection pools instead of churning them


src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -120,16 +128,24 @@ class AssemblerConnectionManager(
...
+    }
+
     fun connect(dbName: String): HikariDataSource {
-        return connect(dbName, assemblerConfiguration.server.clone() as Properties, assemblerConfiguration.ssl)
+        return perDbCache.get(dbName)
     }
 
+    /* TODO: use #perDbCache or delete.
+     */
     fun connect(dbName: String, account: MaterializedViewAccount): HikariDataSource {
```
you-leee: I think this can be as deprecated, unless we want to use `AssemblyAnalyzationController` for tamatoa. @jokedurnez should we maintain that service or not? (querying aggregates, calculations... from materialized views)

makosblade: 👍 for `@Deprecated`

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -75,6 +79,10 @@ class AssemblerConnectionManager(
         metricRegistry: MetricRegistry
 ) {
 
+    private val perDbCache: LoadingCache<String, HikariDataSource> = CacheBuilder
+            .newBuilder()
+            .expireAfterAccess(1, TimeUnit.HOURS)
```
you-leee: I'm not sure, we should let it expire, since we use 1 connection per organization and as far as I know, organizations don't change a lot. As an alternative, we could remove the expiration and remove it manually, when an organization is deleted. 
In that case, we should look out for the order of organization deletion handling (`OrganizationsController.destroyOrganization`) in `Assembler` and `ExternalDatabaseManagementService`

dbt: I'm open to this approach.

For services like this I prefer to avoid tight coupling and would rather free unused resources optimistically. We've already survived allocating new connection pools every few seconds, so once an hour should be ok.

you-leee: That's true, if you think this approach is better, I'm ok with that too.

PR #730, opened 2019-12-13
Fix/fix failing tests

you-leee:


src/test/kotlin/com/openlattice/hazelcast/serializers/SortedPrincipalSetStreamSerializerTest.kt
```diff
@@ -0,0 +1,43 @@
+/*
```
makosblade: Should be able to just have `SortedPrincipalSetStreamSerializer` implement `TestableSelfRegisteringStreamSerializer`

src/test/kotlin/com/openlattice/hazelcast/serializers/SortedPrincipalSetStreamSerializerTest.kt
```diff
@@ -0,0 +1,43 @@
...
+ */
+package com.openlattice.hazelcast.serializers
+
+import com.kryptnostic.rhizome.hazelcast.serializers.AbstractStreamSerializerTest
+import com.openlattice.authorization.Principal
+import com.openlattice.mapstores.TestDataFactory
+import com.openlattice.organizations.SortedPrincipalSet
+import java.util.*
+
+class SortedPrincipalSetStreamSerializerTest
```
katherinebernstein: This test is unnecessary since `SortedPrincipalSetStreamSerializer` implements `TestableSelfRegisteringStreamSerializer`

src/main/kotlin/com/openlattice/hazelcast/serializers/RemoveMembersFromOrganizationAssemblyProcessorStreamSerializer.kt
```diff
@@ -29,15 +29,14 @@ import com.openlattice.assembler.processors.RemoveMembersFromOrganizationAssembl
...
 
 @Component
 class RemoveMembersFromOrganizationAssemblyProcessorStreamSerializer
     : SelfRegisteringStreamSerializer<RemoveMembersFromOrganizationAssemblyProcessor>,
       AssemblerConnectionManagerDependent<Void?> {
     private lateinit var acm: AssemblerConnectionManager
-    @Inject
-    private lateinit var splss: SecurablePrincipalListStreamSerializer
+
+    private val splss = SecurablePrincipalListStreamSerializer()
```
katherinebernstein: just wanted to double check that there won't be any side effects from removing the `@Inject`?

makosblade: probably fine for now but I could see us wanting a singleton for every `StreamSerializer`. 
Actually shouldn't this `StreamSerializer` have a static method to write/read and therefore not need an instance? 🤔

katherinebernstein: it should -- actually yeah I think this was meant to be a temporary/quick fix to get prod back up, but we should add static serialize/deserialize methods in SecurablePrincipalListStreamSerializer and use those instead

you-leee: It wasn't initialized with @Inject.. I didn't look really into it, bc it's fairly easy to create a new instance of it. I don't think there should be any side effect here

src/test/kotlin/com/openlattice/hazelcast/serializers/SortedPrincipalSetStreamSerializerTest.kt
you-leee: right.. 🤦‍♀ thanks!

src/main/kotlin/com/openlattice/hazelcast/serializers/RemoveMembersFromOrganizationAssemblyProcessorStreamSerializer.kt
you-leee: 👍

you-leee: Actually, that is harder than expected, since it's implementing `ListStreamSerializer`

makosblade: It's fine for now but I think `CollectionStreamSerializer` should implement `TestableSelfRegistering` and then the implementation should be passed all the way up

you-leee:
Depends on https://github.com/geekbeast/rhizome/pull/292

src/main/kotlin/com/openlattice/hazelcast/serializers/SecurablePrincipalListStreamSerializer.kt
```diff
@@ -22,15 +22,51 @@ package com.openlattice.hazelcast.serializers
...
+import kotlin.collections.ArrayList
+import kotlin.random.Random
 
 @Component
 class SecurablePrincipalListStreamSerializer
     : ListStreamSerializer<SecurablePrincipalList, SecurablePrincipal>(SecurablePrincipalList::class.java) {
 
+    companion object {
+        @JvmStatic
+        fun createSecurablePrincipalList(): SecurablePrincipalList {
```
makosblade: Any reason not to put this in `generateTestValue()` ?
Its only used for testing so I think it would be best for it not to be static/in a production class.
Alternatively, could move it to `TestDataFactory`

you-leee: I will move it there. It's used in other test classes, for example: `AddMembersToOrganizationAssemblyProcessorStreamSerializerTest`

PR #733, opened 2019-12-16
fix count of entities upserted

katherinebernstein:


src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -317,7 +317,7 @@ class PostgresEntityDataQueryService(
                     //For now we can't track how many entities were updated in a call transactionally.
                     //If we want to check how many entities were written at a specific version that is possible but
                     //expensive.
-                    updatedEntityCount += entities.size
+                    updatedEntityCount += batch.size
```
you-leee: Why not just use `entities.size` in the log?

katherinebernstein: where in the log?

katherinebernstein: oh you mean just not compute updatedEntityCount altogether? We can do that, but I assumed it was written this way in case we ever actually compute the number of entities written (based on the comment above) so I was just keeping the status quo there. But it would be a little perfier to skip tracking the sum, so I'd be open to that

you-leee: Later where we use the value of `updatedEntityCount`. I missed though, that we use it in the WriteEvent too, but basically my question is, why not just use `entities.size` instead of summing the batch sizes?

PR #734, opened 2019-12-16
don't hold pg connections open while sending persistent search emails

katherinebernstein:


src/main/kotlin/com/openlattice/search/PersistentSearchMessengerTask.kt
```diff
@@ -198,33 +198,45 @@ class PersistentSearchMessengerTask : HazelcastFixedRateTask<PersistentSearchMes
...
-                        ps.setObject(1, latestRead)
-                        ps.setObject(2, search.id)
-                        if (logger.isDebugEnabled) {
-                            logger.debug("Updating last read for $aclKey and $search")
-                        } else {
-                            logger.info("Updating last read for $aclKey and ${search.id}")
-                        }
-                        ps.addBatch()
+
+        val persistentSearchesById = PostgresIterable(Supplier<StatementHolder> {
```
you-leee: Minor thing, but can you pls use `BasePostgresIterable` and `StatementHolderSupplier`?

PR #736, opened 2019-12-20
clear out unused in Auth0SyncService

makosblade:
This PR:
- gets rid of if-statement so that users are fully sync'd
- gets rid of unused variables/calls in syncUser

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,17 +22,17 @@ import com.github.spotbugs.SpotBugsTask
  */
 
 plugins {
+    id 'checkstyle'
     id 'eclipse'
     id 'idea'
     id 'jacoco'
-    id 'checkstyle'
-    id "maven-publish"
-    id "signing"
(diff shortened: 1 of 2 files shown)


you-leee:
Btw, isn't this a fix for this? https://jira.openlattice.com/browse/LATTICE-2333

makosblade:
> Btw, isn't this a fix for this? https://jira.openlattice.com/browse/LATTICE-2333

yes!

you-leee:
We should link them together

you-leee:
If you include `LATTICE-2333` in the title, it should link them

makosblade:
Yeah there used to be some slick integration between jira and github where you could create a branch in the appropriate repositories through jira and it would title it from the ticket number/title but I don't think we have it enabled/installed/configured

you-leee:
It worked though :)

PR #737, opened 2020-01-06
fix serializer including enums

jokedurnez:


src/main/java/com/openlattice/hazelcast/serializers/PropertyTypeStreamSerializer.java
```diff
@@ -74,6 +74,7 @@ public static void serialize( ObjectDataOutput out, PropertyType object ) throws
         out.writeBoolean( object.isPii() );
         out.writeInt( object.getAnalyzer().ordinal() );
         out.writeInt( object.getPostgresIndexType().ordinal() );
+        SetStreamSerializers.fastStringSetSerialize( out, object.getEnumValues() );
```
pckeyes: EnumValues is optional, right? I think you need an optional stream serializer method here

src/main/java/com/openlattice/hazelcast/serializers/PropertyTypeStreamSerializer.java
```diff
@@ -87,7 +88,8 @@ public static PropertyType deserialize( ObjectDataInput in ) throws IOException
         Optional<Boolean> piiField = Optional.of( in.readBoolean() );
         Optional<Analyzer> analyzer = Optional.of( analyzers[ in.readInt() ] );
         Optional<IndexType> indexMethod = Optional.of( INDEX_TYPES[ in.readInt() ] );
-        return new PropertyType( id, type, title, description, schemas, datatype, piiField, analyzer, indexMethod );
+        Optional<Set<String>> enumValues = Optional.of( SetStreamSerializers.orderedFastStringSetDeserialize( in ));
```
pckeyes: And I think you need to use the OptionalStreamSerializers methods here too

src/main/java/com/openlattice/hazelcast/serializers/PropertyTypeStreamSerializer.java
you-leee: At this point, enumValues are not optional, but a `LinkedHashSet`, so I think this is the right serializer to use.

PR #738, opened 2020-01-08
Bugfix/lattice 2358 user vs auth0basicuser

you-leee:
Fixes ClassCastException for getUser in UserDirectoryService + some cleanup


src/main/kotlin/com/openlattice/directory/UserDirectoryService.kt
```diff
@@ -0,0 +1,81 @@
...
+    open fun getAllUsers(): Map<String, User> {
+        return ImmutableMap.copyOf(users)
+    }
+
+    fun getUser(userId: String): User {
+        return users.getValue(userId)
+    }
+
+    //TODO: Switch over to a Hazelcast map to relieve pressure from Auth0
+    fun searchAllUsers(searchQuery: String): Map<String, Auth0UserBasic> {
```
dbt: can we change this to use the new User class as well?

you-leee: Probably we could switch over to an aggregator, if we don't search on fields that are not present in User (`roles` and `com/openlattice/organization`) plus it would require to collect all the different type of searches the apps use with `PrincipalDirectoryController.searchAllUsers` via https://auth0.com/docs/users/search/v3

you-leee: I will ask around

you-leee: So, we decided to make a separate ticket for this, since it's not high prio

you-leee: https://jira.openlattice.com/browse/LATTICE-2361

PR #739, opened 2020-01-08
LATTICE-2360 use let, so it won't close HikariDataSource, handle null…

you-leee:
… AssemblerConnectionManager the right way

### DIFF
--- a/src/main/kotlin/com/openlattice/assembler/processors/AddMembersToOrganizationAssemblyProcessor.kt
+++ b/src/main/kotlin/com/openlattice/assembler/processors/AddMembersToOrganizationAssemblyProcessor.kt
@@ -26,13 +26,12 @@ import com.hazelcast.core.ReadOnly
 import com.hazelcast.spi.ExecutionService
 import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor
 import com.openlattice.assembler.AssemblerConnectionManager
+import com.openlattice.assembler.AssemblerConnectionManagerDependent
 import com.openlattice.assembler.OrganizationAssembly
 import com.openlattice.assembler.PostgresDatabases
 import com.openlattice.authorization.SecurablePrincipal
-import com.openlattice.assembler.AssemblerConnectionManagerDependent
 import com.openlattice.edm.EntitySet
 import com.openlattice.edm.type.PropertyType
-import java.lang.IllegalStateException
 import java.util.*
 
 data class AddMembersToOrganizationAssemblyProcessor(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/openlattice/assembler/processors/AddMembersToOrganizationAssemblyProcessor.kt
```diff
@@ -43,21 +42,18 @@ data class AddMembersToOrganizationAssemblyProcessor(
...
-            throw IllegalStateException("Encountered null assembly while trying to add new principals " +
-                    "${authorizedPropertyTypesOfEntitySetsByPrincipal.keys} to organization $organizationId.")
-        } else {
-            val dbName = PostgresDatabases.buildOrganizationDatabaseName(organizationId)
-            acm.connect(dbName)
-                    .use { dataSource ->
-                        acm.addMembersToOrganization(dbName, dataSource, authorizedPropertyTypesOfEntitySetsByPrincipal)
-                    }
-        }
+        assembly ?: throw IllegalStateException("Encountered null assembly while trying to add new principals " +
```
geekbeast: consider check( assembly != null ) { "..." }

src/main/kotlin/com/openlattice/assembler/processors/AddMembersToOrganizationAssemblyProcessor.kt
```diff
@@ -43,21 +42,18 @@ data class AddMembersToOrganizationAssemblyProcessor(
...
+                "${authorizedPropertyTypesOfEntitySetsByPrincipal.keys} to organization $organizationId.")
+
+        val dbName = PostgresDatabases.buildOrganizationDatabaseName(organizationId)
+        acm?.connect(dbName)?.let { dataSource ->
```
geekbeast: What's the purpose of allow acm to be null?

you-leee: In case, the `init` function is not called on the processor I think it's more appropriate to throw an exception then to ignore it (and get e nullpointer later)

src/main/kotlin/com/openlattice/assembler/processors/AddMembersToOrganizationAssemblyProcessor.kt
you-leee: why is that better? or just for re-use purpose?

geekbeast: Idiomatic style for checking state using kotlin std lib.

src/main/kotlin/com/openlattice/assembler/processors/AddMembersToOrganizationAssemblyProcessor.kt
geekbeast: Idiomatic way of doing this in kotlin is to leverage the compiler instead of using nullable type. ```check( ::acm.isInitialized ) { "ACM was not initialized before processing." }```

geekbeast: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/is-initialized.html

you-leee: Oh, cool! Will do!

PR #741, opened 2020-01-10
LATTICE-2364 create separate pod for hazelcast queues

you-leee:
Needs https://github.com/geekbeast/rhizome/pull/295

### DIFF
--- a/src/main/kotlin/com/openlattice/hazelcast/pods/HazelcastQueuePod.kt
+++ b/src/main/kotlin/com/openlattice/hazelcast/pods/HazelcastQueuePod.kt
@@ -29,10 +29,15 @@ import org.springframework.context.annotation.Configuration
 @Configuration
 class HazelcastQueuePod {
 
+    companion object {
+        @JvmField
+        val maxQueueSize = 10000
+    }
+
     @Bean
     fun defaultQueueConfigurer(): QueueConfigurer {
         return QueueConfigurer(BaseHazelcastInstanceConfigurationPod.defaultQueueName) { config ->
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/openlattice/hazelcast/pods/HazelcastQueuePod.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+class HazelcastQueuePod {
+
+    @Bean
+    fun defaultQueueConfigurer(): QueueConfigurer {
+        return QueueConfigurer(BaseHazelcastInstanceConfigurationPod.defaultQueueName) { config ->
+            config.setMaxSize(10000).emptyQueueTtl = 60
```
katherinebernstein: why are we keeping this as the default?

makosblade: would prefer extracting to a constant at least 👍

you-leee: I'm not sure about the values (I kept them as they were), especially why we have a `ttl`, but in case there is no config defined for a queue, I believe this default will apply.

you-leee: After speaking with @geekbeast , the ttl will be deleted

PR #744, opened 2020-01-14
Typesafe Queues

makosblade:
This PR:
- removes `SIGNAL` queue as it is unused in the codebase
- Adds a typesafe queue wrapper


src/main/kotlin/com/openlattice/hazelcast/HazelcastQueue.kt
```diff
@@ -0,0 +1,64 @@
...
+    fun getQueue(hazelcast: HazelcastInstance): IQueue<T> {
+        this.checker.check()
+        return hazelcast.getQueue(name)
+    }
+
+    companion object {
+        private val instanceChecker = UniqueInstanceManager(HazelcastQueue::class.java)
+
+        // When adding new entries to this list, please make sure to keep it sorted and keep the name in sync
+        @JvmField val AUDITING = HazelcastQueue<AuditableEvent>("AUDITING")
```
dbt: `AUDITING` and `EMAIL_SPOOL` don't have `QueueConfigurer`s, is that bad?

src/main/kotlin/com/openlattice/hazelcast/HazelcastQueue.kt
```diff
@@ -0,0 +1,64 @@
...
+        @JvmField val BACKGROUND_INDEXING  = HazelcastQueue<EntitySet>("BACKGROUND_INDEXING")
```
dbt: Also currently unused?

makosblade: Yep, this is used in a future PR

src/main/kotlin/com/openlattice/hazelcast/HazelcastQueue.kt
makosblade: 🤔 It's not ~great~ since they won't have a max size set I think. I can add that

you-leee: They will have, because of the default we set up: `10_000`

dbt: They will have a default size but also a default TTL of 60s.

PR #749, opened 2020-01-17
Feature/local auth

geekbeast:
Extracted some interfaces so that component aren't interacting directly with auth0, allowing us to swap in local only implementations.


src/main/kotlin/com/openlattice/users/LocalUserListingService.kt
```diff
@@ -0,0 +1,70 @@
...
+                        .withClaim("user_id", user.id)
+                        .withClaim("email", user.email)
+                        .withClaim("email_verified", user.isEmailVerified)
+                        .withIssuer(aac.issuer)
+                        .withAudience(aac.audience)
+                        //TODO: Hardcoded for now, but should actually parse algorithm string
+                        .sign(Algorithm.HMAC256(aac.secret))
+
+
+                logger.info("user.id -> $jwt")
```
katherinebernstein: should be ${user.id} i think

src/main/kotlin/com/openlattice/users/LocalUserListingService.kt
```diff
@@ -0,0 +1,70 @@
...
+    private val users = auth0Configuration.users.associateBy { it.id }
+
+    init {
+        logger.info("************************* BEGIN JWT TOKENS *************************")
+        auth0Configuration.clients.forEach { aac ->
+
+            auth0Configuration.users.forEach { user ->
+                val jwt = JWT.create()
+                        .withSubject(user.id)
+                        .withClaim("user_id", user.id)
```
you-leee: Could you put these property names in constants? We also use it in `PrincipalDirectoryController`

src/main/kotlin/com/openlattice/users/Auth0UserListingService.kt
```diff
@@ -0,0 +1,90 @@
...
+            getUsersPage(managementApi, page++, DEFAULT_PAGE_SIZE).items ?: listOf()
+        } catch (ex: Exception) {
+            logger.error("Retrofit called failed during auth0 sync task.", ex)
+            listOf()
+        }
+    }
+
+    //TODO: Move this to a utils class
+    private fun getUsersPage(managementApi: ManagementAPI, page: Int, pageSize: Int): UsersPage {
+        require(pageSize < MAX_PAGE_SIZE) { "Requested page size of $pageSize exceeds max page size of $MAX_PAGE_SIZE " }
```
you-leee: That should be probably `<=`

src/main/kotlin/com/openlattice/users/Auth0UserListingService.kt
```diff
@@ -0,0 +1,90 @@
...
+        return managementApi.users().list(
+                UserFilter()
+                        .withSearchEngine("v3")
+                        .withFields("user_id,email,nickname,app_metadata,identities", true)
```
you-leee: Could you put these property names in constants pls?

src/main/kotlin/com/openlattice/directory/UserDirectoryService.kt
```diff
@@ -22,40 +22,78 @@
...
+interface UserDirectoryService {
+    @Timed
+    fun getAllUsers(): Map<String, User>
+
+    @Timed
+    fun getUser(userId: String): User
+
+    //TODO: Switch over to a Hazelcast map to relieve pressure from Auth0
+    @Timed
+    fun searchAllUsers(searchQuery: String): Map<String, Auth0UserBasic>
```
you-leee: I think we could switch over and use User here as well. If I remember right, it has the "missing" properties named `roles` and `com/openlattice/organization` (also, is this intentional or a typo? 🙃  ) in `app_metadata`

you-leee: https://jira.openlattice.com/browse/LATTICE-2361

you-leee: Oh, just noticed, that the non-local one still has the old implementation...nvm then

src/main/kotlin/com/openlattice/directory/UserDirectoryService.kt
```diff
@@ -22,40 +22,78 @@
...
+
+    override fun getUser(userId: String): User {
+        return users.getValue(userId)
+    }
+
+    override fun searchAllUsers(searchQuery: String): Map<String, Auth0UserBasic> {
+        return users.values.filter { user ->
+            (listOf(user.email, user.name, user.nickname, user.givenName, user.familyName, user.username) +
+                    user.identities.map { it.userId } + user.identities.map { it.connection })
+                    .any { it.contains(searchQuery) }
```
you-leee: I don't think this will work, bc the query is pre-formatted: 
`String exactEmailSearchQuery = "email.raw:\"" + emailSearchQuery + "\"";`
See in: `PrincipalDirectoryController.searchAllUsersByEmail`

src/main/kotlin/com/openlattice/users/LocalUserListingService.kt
```diff
@@ -0,0 +1,70 @@
...
+                        .withClaim("email", user.email)
+                        .withClaim("email_verified", user.isEmailVerified)
+                        .withIssuer(aac.issuer)
+                        .withAudience(aac.audience)
+                        //TODO: Hardcoded for now, but should actually parse algorithm string
+                        .sign(Algorithm.HMAC256(aac.secret))
```
makosblade: Just pinging on the hardcode

src/main/kotlin/com/openlattice/directory/UserDirectoryService.kt
```diff
@@ -22,40 +22,78 @@
...
 import com.openlattice.hazelcast.HazelcastMap
 import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
 
 
-open class UserDirectoryService(auth0TokenProvider: Auth0TokenProvider, hazelcastInstance: HazelcastInstance) {
-    companion object {
-        private val logger = LoggerFactory.getLogger(UserDirectoryService::class.java)
-        const val DEFAULT_PAGE_SIZE = 100
+interface UserDirectoryService {
```
makosblade: Separate file?

src/main/kotlin/com/openlattice/directory/UserDirectoryService.kt
geekbeast: I just inverted it. Search doesn't need to be good.

PR #750, opened 2020-01-18
LATTICE-2375 delete only edges if it's an association es, fix some warnings

you-leee:
Basically remove the `!` when checking if it's an association entity set or not


makosblade:
lgtm, @katherinebernstein knows this code a bit better IIRC

you-leee:
For some context:
This pr updates, that when we delete entities from an association entity set besides deleting the entities themselves, we only delete edges that are connected to it. While when deleting from a non-association entity set, we also delete the association entities connected to those entities via edges.
For example consider these entities, where involvedin is an association type, person and arrest are non-association types:
```
person1 - involvedin1 -arrest1
person2 - involvedin2 -arrest1
person3 - involvedin3 -arrest1
```
If we delete arrest1, we also delete involvedin1,2,3 + the edges connecting them, but if we delete involvedin1, we only delete involvedin1 and the edges, but not arrest1 nor person1

PR #754, opened 2020-01-21
change entitydatastore to use map instead of multimap

dbt:
followup to #753


you-leee:
Actually I already did these changes in task 1621 with including the TODO
https://github.com/openlattice/conductor-client/compare/develop...feature/LATTICE-1621-linking-id-obfuscation#diff-ec931d61da50a1be2b2ef11ee79de88b

you-leee:
That PR is getting too big :)

PR #757, opened 2020-01-22
Remove unused from exDBMService

makosblade:
This PR:
- removes some .getValue on IMap b/c it does extra operations
- Update EntitySetService to use more EntryProcessors instead of .getAll on EntitySets

### DIFF
--- a/src/main/kotlin/com/openlattice/datastore/services/EntitySetService.kt
+++ b/src/main/kotlin/com/openlattice/datastore/services/EntitySetService.kt
@@ -21,8 +21,8 @@
 package com.openlattice.datastore.services
 
 import com.codahale.metrics.annotation.Timed
-import com.google.common.base.Preconditions.checkState
 import com.google.common.base.Preconditions.checkArgument
+import com.google.common.base.Preconditions.checkState
 import com.google.common.collect.Sets
 import com.google.common.eventbus.EventBus
 import com.hazelcast.core.HazelcastInstance
@@ -99,7 +99,7 @@ open class EntitySetService(
 
 
     override fun createEntitySet(principal: Principal, entitySet: EntitySet) {
-        val entityType = entityTypes.getValue(entitySet.entityTypeId)
+        val entityType = Util.getSafely(entityTypes, entitySet.entityTypeId)
         ensureValidEntitySet(entitySet)
 
         if (entitySet.partitions.isEmpty()) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/openlattice/datastore/services/EntitySetService.kt
```diff
@@ -343,12 +340,12 @@ open class EntitySetService(
     override fun getAllEntitySetPropertyMetadataForIds(
             entitySetIds: Set<UUID>
     ): Map<UUID, Map<UUID, EntitySetPropertyMetadata>> {
-        val entitySetsById = entitySets.getAll(entitySetIds)
-        val entityTypesById = entityTypes.getAll(entitySetsById.values.map(EntitySet::getEntityTypeId).toSet())
+        val entitySetTypesById = entitySets.executeOnKeys(entitySetIds, GetEntityTypeFromEntitySetEntryProcessor()) as Map<UUID, UUID>
```
katherinebernstein: I kinda prefer `entityTypesByEntitySetId` to clarify what id refers to

src/main/kotlin/com/openlattice/datastore/services/EntitySetService.kt
```diff
@@ -99,7 +99,7 @@ open class EntitySetService(
 
 
     override fun createEntitySet(principal: Principal, entitySet: EntitySet) {
-        val entityType = entityTypes.getValue(entitySet.entityTypeId)
+        val entityType = Util.getSafely(entityTypes, entitySet.entityTypeId)
```
geekbeast: Why the switch? The left over Util.getSafely in this class is from the original java implementation. Kotlin .getValue is typesafe without needing utility method.

makosblade: Because the implementation of `.getValue()` on an `IMap` also calls `.containsKey`
```
internal inline fun <K, V> Map<K, V>.getOrElseNullable(key: K, defaultValue: () -> V): V {
    val value = get(key)
    if (value == null && !containsKey(key)) {
        return defaultValue()
    } else {
        @Suppress("UNCHECKED_CAST")
        return value as V
    }
}
```

makosblade: As a follow-up, I used `getSafely` specifically because there was a pattern of using it in `EntitySetService` I'm happy to deprecate that method and just use `.get` if you think that's the way to go moving forward

PR #761, opened 2020-01-28
LATTICE-2267 delete also linking entity properties from data table wh…

you-leee:
…en deleting entity properties

### DIFF
--- a/src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
+++ b/src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
@@ -571,54 +571,43 @@ internal val deletePropertyInEntitySet = "DELETE FROM ${DATA.name} WHERE ${ENTIT
 internal val deleteEntitySetEntityKeys = "DELETE FROM ${IDS.name} WHERE ${ENTITY_SET_ID.name} = ? "
 
 /**
- * Preparable SQL deletes all property values of entities in a given entity set in [DATA]
+ * Preparable SQL that deletes selected property values of entities and their linking entities in a given entity set
+ * in [DATA] table.
  *
  * The following bind order is expected:
  *
  * 1. entity set id
- * 2. entity key ids
- *    IF LINKING    checks against ORIGIN_ID
- *    ELSE          checks against ID column
- * 3. partition
- * 4. property type ids
+ * 2. entity key ids (non-linking entities, ID column)
+ * 3. entity key ids (linking entities, ORIGIN_ID column)
+ * 4. partition
+ * 5. property type ids
  */
-internal fun deletePropertiesOfEntitiesInEntitySet(linking: Boolean = false): String {
-    val idsColumn = if (linking) {
-        ORIGIN_ID.name
-    } else {
-        ID_VALUE.name
-    }
-
-    // @formatter:off
-    return "DELETE FROM ${DATA.name} " +
-            "WHERE   ${ENTITY_SET_ID.name} = ? AND " +
-                    "$idsColumn = ANY( ? ) AND " +
-                    "${PARTITION.name} = ? AND " +
-                    "${PROPERTY_TYPE_ID.name} = ANY( ? ) "
+// @formatter:off
+internal val deletePropertiesOfEntitiesInEntitySet =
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -585,11 +577,26 @@ internal val deleteEntitySetEntityKeys = "DELETE FROM ${IDS.name} WHERE ${ENTITY
...
  * 2. entity key ids
+ *    IF LINKING    checks against ORIGIN_ID
+ *    ELSE          checks against ID column
  * 3. partition
  * 4. property type ids
  */
-internal val deletePropertiesOfEntitiesInEntitySet = "DELETE FROM ${DATA.name} " +
-        "WHERE ${ENTITY_SET_ID.name} = ? AND ${ID_VALUE.name} = ANY(?) AND ${PARTITION.name} = ? AND ${PROPERTY_TYPE_ID.name} = ANY(?) "
+internal fun deletePropertiesOfEntitiesInEntitySet(linking: Boolean = false): String {
+    val idsColumn = if (linking) {
```
pckeyes: Is it not the case that when a non-linking entity has properties deleted that the linking entity should also have those properties deleted?

you-leee: Yes, you are right, but this is done in two different transactions, one called with `linking = false` and one with `linking = true`, so we check the non-linking entity key ids against the `ID` column and then against `ORIGIN_ID` column.  If you look at the code here, you will see that we execute this query twice: https://github.com/openlattice/conductor-client/pull/761/files#diff-9eff35dfabde14ebc448dc06ae0893f2R681

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -671,24 +671,36 @@ class PostgresEntityDataQueryService(
...
-
             val idsArr = PostgresArrays.createUuidArray(connection, entities)
 
             // Delete entity properties from data table
-            val ps = connection.prepareStatement(deletePropertiesOfEntitiesInEntitySet)
-            ps.setObject(1, entitySetId)
-            ps.setArray(2, idsArr)
-            ps.setInt(3, partition)
-            ps.setArray(4, propertyTypesArr)
+            val deletedPropertiesCount = connection
```
katherinebernstein: Is there a reason not to combine this into one query and only do one update instead of two? i.e. 
```
DELETE FROM data WHERE entity_set_id = ? AND ( id = ANY( ? ) OR origin_id = ANY( ? ) ) AND ...
```

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -725,20 +737,31 @@ class PostgresEntityDataQueryService(
...
 
-
             val idsArr = PostgresArrays.createUuidArray(connection, entities)
 
             // Delete entity properties from data table
-            val ps = connection.prepareStatement(deleteEntitiesInEntitySet)
-            ps.setObject(1, entitySetId)
-            ps.setArray(2, idsArr)
-            ps.setInt(3, partition)
+            val deletedCount = connection.prepareStatement(deleteEntitiesInEntitySet())
```
katherinebernstein: same question above re combining these updates

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
you-leee: I guess, there isn't (maybe except for logging purposes). 
I just followed the design of the tombstone cases: `updateVersionsForPropertyTypesInEntitiesInEntitySet`, `updateVersionsForPropertyValuesInEntitiesInEntitySet`
Will change it to be a single transaction.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
pckeyes: Awesome :)

PR #766, opened 2020-02-06
Bugfix/resurrect heracles texting

katherinebernstein:
Note: deploying these changes will require adding a `shortCode` field to `twilio.yaml` in s3


src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -489,6 +490,10 @@ class HazelcastOrganizationService(
 
     fun setSmsEntitySetInformation(entitySetInformationList: Collection<SmsEntitySetInformation>) {
         phoneNumbers.setPhoneNumber(entitySetInformationList)
+
+        entitySetInformationList.groupBy { it.organizationId }.forEach { (organizationId, entitySetInfoList) ->
+            organizations.executeOnKey(organizationId, UpdateOrganizationSmsEntitySetInformationEntryProcessor(entitySetInfoList))
```
makosblade: Depending on the size of `entitySetInformationList`, you may want to replace with `.map {  .submitToKey() }.forEach { it.get() }` so that this doesn't wait for the result of each execution before sending the next out over hazelcast b/c `.executeOnKey` is synchronous

src/test/kotlin/com/openlattice/hazelcast/serializers/UpdateOrganizationSmsEntitySetInformationEntryProcessorStreamSerializerTest.kt
```diff
@@ -0,0 +1,20 @@
+package com.openlattice.hazelcast.serializers
+
+import com.kryptnostic.rhizome.hazelcast.serializers.AbstractStreamSerializerTest
+import com.openlattice.assembler.AssemblerConnectionManager
+import com.openlattice.assembler.processors.UpdateMaterializedEntitySetProcessor
+import com.openlattice.mapstores.TestDataFactory
+import com.openlattice.organizations.processors.UpdateOrganizationSmsEntitySetInformationEntryProcessor
+import org.mockito.Mockito
+
+class UpdateOrganizationSmsEntitySetInformationEntryProcessorStreamSerializerTest  : AbstractStreamSerializerTest<UpdateOrganizationSmsEntitySetInformationEntryProcessorStreamSerializer,
```
makosblade: 😬 use `TestableSelfRegisteringStreamSerializer` instead please

src/main/java/com/openlattice/organizations/processors/UpdateOrganizationSmsEntitySetInformationEntryProcessor.kt
```diff
@@ -0,0 +1,40 @@
...
+            organization.smsEntitySetInfo.clear()
+            organization.smsEntitySetInfo.addAll(smsEntitySetInformation)
+            entry.setValue(organization)
+        } else {
+            logger.warn("Organization not found when trying to update value.")
+        }
+        return null
+    }
+
+    override fun equals(other: Any?): Boolean {
```
you-leee: I think, this can be avoided by making it a data class

src/main/kotlin/com/openlattice/notifications/sms/SubscriptionNotification.kt
```diff
@@ -0,0 +1,8 @@
+package com.openlattice.notifications.sms
+
+import java.util.*
```
you-leee: I think this import can be removed

src/main/kotlin/com/openlattice/hazelcast/HazelcastQueue.kt
```diff
@@ -53,15 +53,27 @@ class HazelcastQueue<T> internal constructor(val name: String) : TypedQueueIdent
...
-        @JvmField val BACKGROUND_INDEXING  = HazelcastQueue<EntitySet>("BACKGROUND_INDEXING")
-        @JvmField val EMAIL_SPOOL  = HazelcastQueue<RenderableEmailRequest>("EMAIL_SPOOL")
-        @JvmField val ID_GENERATION  = HazelcastQueue<UUID>("ID_GENERATION")
-        @JvmField val INDEXING = HazelcastQueue<UUID>("INDEXING")
-        @JvmField val QUEUED_INTEGRATION_JOBS = HazelcastQueue<QueuedIntegrationJob>("QUEUED_INTEGRATION_JOBS")
-        @JvmField val LINKING_CANDIDATES = HazelcastQueue<EntityDataKey>("LINKING_CANDIDATES")
-        @JvmField val LINKING_INDEXING = HazelcastQueue<Triple<List<Array<UUID>>, UUID, OffsetDateTime>>("LINKING_INDEXING")
-        @JvmField val LINKING_UNINDEXING = HazelcastQueue<Triple<List<Array<UUID>>, UUID, OffsetDateTime>>("LINKING_UNINDEXING")
-        @JvmField val TWILIO = HazelcastQueue<MessageRequest>("TWILIO")
+        @JvmField
```
dbt: I know it's a bit non-standard but the @JvmField annotation on the same line makes it a lot easier to read IMHO. I'd like to keep it. There was previous discussion on the HazelcastMap PR.

katherinebernstein: I agree, but the autoformatter converts it to this, so I feel like we should either update the autoformatter or leave it this way. I'm happy to change it back by hand but I feel like this will be an ongoing pain

makosblade: Vote for update autoformatter

katherinebernstein: ticket = made
https://jira.openlattice.com/browse/LATTICE-2394

katherinebernstein: Alright I manually reset it and disabled the autoformatter.

In general, I don't love disabling the autoformatter for specific cases since we'll lose *all* autoformatting which leaves room for other bad style patterns without manual upkeep and I tend to prefer consistency anyway, although I think it's fine in this case since HazelcastQueue is pretty simple and shouldn't be updated that often, and I left a TODO to re-enable once the autoformatter is updated.

PR #767, opened 2020-02-06
get rid of listMultimap from dgm api

makosblade:


you-leee:
Associations will be deserialized as  a map and not multimap in this PR already: https://github.com/openlattice/api/pull/409

src/main/kotlin/com/openlattice/data/DataGraphService.kt
```diff
@@ -320,15 +319,13 @@ open class DataGraphService(
     }
 
     override fun createAssociations(
-            associations: ListMultimap<UUID, DataEdge>,
+            associations: Map<UUID, List<DataEdge>>,
             authorizedPropertiesByEntitySetId: Map<UUID, Map<UUID, PropertyType>>
     ): Map<UUID, CreateAssociationEvent> {
 
-        val associationCreateEvents: MutableMap<UUID, CreateAssociationEvent> = mutableMapOf()
+        val associationCreateEvents: MutableMap<UUID, CreateAssociationEvent> = Maps.newLinkedHashMapWithExpectedSize(associations.size)
```
dbt: Rather than making a mutable map, could use the kotlin iterator apis:

```
return assocations.map {
...
entitySetId to CreateAssociationEvent(ids, entityWrite, edgeWrite)
}.toMap()
```

PR #768, opened 2020-02-07
Feature/lattice 2312 audit for external dbs

pckeyes:
-Adds method to return map of fqns to ids
-Service for updating tables/columns
-Enables auditing for external db objects
-Removes organization id from table/column acl keys
-Includes deleting table/column from postgres to delete apis


src/main/kotlin/com/openlattice/hazelcast/processors/organizations/UpdateOrganizationExternalDatabaseColumnEntryProcessor.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor
+import com.openlattice.edm.requests.MetadataUpdate
+import com.openlattice.organization.OrganizationExternalDatabaseColumn
+import java.util.*
+
+data class UpdateOrganizationExternalDatabaseColumnEntryProcessor(val update: MetadataUpdate) :
+        AbstractRhizomeEntryProcessor<UUID, OrganizationExternalDatabaseColumn, OrganizationExternalDatabaseColumn?>() {
+
+    override fun process(entry: MutableMap.MutableEntry<UUID, OrganizationExternalDatabaseColumn>): OrganizationExternalDatabaseColumn? {
```
you-leee: I think the return value shouldn't be nullable if the entry value is not nullable

src/main/kotlin/com/openlattice/hazelcast/processors/organizations/UpdateOrganizationExternalDatabaseTableEntryProcessor.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor
+import com.openlattice.edm.requests.MetadataUpdate
+import com.openlattice.organization.OrganizationExternalDatabaseTable
+import java.util.*
+
+data class UpdateOrganizationExternalDatabaseTableEntryProcessor(val update: MetadataUpdate):
+        AbstractRhizomeEntryProcessor<UUID, OrganizationExternalDatabaseTable, OrganizationExternalDatabaseTable?>() {
+
+    override fun process(entry: MutableMap.MutableEntry<UUID, OrganizationExternalDatabaseTable>): OrganizationExternalDatabaseTable? {
```
you-leee: I think here also, the return value shouldn't be nullable if the entry value is not nullable

src/main/kotlin/com/openlattice/hazelcast/serializers/organizations/UpdateOrganizationExternalDatabaseColumnEntryProcessorStreamSerializer.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    override fun getTypeId(): Int {
+        return StreamSerializerTypeIds.UPDATE_ORG_EXTERNAL_DB_COLUMN_PROCESSOR.ordinal
+    }
+
+    override fun getClazz(): Class<out UpdateOrganizationExternalDatabaseColumnEntryProcessor> {
+        return UpdateOrganizationExternalDatabaseColumnEntryProcessor::class.java
+    }
+
+    override fun read(`in`: ObjectDataInput?): UpdateOrganizationExternalDatabaseColumnEntryProcessor {
```
you-leee: I don't think `ObjectDataInput` should be nullable

src/main/kotlin/com/openlattice/hazelcast/serializers/organizations/UpdateOrganizationExternalDatabaseColumnEntryProcessorStreamSerializer.kt
```diff
@@ -0,0 +1,35 @@
...
+        return UpdateOrganizationExternalDatabaseColumnEntryProcessor(MetadataUpdateStreamSerializer.deserialize(`in`))
+    }
+
+    override fun write(out: ObjectDataOutput?, `object`: UpdateOrganizationExternalDatabaseColumnEntryProcessor) {
```
you-leee: I don't think `ObjectDataOutput` should be nullable

src/main/kotlin/com/openlattice/hazelcast/serializers/organizations/UpdateOrganizationExternalDatabaseTableEntryProcessorStreamSerializer.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    override fun getTypeId(): Int {
+        return StreamSerializerTypeIds.UPDATE_ORG_EXTERNAL_DB_TABLE_PROCESSOR.ordinal
+    }
+
+    override fun getClazz(): Class<out UpdateOrganizationExternalDatabaseTableEntryProcessor> {
+        return UpdateOrganizationExternalDatabaseTableEntryProcessor::class.java
+    }
+
+    override fun read(`in`: ObjectDataInput?): UpdateOrganizationExternalDatabaseTableEntryProcessor {
```
you-leee: I don't think `ObjectDataInput` should be nullable

src/main/kotlin/com/openlattice/hazelcast/serializers/organizations/UpdateOrganizationExternalDatabaseTableEntryProcessorStreamSerializer.kt
```diff
@@ -0,0 +1,35 @@
...
+        return UpdateOrganizationExternalDatabaseTableEntryProcessor(MetadataUpdateStreamSerializer.deserialize(`in`))
+    }
+
+    override fun write(out: ObjectDataOutput?, `object`: UpdateOrganizationExternalDatabaseTableEntryProcessor) {
```
you-leee: I don't think `ObjectDataOutput` should be nullable

src/main/resources/auditing.yaml
```diff
@@ -11,4 +11,4 @@ fqns:
   PRINCIPAL: "ol.userid"
   TIMESTAMP: "general.datetime"
   OPERATION_ID: "ol.auditeventid"
-enabled: false
+enabled: true
```
you-leee: You can also just delete this property, bc it has a `true` default value

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -84,13 +87,13 @@ class ExternalDatabaseManagementService(
...
 
-        val columnAclKey = AclKey(orgId, column.tableId, column.id)
+        val columnAclKey = AclKey(column.tableId, column.id)
         authorizationManager.setSecurableObjectType(columnAclKey, SecurableObjectType.OrganizationExternalDatabaseColumn)
 
         return column.id
     }
 
-    fun createNewColumnObjects(dbName: String, tableName: String, tableId: UUID, orgId: UUID, columnName: Optional<String>): BasePostgresIterable<OrganizationExternalDatabaseColumn> {
+    fun getColumnMetadata(dbName: String, tableName: String, tableId: UUID, orgId: UUID, columnName: Optional<String>): BasePostgresIterable<OrganizationExternalDatabaseColumn> {
```
you-leee: Since this is only used in indexer, I think it should be moved there unless we want to use it elsewhere too.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -376,10 +505,19 @@ class ExternalDatabaseManagementService(
             if (privileges.contains(PostgresPrivileges.INSERT) || privileges.contains(PostgresPrivileges.UPDATE))
                 permissions.add(Permission.WRITE)
             aces.executeOnKey(aceKey, PermissionMerger(permissions, objectType, OffsetDateTime.MAX))
+            return@map Acl(aclKeyUUIDs, setOf(Ace(principal, permissions, Optional.empty())))
         }
     }
 
     /*PRIVATE FUNCTIONS*/
+    private fun getNameFromFqnString(fqnString: String): String {
+        return fqnString.split(".")[1]
```
you-leee: You could also create a new `FullQualifiedName` here and then return its name. That way, it checks if it's not a malformed fqn.
`return FullQualifiedName(fqnString).name`

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -167,7 +170,7 @@ class ExternalDatabaseManagementService(
         return organizationExternalDatabaseColumns.getValue(columnId)
     }
 
-    fun getColumnNamesByTable(orgId: UUID, dbName: String): Map<String, Set<String>> {
+    fun getColumnNamesByTable(dbName: String): Map<String, Set<String>> {
         val columnNamesByTableName = HashMap<String, MutableSet<String>>()
```
you-leee: I think you could create the map with an expected size of `organizationExternalDatabaseTables.size`

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -179,37 +182,127 @@ class ExternalDatabaseManagementService(
...
+            assemblerConnectionManager.connect(dbName).let {
+                it.connection.createStatement().use { stmt ->
+                    stmt.execute("ALTER TABLE $tableName $dropColumnsSql")
+                }
+            }
+
+
+
+            //delete securable objects
             columnIds.forEach {
```
you-leee: Shouldn't this code block be just a call to `deleteOrganizationExternalDatabaseColumnObjects` ?

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -271,12 +368,30 @@ class ExternalDatabaseManagementService(
...
+        //split acls between table and column objects
+        val aclsByType = acls.partition { it.aclKey.size == 1 }
+
+        val tableAclsByOrg = aclsByType.first.groupBy {
+            organizationExternalDatabaseTables.getValue(it.aclKey[0]).organizationId
+        }
+        val columnAclsByOrg = aclsByType.second.groupBy {
+            organizationExternalDatabaseColumns.getValue(it.aclKey[1]).organizationId
+        }
+        val orgIds = tableAclsByOrg.keys.union(columnAclsByOrg.keys)
```
you-leee: You could just do: `val orgIds = tableAclsByOrg.keys + columnAclsByOrg.keys`

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -271,12 +368,30 @@ class ExternalDatabaseManagementService(
...
+
+        orgIds.forEach { orgId ->
+            val orgAcls = mutableListOf<Acl>()
+            if (tableAclsByOrg.isNotEmpty()) tableAclsByOrg.getValue(orgId).toMutableList()
```
you-leee: Isn't this supposed to add those values to `orgAcls`?

you-leee: Also, you can just do `val orgAcls = tableAclsByOrg[orgId]?.toMutableList()` and then `columnAclsByOrg[orgId]?.let { orgAcls.addAll(it) }`

src/main/resources/auditing.yaml
pckeyes: Oh I forgot to change that back to false. I think dave added that field recently?

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: Oh good idea

src/main/resources/auditing.yaml
you-leee: Not sure about that.. @dbt ?

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: yea it should, forgot to do that bit. ty for catching!

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: is that better?

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: yes. whoops.

you-leee: Sorry, the first expression should be 
`val orgAcls = tableAclsByOrg[orgId]?.toMutableList() ?: mutableListOf()`

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
you-leee: Well, `union` will first create a `LinkedHashSet` with the elements of the first collection (`tableAclsByOrg.keys`), so it will have an expected size of that and then add all the elements from the second collection (`columnAclsByOrg.keys`), whereas `plus` will create an empty `LinkedHashSet` with the expected size of both collections and then add all elements from both collections. So if there are more elements in the second collection, it is better to use `plus`

```
public infix fun <T> Iterable<T>.union(other: Iterable<T>): Set<T> {
    val set = this.toMutableSet()
    set.addAll(other)
    return set
}
```

```
public operator fun <T> Set<T>.plus(elements: Iterable<T>): Set<T> {
    val result = LinkedHashSet<T>(mapCapacity(elements.collectionSizeOrNull()?.let { this.size + it } ?: this.size * 2))
    result.addAll(this)
    result.addAll(elements)
    return result
}
```

pckeyes: Hmm well we won't know which collection has more elements so I can switch to plus to be safe

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -179,47 +183,123 @@ class ExternalDatabaseManagementService(
         return columnNamesByTableName
     }
 
+    /*UPDATE*/
+    fun updateOrganizationExternalDatabaseTable(orgId: UUID, tableFqnToId: Pair<String, UUID>, update: MetadataUpdate) {
+        update.name.ifPresent {
+            val newTableFqn = FullQualifiedName(orgId.toString(), it)
+            val oldTableName = getNameFromFqnString(tableFqnToId.first)
+            val dbName = PostgresDatabases.buildOrganizationDatabaseName(orgId)
+            assemblerConnectionManager.connect(dbName).let { dataSource ->
```
makosblade: just curious, why do we use `assemblerConnectionManager.connect(dbName)` instead of `hds` in these methods?

pckeyes: assemblerConnectionManager connects to atlas, which is where these tables are stored.

src/main/resources/auditing.yaml
dbt: I added it so people didn't keep commenting out the body of the audit method when running shuttle jobs, yeah :)

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -179,47 +183,123 @@ class ExternalDatabaseManagementService(
...
+                dataSource.connection.createStatement().use { stmt ->
```
dbt: dataSource.connection checks out a connection from the pool -- you need to close it.

conversely you don't need to close statements for connections you're closing, so you can replace this with
dataSource.connection.use { conn -> 
  stmt = conn.createStatement()
   stmt.execute(...)
}

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -179,47 +183,123 @@ class ExternalDatabaseManagementService(
...
+                    .entrySet(belongsToTable(tableId))
+                    .map { FullQualifiedName(tableId.toString(), it.value.name).toString() to it.key }
+                    .toMap()
+            val columnsByTable = mapOf(tableFqnToId to columnIdByFqn)
+            deleteOrganizationExternalDatabaseColumns(orgId, columnsByTable)
+
+            //delete tables from postgres
+            val dbName = PostgresDatabases.buildOrganizationDatabaseName(orgId)
+            assemblerConnectionManager.connect(dbName).let { dataSource ->
+                dataSource.connection.createStatement().use { stmt ->
```
dbt: same thing here -- put the connection in the use block

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -179,47 +183,123 @@ class ExternalDatabaseManagementService(
...
+            val tableName = getNameFromFqnString(tableFqnToId.first)
+            val tableId = tableFqnToId.second
+            val columnNames = columnIdsByFqn.keys.map { getNameFromFqnString(it) }.toSet()
+            val columnIds = columnIdsByFqn.values.toSet()
+
+            //delete columns from postgres
+            val dropColumnsSql = createDropColumnSql(columnNames)
+            val dbName = PostgresDatabases.buildOrganizationDatabaseName(orgId)
+            assemblerConnectionManager.connect(dbName).let {
+                it.connection.createStatement().use { stmt ->
```
dbt: connection.use

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -271,12 +351,29 @@ class ExternalDatabaseManagementService(
...
+            organizationExternalDatabaseColumns.getValue(it.aclKey[1]).organizationId
+        }
+        val orgIds = tableAclsByOrg.keys + columnAclsByOrg.keys
+
+        orgIds.forEach { orgId ->
+            val orgAcls = tableAclsByOrg[orgId]?.toMutableList() ?: mutableListOf()
+            columnAclsByOrg[orgId]?.let { orgAcls.addAll(it) }
             val dbName = PostgresDatabases.buildOrganizationDatabaseName(orgId)
             assemblerConnectionManager.connect(dbName).let { dataSource ->
                 val conn = dataSource.connection
```
dbt: I realize this isn't your code but it also leaks db connections all over the place.

acm.connect(dbname).connection.use { conn -> 
}

and below where it does dataSource.connection.autoCommit also leaks a connection, change that to use conn also.

pckeyes: Oh it's all my code lol just from previous PRs. Cleaning it all up!

PR #769, opened 2020-02-08
LATTICE-2395 Fix HazelcastMap.values()

dbt:
- Also valueOf


src/main/java/com/openlattice/hazelcast/UniqueInstanceManager.kt
```diff
@@ -2,26 +2,40 @@ package com.openlattice.hazelcast
...
         override fun check() {
             throw IllegalStateException("This is a duplicate of a previous id, constructed here", previous)
         }
+
+        override fun ok(): Boolean {
+            return false
+        }
     }
 
     fun checkInstance(key: String): InstanceChecker {
```
you-leee: Extremely minor, but there is an extra semicolon in the next line

src/main/java/com/openlattice/hazelcast/HazelcastMap.kt
```diff
@@ -59,13 +59,17 @@ import com.openlattice.rhizome.hazelcast.DelegatedStringSet
...
-import com.openlattice.rhizome.service.QueueState
-import org.slf4j.LoggerFactory
 import java.util.*
 
 class HazelcastMap<K, V> internal constructor(val name: String) : TypedMapIdentifier<K, V> {
     private val checker = instanceChecker.checkInstance(name)
 
+    init {
+        if (checker.ok()) {
+            valuesCache.add(this)
```
you-leee: How come we use a list as a cache and not a map, where the keys are the names?

dbt: To keep the same sorting guarantees as the previous enum. Probably not necessary though.

PR #775, opened 2020-02-14
DataTransporter service for new materialized views and top utilizers

dbt:


src/main/kotlin/com/openlattice/postgres/PostgresDataTables.kt
```diff
@@ -142,6 +143,8 @@ class PostgresDataTables {
                     .name(prefix+"_read_data_idx")
                     .ifNotExists()
 
+            val needsPropagateIndex = PostgresExpressionIndexDefinition(tableDefinition, "(${LAST_WRITE.name} > ${LAST_PROPAGATE.name})")
```
katherinebernstein: Is last_propagate being used to mean that a row needs to be (re)materialized now?

dbt: Yes.

katherinebernstein: Are we going to introduce a new column for old propagation once graph queries are fixed (ie auto-setting data values based on neighbor data values, etc) or are we just scrapping that feature?

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -119,6 +119,7 @@ private fun mapOuterMetaDataToColumnSql(metadataOption: MetadataOption): String
         // TODO should be just last_write with comma prefix after empty rows are eliminated https://jira.openlattice.com/browse/LATTICE-2254
         MetadataOption.LAST_WRITE -> ",max(${LAST_WRITE.name}) AS ${mapMetaDataToColumnName(metadataOption)}"
         MetadataOption.ENTITY_KEY_IDS -> ",array_agg(${ORIGIN_ID.name}) as ${ENTITY_KEY_IDS_COL.name}"
+        MetadataOption.VERSION -> ",max(${VERSION.name}) As ${mapMetaDataToColumnName(metadataOption)}"
```
you-leee: Just a minor thing, do we need this metadata option anymore? What was the purpose of it?

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
```diff
@@ -0,0 +1,90 @@
...
+        val propertyIds = entry.value.columns.keys
+        val entitySetIds = entitySets.map { it.id }.toSet()
+        val transporter = acm.connect("transporter")
+        entitySets.filter {
+            if (it.isLinking) {
+                logger.error("Skipping linking entity set {} ({})", it.name, it.id)
+            }
+            !it.isLinking
+        }
+        if (!hasModifiedData(transporter, entitySetIds, propertyIds, partitions)) {
```
you-leee: Won't this call be too big? It's querying with all the entity set ids of an entity type.
We used flags before to flag materialized entity sets, when data was updated

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
dbt: correct, we don't.

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
dbt: the where exists (subquery) should be an extremely optimized path, but I can't know for sure without benchmarking this against prod-sized data.

src/main/kotlin/com/openlattice/postgres/PostgresDataTables.kt
dbt: We can add a new column later when we get to graph propagation.

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
you-leee: 👍

src/main/kotlin/com/openlattice/postgres/PostgresDataTables.kt
katherinebernstein: Might be worth renaming this column then? To avoid confusion with graph propagation

katherinebernstein: To like last_materialize or last_assemble or last_transport or something

src/main/kotlin/com/openlattice/transporter/TransporterUtils.kt
```diff
@@ -0,0 +1,71 @@
...
+            " AND ${PostgresColumn.PROPERTY_TYPE_ID.name} = '${id}' " +
+            " AND ${PostgresColumn.ORIGIN_ID.name} = '${IdConstants.EMPTY_ORIGIN_ID.id}' " +
+            " AND ${DataTables.LAST_WRITE.name} > ${PostgresColumn.LAST_PROPAGATE.name} " +
+            "RETURNING ${PostgresColumn.ENTITY_SET_ID.name}, ${PostgresColumn.ID.name}, ${PostgresColumn.VERSION.name}, " +
+            " CASE WHEN ${PostgresColumn.VERSION.name} > 0 then $srcCol else null end as $destCol"
+
+    val pk = listOf(PostgresColumn.ENTITY_SET_ID, PostgresColumn.ID).joinToString(", ") { it.name }
+    val cols = "$pk,$destCol"
+
+    val modifyDestination = "INSERT INTO $destTable " +
```
you-leee: I'm not 100% sure about this, but won't this insert also the null values, of the `propCol `, where the version was <=0 in data table?

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
```diff
@@ -0,0 +1,129 @@
...
+        }
+        futures.forEach { it.second.get() }
+        val setsPolled = futures.map { it.first }.sum()
+        val duration = timer.observeDuration()
+        logger.debug("Total poll duration time for {} entity sets in {} entity types: {} sec", setsPolled, futures.size, duration)
+    }
+
+    private fun validEntitySets(entityTypeId: UUID): Set<EntitySet> {
+        return entitySetService.getEntitySetsOfType(entityTypeId)
+                .filter { !it.isLinking && !it.flags.contains(EntitySetFlag.AUDIT) }.toSet()
```
you-leee: How/where will be linking entity sets be materialized?

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
```diff
@@ -0,0 +1,129 @@
...
+                .help("time to do one database sync poll")
+                .register()
+    }
+
+    private lateinit var acm: AssemblerConnectionManager
+    private lateinit var transporter: HikariDataSource
+
+    private val transporterState = HazelcastMap.TRANSPORTER_DB_COLUMNS.getMap( hazelcastInstance )
+
+    private fun syncTable(et: EntityType): Optional<Future<*>> {
```
you-leee: Will hard deletes be handled in background task?

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
dbt: views for linking entity sets shouldn't need to be copied explicitly.

```
create or replace view "EntitySetName" as
select
  es.id as entity_set_id,
  ids.linking_id as id, -- todo: obfuscate
  array_agg("property-uuid") FILTER (WHERE "property-uuid" is not null) as "property.name", -- multivalued = true
  (array_agg("property-uuid") FILTER (WHERE "property-uuid" is not null))[1] as "property.name", -- multivalued = false
from entity_sets es 
  join ids on (ids.entity_set_id = any(es.linked_entity_sets))
  join "et_entity-type-uuid" t on (ids.id = t.id)
where es.id = 'entity-set-uuid'
  and es.linking_id is not null
group by es.id, ids.linking_id;
```

(We will also need to create views for non-linked entity sets to deal with subsetting by entity set id, using property name instead of uuid for columns, etc).

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
dbt: I updated this to use the ids table, which should be even faster.

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
dbt: I just pushed a change to add hard deletes.

syncTable() is just there to sync the materialized table definition, though, not to handle data writes.

src/main/kotlin/com/openlattice/transporter/TransporterUtils.kt
```diff
@@ -0,0 +1,110 @@
...
+            " AND ${DataTables.LAST_WRITE.name} > ${PostgresColumn.LAST_PROPAGATE.name} " +
+            "RETURNING ${PostgresColumn.ENTITY_SET_ID.name}, ${PostgresColumn.ID.name}, ${PostgresColumn.VERSION.name}"
+    val createMissingRows = "INSERT INTO $destTable ($pk) " +
+            "SELECT $pk " +
+            "FROM src " +
+            "WHERE ${PostgresColumn.VERSION.name} > 0 " +
+            "ON CONFLICT ($pk) DO NOTHING"
+    val deleteRows = "DELETE FROM $destTable t " +
+            "USING src " +
+            "WHERE src.${PostgresColumn.VERSION.name} <= 0 AND " +
```
you-leee: > I just pushed a change to add hard deletes.
>

This will I think only handle hard deletes , if they are not yet picked up by the background task (that picks up rows where version = 0), because after that they will be permanently deleted. Soft deletes should be fine I think this way.

you-leee: Oh, or will that task wait until last_propagate (or last_materialize..) is set before deleting entities?

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
```diff
@@ -0,0 +1,97 @@
...
+        transport(entry)
+        return null
+    }
+
+    private fun transport(entry: Map.Entry<UUID, TransporterColumnSet>) {
+        if (entitySets.isEmpty() || partitions.isEmpty()) {
+            return
+        }
+        val tableName = tableName(entry.key)
+        val propertyIds = entry.value.columns.keys
```
makosblade: I believe this is unused

src/main/kotlin/com/openlattice/transporter/processors/TransporterSynchronizeTableDefinitionEntryProcessor.kt
```diff
@@ -0,0 +1,78 @@
...
+        AbstractRhizomeEntryProcessor<UUID, TransporterColumnSet, Void?>(),
+        Offloadable,
+        AssemblerConnectionManagerDependent<TransporterSynchronizeTableDefinitionEntryProcessor> {
+    companion object {
+        private val logger = LoggerFactory.getLogger(TransporterSynchronizeTableDefinitionEntryProcessor::class.java)
+    }
+
+    private lateinit var acm: AssemblerConnectionManager
+
+    override fun process(entry: MutableMap.MutableEntry<UUID, TransporterColumnSet>): Void? {
```
makosblade: Any reason this and `transport()` can't be fused together?

src/main/kotlin/com/openlattice/transporter/TransporterUtils.kt
```diff
@@ -0,0 +1,110 @@
+package com.openlattice.transporter
```
makosblade: Any chance you can name this class `TransporterSqlUtils` to head off the `Utils` => dumping grounds risk?

src/main/kotlin/com/openlattice/postgres/PostgresDataTables.kt
dbt: The more I think about it the more I think a new column is safer.

src/main/kotlin/com/openlattice/transporter/TransporterUtils.kt
dbt: Yeah we'll need to deal with that in BackgroundIndexedEntitiesDeletionService, but I'm waiting on that PR until this all shakes out. :)

src/main/kotlin/com/openlattice/transporter/TransporterUtils.kt
dbt: I think TransporterQueries is probably the right name.

src/main/kotlin/com/openlattice/transporter/processors/TransporterSynchronizeTableDefinitionEntryProcessor.kt
dbt: Honestly just to avoid having to keep jamming nulls into the early returns.

src/main/kotlin/com/openlattice/transporter/TransporterQueries.kt
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * column bindings are
+ * 1 - partitions array
+ * 2 - entity set ids array
+ *
+ */
+fun updateOneBatchForProperty(destTable: String, propId: UUID, propCol: TransporterColumn): String {
+    val srcCol = propCol.srcCol
+    val destCol = propCol.destColName
+    val dataCol = "CASE WHEN VERSION > 0 THEN $srcCol else null END as $destCol"
```
katherinebernstein: What do `srcCol` and `destCol` map to? And why is the decision about which contains data affected by `version`?

src/main/kotlin/com/openlattice/transporter/TransporterQueries.kt
```diff
@@ -0,0 +1,145 @@
...
+            "SELECT ${ENTITY_SET_ID.name}, " +
+            " ${LINKING_ID.name} AS ${ID_VALUE.name}, " +
+            " ${ID_VALUE.name} AS ${ORIGIN_ID.name} " +
+            "FROM src " +
+            "WHERE ${VERSION.name} > 0 " +
+            " AND ${LINKING_ID.name} IS NOT NULL " +
+            "ON CONFLICT ($pk) DO NOTHING"
+    val deleteRows = "DELETE FROM $destTable t " +
+            "USING src " +
+            "WHERE src.${VERSION.name} <= 0 " +
```
katherinebernstein: I'm not sure what the current state of this is, but there should theoretically be a job that finds ids with `version = 0` (i.e. hard deleted), and goes and does all the cleanup required to make sure that entity is removed everywhere (data/ids/e table, materialized views, elasticsearch, linking clusters, etc). Seems like a good idea to include <= 0 in case this gets hit first, but was there any code added to a background hard-delete-cleanup job to make sure the rows will for sure be removed, even if that background job is executed before this one? (if such a job exists, lol)

src/main/kotlin/com/openlattice/transporter/TransporterQueries.kt
```diff
@@ -0,0 +1,145 @@
...
+            " AND t.${ENTITY_SET_ID.name} = src.${ENTITY_SET_ID.name} " +
+            " AND t.${ID_VALUE.name} in (src.${ID_VALUE.name},src.${LINKING_ID.name})"
+    return "WITH src as ($updateLastPropagate), " +
+            "inserts as ($createMissingRows)," +
+            "insertLinks as ($createMissingLinkedRows) " +
+            deleteRows
+}
+
+fun addAllMissingColumnsQuery(table: PostgresTableDefinition): String =
```
katherinebernstein: unclear if we are about this right now, but property types can also be removed from entity types -- are we going to remove those columns in that case? (or just leave that data in case it's re-added and remove it from the MVs)

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
```diff
@@ -0,0 +1,137 @@
...
+            futures.forEach { it.second.get() }
+        } finally {
+            val duration = timer.observeDuration()
+            logger.debug("Total poll duration time for {} entity sets in {} entity types: {} sec", setsPolled, futures.size, duration)
+        }
+    }
+
+    private fun validEntitySets(entityTypeId: UUID): Set<EntitySet> {
+        return entitySetService.getEntitySetsOfType(entityTypeId)
+                .filter { !it.isLinking && !it.flags.contains(EntitySetFlag.AUDIT) }.toSet()
```
katherinebernstein: we *will* theoretically at some point need to be able to run graph queries on audit data, unless we change our approach for feeds.... 😬

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
```diff
@@ -0,0 +1,137 @@
...
+        val entityType = e.associationType.associationEntityType
+        if (entityType != null) {
+            this.syncTable(entityType)
+        }
+    }
+
+    @Subscribe
+    fun handleEntityTypeDeleted(e: EntityTypeDeletedEvent) {
+        this.transporterState.delete(e.entityTypeId)
+//        TODO("And drop the table")
```
katherinebernstein: are these TODOs for the initial release?

src/main/kotlin/com/openlattice/transporter/services/TransporterConfiguration.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.kryptnostic.rhizome.configuration.annotation.ReloadableConfiguration
+import java.util.*
+
+@ReloadableConfiguration(uri="transporter.yaml")
+data class TransporterConfiguration(
+        @JsonProperty val server: Properties,
+        @JsonProperty val ssl: Boolean = true,
+        @JsonProperty val once: Boolean = true
```
katherinebernstein: what does `once` affect?

src/test/kotlin/com/openlattice/transporter/TransporterQueriesTest.kt
```diff
@@ -0,0 +1,126 @@
...
+                        " and property_type_id = '${propertyTypeId}'").use {rs ->
+                    rs.next()
+                    rs.getInt(1)
+                }
+            }
+        }
+        val query = updateOneBatchForProperty(table, propertyTypeId, propCol)
+        val updateCount = transporter.connection.use {conn->
+            conn.prepareStatement(query).use {ps ->
+                ps.setArray(1, PostgresArrays.createIntArray(conn, (0..257).toList()))
```
katherinebernstein: sad that this 258 count partitions list needs to be accounted for :(

src/main/kotlin/com/openlattice/transporter/TransporterQueries.kt
you-leee: I think the plan is here, that that job won't run on entries, where `transportTimestampColumn < last_write`.. but both solutions would be fine I think.

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
```diff
@@ -0,0 +1,109 @@
...
+        return null
+    }
+
+    private fun transport(entry: Map.Entry<UUID, TransporterColumnSet>) {
+        if (entitySets.isEmpty() || partitions.isEmpty()) {
+            return
+        }
+        val tableName = tableName(entry.key)
+        val entitySetIds = entitySets.map { it.id }.toSet()
+        val transporter = acm.connect("transporter")
```
you-leee: Might be worth to put `transporter` into a constant and use it in this file and others too

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
```diff
@@ -0,0 +1,137 @@
...
+            Optional.empty()
+        }
+    }
+
+    override fun init(acm: AssemblerConnectionManager): Void? {
+        this.acm = acm
+        this.transporter = acm.connect("transporter")
+        executor.submit {
+            val entityTypes = dataModelService.entityTypes.toList()
+            logger.info("initializing DataTransporterService with {} types", entityTypes.size)
```
you-leee: Just a minor thing, but can you please use uppercase for the log entries first word?

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
```diff
@@ -0,0 +1,137 @@
...
+    fun handleClearAllData(e: ClearAllDataEvent) {
+        TODO("truncate all tables")
+    }
+
+    @Subscribe
+    fun handlePropertyTypesAddedToEntityTypeEvent(e: PropertyTypesAddedToEntityTypeEvent) {
+        this.transporterState.executeOnKey(e.entityType.id, TransporterSynchronizeTableDefinitionEntryProcessor(e.newPropertyTypes).init(acm))
+    }
+}
+
```
you-leee: I think we should also handle entity type and property type name changes if the tables columns use their names and not ids and case of linking entity sets, if linked entity set gets added/removed the data needs to be changed for those tables.

you-leee:
Where will be the actual materialization handled?

src/main/kotlin/com/openlattice/transporter/TransporterQueries.kt
katherinebernstein: Does that affect whether or not we need code in a background deletion task?

Assuming we have a cleanup job that deletes rows from the data table with version = 0 -- if that background job finds a hard-deleted row before this background job finds it, then unless that one also cleans up the transported one, that row won't ever be deleted.

src/main/kotlin/com/openlattice/transporter/TransporterQueries.kt
dbt: srcCol is the result of `PostgresDataTables.getColumnDefinition()` on the property, and destCol is the uuid of the property, but quoted. The constructor for TransporterColumn isn't well documented, I'll update that (and rename the variables to something a little more clear).

src/main/kotlin/com/openlattice/transporter/TransporterQueries.kt
dbt: `BackgroundIndexedEntitiesDeletionService.selectDeletedIds` is where we select hard deleted entity ids for removal. I've updated it here: openlattice/indexer#117

src/main/kotlin/com/openlattice/transporter/TransporterQueries.kt
dbt: We should remove the columns eventually, yes. EdmService only allows this when there aren't any entity sets for a data type, so there should never be stale data in the column (but there could be an all-nulls column hanging out forever).

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
```diff
@@ -0,0 +1,109 @@
...
+import java.util.*
+
+class TransporterPropagateDataEntryProcessor(val entitySets: Set<EntitySet>, val partitions: Collection<Int>):
+        AbstractReadOnlyRhizomeEntryProcessor<UUID, TransporterColumnSet, Void?>(),
+        Offloadable,
+        AssemblerConnectionManagerDependent<TransporterPropagateDataEntryProcessor>
+{
+    companion object {
+        private val logger = LoggerFactory.getLogger(TransporterPropagateDataEntryProcessor::class.java)
+        val id_counter: Counter = Counter.build()
```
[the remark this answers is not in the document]
dbt: prometheus put me a in a go mood. my bad.

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
dbt: I'll put audits back in.

src/main/kotlin/com/openlattice/transporter/services/TransporterConfiguration.kt
dbt: was a way for me to run the full copy once and then exit. I'll remove it.

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private fun transport(entry: Map.Entry<UUID, TransporterColumnSet>) {
+        if (entitySets.isEmpty() || partitions.isEmpty()) {
+            return
+        }
+        val tableName = tableName(entry.key)
+        val entitySetIds = entitySets.map { it.id }.toSet()
+        val transporter = acm.connect("transporter")
+        entitySets.filter {
```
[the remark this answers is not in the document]
dbt: linked entity sets aren't transported, they're reassembled in views. The transporter API shouldn't be called with linked entity sets.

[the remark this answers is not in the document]
you-leee: I think at this point they are already filtered out in `DataTransporterService.validEntitySets` and the return value of the filter is not used.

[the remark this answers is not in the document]
dbt: Yes. Fixed.

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
```diff
@@ -0,0 +1,137 @@
...
+                    transporterState.submitToKey(et.id, TransporterSynchronizeTableDefinitionEntryProcessor(props).init(acm))
+            )
+        } else {
+            Optional.empty()
+        }
+    }
+
+    override fun init(acm: AssemblerConnectionManager): Void? {
+        this.acm = acm
+        this.transporter = acm.connect("transporter")
```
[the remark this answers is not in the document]
dbt: Yeah.. I think it will be saner to put in a transporter.yaml and pass that around instead of abusing ACM this way.

src/test/kotlin/com/openlattice/transporter/TransporterQueriesTest.kt
dbt: It probably doesn't, just needed something there. These tests shouldn't touch real data.

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
dbt: Property type name changes should be handled by assembler views, correct?

dbt:
> Where will be the actual materialization handled?

TransporterPropagateDataEntryProcessor does the actual data copying.
Regular wildcard updates are triggered by TransporterRunSyncTask, which runs every 30 seconds. Datastore can and should call the entryprocessor when data is saved to the db.

src/main/kotlin/com/openlattice/transporter/processors/TransporterSynchronizeTableDefinitionEntryProcessor.kt
```diff
@@ -0,0 +1,66 @@
...
+        }
+        val newColumns = entry.value.withProperties(newProps)
+        val table = tableDefinition(entry.key, newColumns.columns.values.map { it.destCol() })
+        transporter.connection.use { conn ->
+            var lastSql = ""
+            try {
+                conn.createStatement().use { st ->
+                    lastSql = table.createTableQuery()
+                    st.execute(lastSql)
+                    lastSql = addAllMissingColumnsQuery(table)
```
you-leee: Maybe I'm missing here something, but won't the `createTableQuery()` add the columns already?

src/main/kotlin/com/openlattice/transporter/processors/TransporterSynchronizeTableDefinitionEntryProcessor.kt
```diff
@@ -0,0 +1,66 @@
...
+    private lateinit var acm: AssemblerConnectionManager
+
+    override fun process(entry: MutableMap.MutableEntry<UUID, TransporterColumnSet>): Void? {
+        // transport is a separate function strictly to allow for easier early returns
+        transport(entry)
+        return null
+    }
+
+    private fun transport(entry: MutableMap.MutableEntry<UUID, TransporterColumnSet>) {
+        val transporter = acm.connect("transporter")
```
you-leee: Minor thing, but I think, we shouldn't connect before there is a change for early return

src/main/kotlin/com/openlattice/transporter/processors/TransporterSynchronizeTableDefinitionEntryProcessor.kt
you-leee: Oh, nvm, this is for handling both new columns and table creation, right?

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
you-leee: Yes, you are right, sorry! I was a bit confused over whether we use UUID or names for property types.

src/main/kotlin/com/openlattice/transporter/types/TransporterColumnSet.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import com.openlattice.ApiUtil
+import com.openlattice.edm.PostgresEdmTypeConverter
+import com.openlattice.edm.type.PropertyType
+import com.openlattice.postgres.PostgresColumnDefinition
+import com.openlattice.postgres.PostgresDataTables
+import com.openlattice.postgres.PostgresDatatype
+import java.util.*
+
+data class TransporterColumn(val srcCol: String, val destColName: String, val dataType: PostgresDatatype) {
```
you-leee: Can you please add some javadoc for the variables and the `destCol()` function? (Also for `TransporterColumnSet `)

src/test/resources/assembler.yaml
```diff
@@ -0,0 +1,12 @@
+server:
+  jdbcUrl: "jdbc:postgresql://localhost:5432"
+  username: "oltest"
+  password: "test"
+  maximumPoolSize: 5
+  connectionTimeout: 60000
+foreignHost: "localhost"
+foreignDbName: "openlattice"
+foreignPort: 5433
```
you-leee: I think this should be 5432

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
```diff
@@ -0,0 +1,109 @@
...
+                val entitySetArray = PostgresArrays.createUuidArray(conn, entitySetIds)
+                val partitions = PostgresArrays.createIntArray(conn, partitions)
+                lastSql = updateIdsForEntitySets(tableName)
+                val idUpdates = conn.prepareStatement(lastSql).use {ps ->
+                    ps.setArray(1, partitions)
+                    ps.setArray(2, entitySetArray)
+                    ps.executeUpdate()
+                }
+                val colUpdates = entry.value.map { (id, col) ->
+                    val query = updateOneBatchForProperty(tableName, id, col)
```
you-leee: Why assign this string to a separate variable and not just to `lastSql`?

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
you-leee: The return value of this filter is still ignored

you-leee:
> > Where will be the actual materialization handled?
> 
> TransporterPropagateDataEntryProcessor does the actual data copying.
> Regular wildcard updates are triggered by TransporterRunSyncTask, which runs every 30 seconds. Datastore can and should call the entryprocessor when data is saved to the db.

Thanks! I was actually thinking about the assembler views, what the organization users will access.

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
dbt: Fixed.

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
dbt: I got rid of the ACM dependency completely and moved to a separate configuration file. No reason to hardcode that assumption everywhere.

src/main/kotlin/com/openlattice/transporter/processors/TransporterSynchronizeTableDefinitionEntryProcessor.kt
dbt: Yes.

src/main/kotlin/com/openlattice/transporter/types/TransporterColumnSet.kt
dbt: I renamed them, because they were definitely confusing. I can add some docs if they're still confusing.

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
dbt: Fixed.

dbt:
> Thanks! I was actually thinking about the assembler views, what the organization users will access.

We should put that into Assembler, like it was before -- I don't want this to also have to know about organizations and other databases.

you-leee:
> > Thanks! I was actually thinking about the assembler views, what the organization users will access.
> 
> We should put that into Assembler, like it was before -- I don't want this to also have to know about organizations and other databases.

Cool, thanks for the explanation! I guess that will be a different PR then

dbt:
> Cool, thanks for the explanation! I guess that will be a different PR then

Yeah this is already way too big to review.

src/main/kotlin/com/openlattice/transporter/processors/TransporterPropagateDataEntryProcessor.kt
```diff
@@ -0,0 +1,109 @@
...
+        transport(entry)
+        return null
+    }
+
+    private fun transport(entry: Map.Entry<UUID, TransporterColumnSet>) {
+        if (entitySets.isEmpty() || partitions.isEmpty()) {
+            return
+        }
+        val tableName = tableName(entry.key)
+        entitySets.filter { it.isLinking }.forEach {
```
you-leee: Is this extra iteration only to log filtered linking entity sets?

dbt: Yes. We can skip all the internal islinking checks here if we trust caller paths.

src/main/kotlin/com/openlattice/transporter/services/DataTransporterService.kt
```diff
@@ -0,0 +1,193 @@
...
+                .help("Time to do a refresh for a single entity set after a write event")
+                .register()
+        val errorCount: Counter = Counter.build()
+                .namespace(transporterNamespace)
+                .name("error_count")
+                .help("Count of errors during transport operations")
+                .register()
+    }
+
+    private val transporter = data.datastore()
```
makosblade: This appears to be unused

PR #778, opened 2020-02-19
Bugfix/lattice 2268 fix delete entities and neighbours

you-leee:


src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
```diff
@@ -140,38 +136,94 @@ class DataDeletionService(
 
         val includeClearedEdges = deleteType == DeleteType.Hard
 
-        val entitySetIdToEntityDataKeysMap: Map<UUID, List<EntityDataKey>> = dgm
+        val entitySetIdToEntityDataKeys: Map<UUID, Set<UUID>> = dgm
                 .getEdgesConnectedToEntities(entitySetId, entityKeyIds, includeClearedEdges)
-                .filter { edge -> edge.dst.entitySetId == entitySetId && filteringNeighborEntitySetIds.contains(edge.src.entitySetId) || edge.src.entitySetId == entitySetId && filteringNeighborEntitySetIds.contains(edge.dst.entitySetId) }
+                .filter { edge ->
+                    edge.dst.entitySetId == entitySetId && filteringNeighborEntitySetIds.contains(edge.src.entitySetId)
```
katherinebernstein: this looks like it needs parenthesis, based on how I'm interpreting this logic

katherinebernstein: ie `(a && b) || (c && d)`

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
```diff
@@ -140,38 +136,94 @@ class DataDeletionService(
...
+        val authorizedPropertiesByEntitySets = getAuthorizedPropertyTypesForDeleteByEntitySet(
+                setOf(entitySetId) + entitySetIdToEntityDataKeys.keys, // include original entity set too
+                deleteType,
+                Optional.empty(),
+                principals
+        )
+
+        /* Check association permissions */
+
+        // only original entity set can be association entity set
```
katherinebernstein: why can only the original entity set be an association entity set?

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
```diff
@@ -140,38 +136,94 @@ class DataDeletionService(
...
+                    authorizedAssociationPropertiesByEntitySets.getValue(entitySetId)
+            )
+            numUpdates += associationNumUpdates
+        }
 
-        // TODO what is/was this for?
-        val neighborDeleteEvents = Lists.newArrayList<AuditableEvent>()
 
-        numUpdates += entitySetIdToEntityDataKeysMap.entries.stream().mapToInt { entry ->
+        /* 3 - Delete neighbors */
```
katherinebernstein: Seems like this will be more performant if we figure out exactly which entities + edges will need to be deleted up front, and then do a bulk delete at the end

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
you-leee: Even without parenthesis it would interpret it this way since AND > OR, but will add the them to make it clear.

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
you-leee: Because all the other entity sets in `entitySetIdToEntityDataKeys` are either src or dst entity set ids from edges:
```
 val entitySetIdToEntityDataKeys: Map<UUID, Set<UUID>> = dgm
        .getEdgesConnectedToEntities(entitySetId, entityKeyIds, includeClearedEdges)
        ...
        .flatMap { edge -> setOf(edge.src, edge.dst) }
        ...
```

Unless src and dst entities in edges can come from association entity sets.

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
you-leee: Maybe I'm missing here something, but I don't think we can do a batch delete here since right now we delete by entity set (and partition) in PEDQS and the `entitySetIdToEntityDataKeysMap` is already mapped by entity set ids.
Or did you mean something else by bulk delete?

katherinebernstein: yeah true, although you could still parallelize by submitting all the clear/delete calls at once instead of calling sequentially and waiting for each one to complete before kicking off the next.

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
```diff
@@ -135,43 +131,99 @@ class DataDeletionService(
...
+        /* Delete neighbors */
 
-        numUpdates += entitySetIdToEntityDataKeysMap.entries.stream().mapToInt { entry ->
+        numUpdates += entitySetIdToEntityDataKeys.entries.stream().mapToInt { entry ->
             val neighborEntitySetId = entry.key
-            val neighborEntityKeyIds = entry.value.map { it.entityKeyId }.toSet()
+            val neighborEntityKeyIds = entry.value
 
-            val neighborWriteEvent = clearOrDeleteEntities(neighborEntitySetId,
+            val neighborWriteEvent = clearOrDeleteAuthorizedEntities(
```
katherinebernstein: I could be reading this wrong, but I don't think you need to call `clearOrDeleteAuthorizedEntities` if you're also calling `clearOrDeleteAuthorizedAssociations`

you-leee: `clearOrDeleteAuthorizedAssociations` only deletes the connected association entities of the requested entities, so I think they both need to be called.. although in the opposite order, so the edges connected to entities won't be lost

you-leee: `clearOrDeleteAuthorizedAssociations` collects all connected edges first then it maps them to only contain the edge entityKeyIds from the DataEdgeKeys and then it deletes data, ids and edges of them

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
you-leee: 👍

src/main/kotlin/com/openlattice/data/storage/DataDeletionService.kt
```diff
@@ -248,32 +318,14 @@ class DataDeletionService(
...
-
+    /* Association deletion and auth */
 
-    private fun authorizeAndDeleteAssociations(
+    /**
+     * Checks authorizations on connected association entities and clears/deletes them along with edges connected to
+     * them.
+     * If entity set is an association entity set, it only deletes its edges.
+     */
+    private fun clearOrDeleteAssociations(
```
pckeyes: Out of curiosity, if a person doesn't have auth to clear or delete an association entity but has auth to delete the entity to which it is attached, will we end up with associations attached to nothing?

you-leee: That is exactly what we are tying to avoid :)

When we do a simple delete (not neighbors delete) of non-association entities
1. We first check authorizations on the requested entity set + properties (and fail call if checks fail)
2. Then we call this function, which also first checks authorizations on connected association entities and just then deletes association entities and the edges connected to them
3. As the last step we delete the normal entities

If we delete from an association entity set, we don't have to check connected associations, so we only delete its' edges in this function.

And in case of neighbors delete (when we delete from an entity set + its requested neighboring entity sets), we do basically the same: collect all neighbors, then check authorization, then delete associations and edges and as a last step delete entities.

pckeyes: Gotcha, thank you for laying that out!

PR #780, opened 2020-02-22
Convert OffsetDateTime values to UTC before computing hash

katherinebernstein:
This shouldn't be merged until we run a migration to fix all existing hashes, otherwise we'll end up with lots of duplicate and unmodifiable data


katherinebernstein:
this is probably irrelevant now because of https://github.com/openlattice/api/pull/418

PR #782, opened 2020-02-25
only add members to organization that are not already members

katherinebernstein:


katherinebernstein:
Also open to creating a new entry processor specifically for adding members, because forcing all the other ones to return null feels bad

PR #783, opened 2020-02-26
get counts of entity sets needing linking

jokedurnez:


src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
```diff
@@ -21,6 +21,7 @@
 
 package com.openlattice.linking.graph
 
+import com.google.common.collect.Maps
```
dbt: unused?

src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
```diff
@@ -402,4 +418,12 @@ private val ENTITY_KEY_IDS_NOT_LINKED = "SELECT ${ENTITY_SET_ID.name},${ID.name}
 private val LINKABLE_ENTITY_SET_IDS = "SELECT ${ID.name} " +
         "FROM ${ENTITY_SETS.name} " +
         "WHERE ${ENTITY_TYPE_ID.name} = ANY(?) AND NOT ${ID.name} = ANY(?) AND ${ID.name} = ANY(?) "
-// @formatter:on
+
+private val ENTITY_COUNT_NOT_LINKED = "SELECT ${ENTITY_SET_ID.name},COUNT(${ID.name}) " +
+        "FROM ${IDS.name} " +
+        "WHERE ${PARTITION.name} = ANY(?) " +
+            "AND ${ENTITY_SET_ID.name} = ANY(?) " +
+            "AND ${LAST_LINK.name} < ${LAST_WRITE.name} " +
```
makosblade, line 426: Note that this will select everything that is indexable and not linked but not necessarily everything that *should be linked already* since entities have to be indexed before they can be linked. 
So if this endpoint is going to be used to determine whether things are linking correctly, it'll only count elements that can be linked and won't identify issues caused by lack of indexing (which should be totally fine).

jokedurnez: Sounds like there's anyway no way around it right?

makosblade: Could add `AND ${LAST_INDEX.name} > ${LAST_WRITE.name}`
That way you filter out things that haven't been indexed yet and therefore won't be picked up by linker

jokedurnez: From a user perspective, we'd want to include the non-indexed rows though right?  The user doesn't really care about indexed vs non-indexed.

katherinebernstein: +1 to making 100% sure that this is what we want. There's an important distinction between "stale rows that are not currently eligible for linking" vs "stale rows that should be linked" -- I can see the first case being confusing for people who get back a number and don't ever see it decrease (when in reality it *won't* decrease given its current state), depending what this is being used for.

jokedurnez: @katherinebernstein eventually they should be linked right?  Is this about potential new data that is awaiting indexing?  Or am I not understanding what `AND ${LAST_INDEX.name} > ${LAST_WRITE.name}` would filter out?

src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
katherinebernstein: +1

src/main/kotlin/com/openlattice/linking/graph/PostgresLinkingQueryService.kt
katherinebernstein: Theoretically, as long as indexer is running + working. It's just important to understand the distinction and make sure we're answering the right question here -- # of entities linker hasn't gotten to yet (but should/will) vs # of entities that haven't been (re-)linked after a write (whether or not linker is currently able to pick them up)

katherinebernstein: Ok more clearly, are you trying to answer:
1) How many entities have stale/absent linking_ids?
2) How many entities are ready to be linked but haven't been linked yet?
3) How many entities can't yet be searched according to a current+accurate linking cluster? (throwing this one into the mix as well)

jokedurnez: I think the question is: how many records should be linked but aren't.  Independent of the reason why they aren't linked, whether it's indexer or linker, because that difference is not relevant from a user perspective.  @kengie what do you think?

kengie: I think it should also be #2 entities that should be linked but aren't yet. But could #3 also include entities that should be in #2? It does seem like adding `AND ${LAST_INDEX.name} > ${LAST_WRITE.name}`
could be helpful because then if this endpoint returns a count of 0 but I know there should be more people in the final dataset, we could narrow down the issue to indexer and not linker? 

Also, I can't think of an instance where there would be a person entity set that is in a linked entity set, with "stale rows that are not currently eligible for linking" that shouldn't turn into "stale rows that should be linked". When would that happen?

jokedurnez: ok then this is good to go !

PR #784, opened 2020-03-02
move transforms to cond-client

pckeyes:
With flights moved into conductor-client, in order for hazelcast to deserialize the flight correctly, transforms also need to be in conductor-client. This PR moves transforms and the classes that they depend on into conductor-client.


pckeyes:
> Perfect chance for some Kotlinification but no worries if not :D

Since most of these files are eventually going to be depreciated/deleted I'd rather not spend the time to kotlinify them.

PR #787, opened 2020-03-07
Feature/lattice 2292 row based permissions

pckeyes:
Logic for creating, reading, and deleting row security policies.

For context, row security policies are stored within the `pg_policies` table with the following structure:
```
  Column   |  Type  | Collation | Nullable | Default
------------+--------+-----------+----------+---------
 schemaname | name   |           |          |
 tablename  | name   |           |          |
 policyname | name   |           |          |
 permissive | text   |           |          |
 roles      | name[] |           |          |
 cmd        | text   |           |          |
 qual       | text   |           |          |
 with_check | text   |           |          |
```

`cmd` is the command the policy restricts (eg, select, insert)
`qual` is the filter for visibility on rows that already exist
`with_check` is the filter for creating/altering rows
`permissive` indicates whether applicable policies will be combined using AND or OR
further information can be found here: https://www.postgresql.org/docs/11/sql-createpolicy.html

This PR does not create any infrastructure for storing representations of these row policies in our system. Instead we will rely on this table for storing policies.


src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -379,6 +387,50 @@ class ExternalDatabaseManagementService(
...
 
+    fun addRowSecurityPolicy(orgId: UUID, tableName: String, policyName: String, rowSecurityPolicy: RowSecurityPolicy) {
+        val dbName = PostgresDatabases.buildOrganizationDatabaseName(orgId)
+        val userNames = rowSecurityPolicy.userIds.map { getDBUser(it) }
+        var createRowPolicySql = "CREATE POLICY $policyName ON $tableName " +
+                "FOR ${rowSecurityPolicy.privilege} " +
+                "TO ${userNames.joinToString(", ")} "
+        rowSecurityPolicy.readFilter.ifPresent { createRowPolicySql += "USING $it " }
+        rowSecurityPolicy.writeFilter.ifPresent { createRowPolicySql += "WITH CHECK $it" }
+        assemblerConnectionManager.connect(dbName).let {
```
you-leee: I think we should use `use` instead of `let` to close the connection properly

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -379,6 +387,50 @@ class ExternalDatabaseManagementService(
...
+                    PostgresPrivileges.valueOf(rs.getString(COMMAND_COLUMN)),
+                    Optional.ofNullable(rs.getString(QUAL_COLUMN)),
+                    Optional.ofNullable(rs.getString(CHECK_COLUMN))
+            )
+        }.toSet()
+    }
+
+    fun deleteRowSecurityPolicy(orgId: UUID, tableName: String, policyName: String) {
+        val dbName = PostgresDatabases.buildOrganizationDatabaseName(orgId)
+        assemblerConnectionManager.connect(dbName).let {
```
you-leee: Also here, I think we should use `use`

PR #788, opened 2020-03-10
it feels bad but it works

katherinebernstein:


you-leee:
Are we doing a mechanic job instead?

katherinebernstein:
nope I just renamed the column in sms_information to entity_set_ids -- looks like it's actually been quietly broken for a while

PR #790, opened 2020-03-11
LATTICE 2419 return uuids with table or column objects

pckeyes:
getExternalDatabaseTables() now returns a Map<UUID, OrganizationExternalDatabaseTable> and getExternalDatabaseTablesWithColumns() now returns a Map<Pair<UUID, OrganizationExternalDatabaseTable>, Map<UUID, OrganizationExternalDatabaseColumn>>. This enables using accessCheckForPrincipals instead of isAuthorized for checking permissions.


src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -124,14 +124,20 @@ class ExternalDatabaseManagementService(
         return organizations.keys
     }
 
-    fun getExternalDatabaseTables(orgId: UUID): Set<OrganizationExternalDatabaseTable> {
-        return organizationExternalDatabaseTables.values(belongsToOrganization(orgId)).toSet()
+    fun getExternalDatabaseTables(orgId: UUID): Map<UUID, OrganizationExternalDatabaseTable> {
+        return organizationExternalDatabaseTables.entrySet(belongsToOrganization(orgId))
```
katherinebernstein: seems like you should be able to just `return organizationExternalDatabaseTables.entrySet(belongsToOrganization(orgId))`

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -124,14 +124,20 @@ class ExternalDatabaseManagementService(
...
+                    it.key to it.value
+                }.toMap()
     }
 
-    fun getExternalDatabaseTablesWithColumns(orgId: UUID): Map<OrganizationExternalDatabaseTable, Set<OrganizationExternalDatabaseColumn>> {
+    fun getExternalDatabaseTablesWithColumns(orgId: UUID): Map<Pair<UUID, OrganizationExternalDatabaseTable>, Map<UUID, OrganizationExternalDatabaseColumn>> {
         val tables = getExternalDatabaseTables(orgId)
         return tables.map {
-            it to (organizationExternalDatabaseColumns.values(belongsToTable(it.id)).toSet())
+            Pair(it.key, it.value) to (organizationExternalDatabaseColumns.entrySet(belongsToTable(it.key))
```
katherinebernstein: same thing here with skipping the `.map { .. } .toMap()`

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: that returns a set<map.entry<uuid, table>>, which is fine too. but I will probably still have to do a map{} in the controller.

src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
pckeyes: same thing again with having to move the map{}s to the controller.

PR #791, opened 2020-03-12
Bugfix/lattice 2417 auth0 user sync 1000

you-leee:
The goal of this ticket is to overcome the 1000 user limit of auth0 user api (https://auth0.com/docs/users/search/v3/get-users-endpoint#limitations)

The idea is to first initialize all users with creating a user export job, which will produce a file of all users. We read those in and sync all of them once. This will be the `Auth0SyncInitializationTask`.
After that in the `Auth0SyncTask` we only query the updated users (where lastSync < `updated_at` <= now) via auth0 api and update the users mapstore with them. In the same `Auth0SyncTask` we go through all of the users in the mapstore and sync enrollments and authentication

Users API searchable fields: https://auth0.com/docs/users/search/v3/query-syntax#searchable-fields
User export job: https://auth0.com/docs/api/management/v2?_ga=2.231290719.1973299790.1583518412-1574661569.1583518412#!/Jobs/post_users_exports


src/main/kotlin/com/openlattice/users/export/UserExportJob.kt
```diff
@@ -0,0 +1,66 @@
...
+    init {
+        if (status == JobStatus.completed) {
+            require(location.isPresent) { "Location should be provided if job is completed." }
+        }
+    }
+}
+
+enum class JobStatus {
+    completed,
+    pending,
```
makosblade: Can you all capitals these please?

src/main/kotlin/com/openlattice/users/export/UserExportEntity.kt
```diff
@@ -0,0 +1,91 @@
...
+        private val logger = LoggerFactory.getLogger(UserExportJobRequest::class.java)
+    }
+
+    /**
+     * Submits a user export job to auth0.
+     */
+    fun submitExportJob(exportJob: UserExportJobRequest): Job {
+        val url = baseUrl
+                .newBuilder()
+                .addPathSegments("api/v2/jobs/users-exports")
```
makosblade: Can you extract this to a constant?

src/main/kotlin/com/openlattice/users/export/UserExportEntity.kt
```diff
@@ -0,0 +1,91 @@
...
+                .build()
+                .toString()
+
+        val body = RequestBody.create(MediaType.parse("application/json"), mapper.writeValueAsBytes(exportJob))
```
makosblade: I think you maybe meant to use the above `CONTENT_TYPE_APPLICATION_JSON` here?

src/main/kotlin/com/openlattice/users/export/UserExportEntity.kt
makosblade: Or just the prefix `api/v2/jobs/`

src/main/kotlin/com/openlattice/users/export/UserExportEntity.kt
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    /**
+     * Retrieves a job result from auth0 by [jobId].
+     */
+    fun getJob(jobId: String): UserExportJobResult {
+        val url = baseUrl
+                .newBuilder()
+                .addPathSegments("api/v2/jobs/$jobId")
```
makosblade: Can extract the prefix for this

src/main/kotlin/com/openlattice/users/export/UserExportEntity.kt
you-leee: Good idea, will do!

src/main/kotlin/com/openlattice/users/export/UserExportJob.kt
you-leee: That will cause problems when we try to deserialize `JobStatus` in `UserExportJobResult ` because it's lowercase in the response

src/main/kotlin/com/openlattice/users/Auth0SyncTask.kt
```diff
@@ -22,25 +22,31 @@
...
 
 /**
  * This is the auth0 synchronization task that runs every REFRESH_INTERVAL_MILLIS in Hazelcast. It requires that
  * syncDependencies be initialized within the same JVM in order to function properly.
- *
  */
 class Auth0SyncTask : HazelcastFixedRateTask<Auth0SyncTaskDependencies>, HazelcastTaskDependencies {
     private val syncSemaphore = Semaphore(MAX_JOBS)
+
+    private var initialized = false
```
dbt: I don't think having local state works. The task has to be able to run on any of several conductors, and if the one conductor where the task has run ever restarts, it'll never run again.

This data needs to be stored in some global state. I'm looking into Hazelcast's StatefulTask but I'm not sure it'll work with accessing state outside the runTask method (i.e. in the initializeUsers method).

src/main/kotlin/com/openlattice/users/Auth0Utils.kt
```diff
@@ -6,20 +6,81 @@ import com.auth0.exception.Auth0Exception
...
+    val exportEntity = auth0ApiExtension.userExport()
+    val job = exportEntity.submitExportJob(
+            UserExportJobRequest(listOf(USER_ID, EMAIL, NICKNAME, APP_METADATA, IDENTITIES))
+    )
+
+    // will fail if export job hangs too long with too many requests error (429 status code)
+    // https://auth0.com/docs/policies/rate-limits
+    var exportJobResult = exportEntity.getJob(job.id)
+    while (exportJobResult.status == JobStatus.pending) {
+        exportJobResult = exportEntity.getJob(job.id)
```
dbt: This needs some kind of time-based backoff -- hammering the status endpoint is likely to get us throttled/blacklisted.

src/main/kotlin/com/openlattice/users/export/UserExportJob.kt
dbt: You can pass a string into the constructor and put @JsonValue on a getter for the string and it should use that instead of the default enum NAME.

you-leee: Okay, will do!

src/main/kotlin/com/openlattice/users/Auth0SyncTask.kt
you-leee: Nice catch, you're right!

src/main/kotlin/com/openlattice/users/export/UserExportJob.kt
you-leee: It does not though :(

PR #792, opened 2020-03-13
Feature/lattice 2416 tables as metadata columns as data

pckeyes:
This PR addresses the need to be able to view atlas table metadata without being able to see table data. We will now only grant postgres privileges on columns and not on entire tables. This will enable us to
-Use the OrgExternalDbTable object for permissioning table metadata
-Use the OrgExternalDbColumn object for permissioning table data on a column by column basis. 

This will make the external db objects act more like entity sets. The table will approximate the entity set and the columns will approximate the property types. 

One issue is that postgres privileges on entire tables will have no real meaning in our system, as read permissions on OrgExternalDbTable objects will only mean you can see the metadata. Super open to suggestions about this.

### DIFF
--- a/src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
+++ b/src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
@@ -498,21 +498,6 @@ class ExternalDatabaseManagementService(
         return true
     }
 
-    private fun getMaintainColumnPrivilegesSql(aclKey: AclKey, principal: Principal, tableName: String, dbUser: String): List<String> {
-        val maintenanceStmts = mutableListOf<String>()
-        organizationExternalDatabaseColumns.values(belongsToTable(aclKey.last())).forEach {
-            val columnAclKey = AclKey(listOf(aclKey[0], it.id))
-            val columnAceKey = AceKey(columnAclKey, principal)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -351,29 +351,19 @@ class ExternalDatabaseManagementService(
...
-        //split acls between table and column objects
-        val aclsByType = acls.partition { it.aclKey.size == 1 }
-
-        val tableAclsByOrg = aclsByType.first.groupBy {
-            organizationExternalDatabaseTables.getValue(it.aclKey[0]).organizationId
-        }
-        val columnAclsByOrg = aclsByType.second.groupBy {
+    fun executePrivilegesUpdate(action: Action, columnAcls: List<Acl>) {
+        val columnAclsByOrg = columnAcls.groupBy {
             organizationExternalDatabaseColumns.getValue(it.aclKey[1]).organizationId
```
you-leee: Might be worth to batch this call with a predicate or entry processor and keep it in a separate map with `aclKey[1]` as keys and ` OrganizationExternalDatabaseColumn` as value.
It is also used later in `getTableAndColumnNames`, so that map could just be passed there and we could avoid having multiple calls to the hazelcast map.

PR #794, opened 2020-03-17
LATTICE-2418 pass whole entity set to EntitySetDataDeletedEvent

you-leee:
This pr addresses this problem: When we delete an entity set and with it delete all of its entities, an EntitySetDataDeletedEvent gets posted (when the entity set is still existing) but sometimes by the time a subscriber handles the event, the entity set is already deleted, so it won't find its entity type, partitions...etc and an error gets thrown.


geekbeast:
We should fix entity set deletion instead of continuing to monkey patch.

PR #795, opened 2020-03-18
LATTICE 2423 filter data returned by read permission

pckeyes:
This PR
-Filters table data to only return data from columns on which user has read permissions

### DIFF
--- a/src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
+++ b/src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
@@ -128,11 +128,11 @@ class ExternalDatabaseManagementService(
         return organizationExternalDatabaseTables.values(belongsToOrganization(orgId)).toSet()
     }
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -140,18 +140,22 @@ class ExternalDatabaseManagementService(
...
 
-    fun getExternalDatabaseTableData(orgId: UUID, tableId: UUID, rowCount: Int): Map<UUID, List<Any?>> {
+    fun getExternalDatabaseTableData(
+            orgId: UUID,
+            tableId: UUID,
+            authorizedColumns: Set<OrganizationExternalDatabaseColumn>,
+            rowCount: Int): Map<UUID, List<Any?>> {
         val tableName = organizationExternalDatabaseTables.getValue(tableId).name
-        val columns = organizationExternalDatabaseColumns.values(belongsToTable(tableId))
+        val columnNamesSql = authorizedColumns.map { it.name }.joinToString(", ")
```
you-leee: This could be replaced with `authorizedColumn.joinToString(", ") { it.name }`

PR #800, opened 2020-03-30
quos on column names

jokedurnez:


src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -477,17 +477,17 @@ class ExternalDatabaseManagementService(
     }
 
     private fun createDropColumnSql(columnNames: Set<String>): String {
-        return columnNames.joinToString(", ") { "DROP COLUMN $it" }
+        return columnNames.joinToString(", ") { "DROP COLUMN \"$it\"" }
```
katherinebernstein, line 480: can you use the `quote` method to do this? i.e. `"DROP COLUMN ${quote(it)}"`

(and same for the rest)

PR #805, opened 2020-04-13
Tombstone after successful serialization.

geekbeast:


katherinebernstein:
Have you tested the sql this outputs? Just want to double check that the empty set property values don't cause errors down the line.

geekbeast:
I didn't check the SQL, but PEDQS.upsertEntities the inner values map will be empty and result in no batches being added (which results in a no-op when execute batch is called).

PR #816, opened 2020-05-05
Feature/lattice 2426 scheduled task service

katherinebernstein:


src/main/kotlin/com/openlattice/scheduling/ScheduledTask.kt
```diff
@@ -0,0 +1,10 @@
+package com.openlattice.scheduling
+
+import java.time.OffsetDateTime
+import java.util.*
+
+data class ScheduledTask(
+        val id: UUID,
```
makosblade, line 7: I feel like this should have a default value that is a newly generated UUID

PR #831, opened 2020-05-27
add API to delete a user

katherinebernstein:


katherinebernstein:
> This is insufficient to properly delete a user. All permissions granted to the user should be cleared and the securable object principal corresponding to the user should be removed. In addition, the user should be removed from all organizations to which they belong.

All that should be covered in `HazelcastAuthorizationService.deletePrincipal`

geekbeast:
> > This is insufficient to properly delete a user. All permissions granted to the user should be cleared and the securable object principal corresponding to the user should be removed. In addition, the user should be removed from all organizations to which they belong.
> 
> All that should be covered in `HazelcastAuthorizationService.deletePrincipal`

Is there a change I'm missing here? I don't see that function.

geekbeast:
I see you are referring to HazelcastPrincipalService deletePrincipal not HazelcastAuthorizationService.  Okay looks good.

katherinebernstein:
> I see you are referring to HazelcastPrincipalService deletePrincipal not HazelcastAuthorizationService. Okay looks good.

Oops yeah PrincipalService! Thanks!

PR #836, opened 2020-05-28
Reduce churn of organization when no-op entry processors are executed.

geekbeast:


src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -523,34 +539,37 @@ class HazelcastOrganizationService(
...
+            val modified = it.members.remove(principal)
+            Result(modified, modified)
+        })
     }
 
     fun updateRoleGrant(organizationId: UUID, roleId: UUID, grant: Grant) {
         organizations.executeOnKey(organizationId, OrganizationEntryProcessor {
             it.grants.getOrPut(roleId) { mutableMapOf() }[grant.grantType] = grant
-            null
+            Result(null) //TODO: Being lazy not implementing diff as this should rarely be called.
```
solomontang, line 551: Looks like the build is failing without handling this case tho

```OrganizationEntryProcessorStreamSerializerTest.kt: (39, 13): Null can not be a value of a non-null type OrganizationEntryProcessor.Result```

Comment (geekbeast) [reply src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt:551]: The entry processor in the test wasn't wrapping the result. Should be fixed now.

File: src/main/java/com/openlattice/postgres/ResultSetAdapters.java [hunk 2]
```diff
@@ -40,13 +40,13 @@
 import com.openlattice.collections.EntityTypeCollection;
```
makosblade, line 40: I believe this file change is breaking the build

makosblade: should be good now

PR #837, opened 2020-05-28
Feature/auth using read only entry processor

katherinebernstein:


src/main/kotlin/com/openlattice/authorization/processors/AuthorizationEntryProcessor.kt
```diff
@@ -0,0 +1,15 @@
+package com.openlattice.authorization.processors
+
+import com.openlattice.authorization.AceKey
+import com.openlattice.authorization.AceValue
+import com.openlattice.authorization.DelegatedPermissionEnumSet
+import com.openlattice.authorization.Permission
+import com.openlattice.rhizome.hazelcast.entryprocessors.AbstractReadOnlyRhizomeEntryProcessor
+import java.util.*
+
+class AuthorizationEntryProcessor : AbstractReadOnlyRhizomeEntryProcessor<AceKey, AceValue, DelegatedPermissionEnumSet>() {
```
makosblade, line 10: I am unsure but this may also be `Offloadable`

katherinebernstein: I thought it inherits that from `AbstractReadOnlyRhizomeEntryProcessor`? But I guess it only inherits `ReadOnly` now that I look at it

src/main/java/com/openlattice/authorization/HazelcastAuthorizationService.java
```diff
@@ -440,55 +430,57 @@ public void deletePermissions( AclKey aclKey ) {
...
-        AuthorizationAggregator agg = aces
-                .aggregate( new AuthorizationAggregator(
-                                transformValues( requests, HazelcastAuthorizationService::noAccess ) ),
-                        matches( requests.keySet(), principals ) );
 
-        return agg.getPermissions();
+        Map<AclKey, EnumMap<Permission, Boolean>> permissionMap = Maps.newHashMap( transformValues( requests,
+                HazelcastAuthorizationService::noAccess ) );
+
+        aces.executeOnEntries( new AuthorizationEntryProcessor(), matches( requests.keySet(), principals ) )
```
geekbeast, line 437: It turns out executeOnEntries doesn't properly hold locks (there are hazelcast issues about this and we had issues with id generation as a result).  The correct thing to do here is to generate all AceKey and then only execute on those keys using an Offloadable and ReadOnlyEntryProcessor. This will also avoid executing each authorization queries on all nodes, which is good for scalability.

katherinebernstein: so basically switch to executeOnKeys instead of executeOnEntries? Or are you saying the entry processor itself should be changed in some way? (AuthorizationEntryProcessor is already Offloadable and ReadOnly)

geekbeast: Switch to executeOnKeys instead of executeOnEntries, no changes required to entry processor. The catch is that it will require serialization of a large number of keys instead of just the entry processor. We will want to have before and after timing numbers if possible to see if it hurts performance (i.e we may see a perf hit from serializing the keys, but it may scale better in the long wrong).

katherinebernstein: I made that change :)

katherinebernstein: oh. you merged. lol

PR #839, opened 2020-06-01
Feature/lattice 2503 new crisis report alerts

solomontang:
# Changes
- appends `alertName` metadata to email subject 
- update tags and templates to include `bhr.affiliation` and `bhr.militaryStatus`


src/main/kotlin/com/openlattice/search/renderers/BHRAlertEmailRenderer.kt
```diff
@@ -116,10 +120,11 @@ class BHRAlertEmailRenderer {
                 persistentSearch: PersistentSearch,
                 report: Map<FullQualifiedName, Set<Any>>,
                 userEmail: String,
-                neighbors: List<NeighborEntityDetails>
+                neighbors: List<NeighborEntityDetails>,
+                alertName: String
         ): RenderableEmailRequest {
 
-            val subject = "New Behavioral Health Report"
+            val subject = "New Crisis Report ($alertName)"
```
katherinebernstein, line 127: I think it would be a bit cleaner to save `alertName` on the `alertMetadata` and look it up from there instead of adding to the enum -- this will also make it much easier to add additional BHR alert labels in the future (especially important if they might vary by organization).

i.e. when you're creating the persistent search on the frontend:
```
const persistentSearch = {
    type: 'BHR_ALERT',
    ...,
    alertMetadata: {
        alertName: 'University of Iowa Affiliate' // or Veteran, or whatever
    },
    ...
}
```
and then in `BHRAlertEmailRenderer` at the top under `/* METADATA TAGS */`:
```
const val ALERT_NAME = "alertName"
```
and in `renderEmail`:
```
val alertName = persistentSearch.alertMetadata[ALERT_NAME]?.let { " ($it)" } ?: ""
val subject = "New Crisis Report$alertName"
```

src/main/kotlin/com/openlattice/search/PersistentSearchEmailRenderer.kt
```diff
@@ -26,7 +26,9 @@ class PersistentSearchEmailRenderer {
 
             when (persistentSearch.type) {
                 PersistentSearchNotificationType.ALPR_ALERT -> email = AlprAlertEmailRenderer.renderEmail(persistentSearch, entity, userEmail, neighbors, dependencies.mapboxToken)
-                PersistentSearchNotificationType.BHR_ALERT -> email = BHRAlertEmailRenderer.renderEmail(persistentSearch, entity, userEmail, neighbors)
+                PersistentSearchNotificationType.BHR_ALERT -> email = BHRAlertEmailRenderer.renderEmail(persistentSearch, entity, userEmail, neighbors, "Unsheltered Homeless")
+                PersistentSearchNotificationType.CRISIS_AFFILIATION_ALERT -> email = BHRAlertEmailRenderer.renderEmail(persistentSearch, entity, userEmail, neighbors, "University of Iowa Affiliate")
+                PersistentSearchNotificationType.CRISIS_VETERAN_ALERT -> email = BHRAlertEmailRenderer.renderEmail(persistentSearch, entity, userEmail, neighbors, "Veteran")
```
UnsungHero97, line 31: Please replace all `BHR` with `CARE`.

src/main/kotlin/com/openlattice/search/renderers/BHRAlertEmailRenderer.kt
```diff
@@ -28,6 +28,8 @@ private val INCIDENT_DATE_TIME_FQN = FullQualifiedName("bhr.datetimeOccurred")
 private val NATURE_OF_CRISIS_FQN = FullQualifiedName("bhr.dispatchReason")
 private val HOUSING_SITUATION_FQN = FullQualifiedName("housing.living_arrangements")
 private val DISPOSITION_FQN = FullQualifiedName("bhr.disposition")
+private val AFFILIATION_FQN = FullQualifiedName("bhr.affiliation")
+private val MILITARY_STATUS_FQN = FullQualifiedName("bhr.militaryStatus")
```
UnsungHero97, line 32: Are these part of the v2 data model?

src/main/kotlin/com/openlattice/search/renderers/BHRAlertEmailRenderer.kt
solomontang: `...?.let { " ($it)" } ?: ""`
This is quite the syntax...

src/main/kotlin/com/openlattice/search/renderers/BHRAlertEmailRenderer.kt
solomontang: No. Subscriptions are not available for v2 orgs. Will have to set up a new renderer and update the data model accordingly as only first degree neighbors are available to create the email content.

src/main/kotlin/com/openlattice/search/PersistentSearchEmailRenderer.kt
solomontang: @katherinebernstein has warned me that changing existing notification type enums will break things.

PR #849, opened 2020-06-13
Handle materialization events using an entry listener.

geekbeast:
Cleans up HazelcastAuthorizationService and streamlines how we detect events related to permissions changes.
- Adds functions for bulk initialization of permissions as long as objects are all of the same securable object type.
- Removes signalMaterializedEvents code (this also pushes processing to conductor instead of datastore)
- Creates a kotlin utility function as bridge to full kotlinification of this service.


src/main/kotlin/com/openlattice/authorization/listeners/PermissionMapListener.kt
```diff
@@ -0,0 +1,59 @@
...
+                MaterializePermissionChangeEvent(
+                        event.key.principal,
+                        setOf(event.key.aclKey.first()),
+                        event.value.securableObjectType
+                )
+        )
+    }
+
+    private fun isMaterializationEvent(event: EntryEvent<AceKey, AceValue>): Boolean {
+        return event.value.contains(Permission.MATERIALIZE) && event.key.principal.type == PrincipalType.ORGANIZATION && (event.value.securableObjectType == SecurableObjectType.EntitySet || event.value.securableObjectType == SecurableObjectType.PropertyTypeInEntitySet)
```
makosblade, line 52: Can you break this into multiple lines plz

src/main/kotlin/com/openlattice/datastore/services/EntitySetService.kt
```diff
@@ -134,14 +137,14 @@ open class EntitySetService(
             )
 
             entityType.properties
+                    .parallelStream()
                     .map { propertyTypeId -> AclKey(entitySetId, propertyTypeId) }
-                    .onEach { aclKey ->
-                        authorizations.setSecurableObjectType(aclKey, SecurableObjectType.PropertyTypeInEntitySet)
-                    }
                     .forEach { aclKey ->
+                        authorizations.setSecurableObjectType(aclKey, PropertyTypeInEntitySet)
```
katherinebernstein, line 143: these should be using the new bulk setSecurableObjectTypes + addPermissions calls, yeah?

src/main/java/com/openlattice/authorization/AuthorizationManager.java
```diff
@@ -31,6 +31,7 @@
 import java.time.OffsetDateTime;
 import java.util.*;
 import java.util.stream.Stream;
+import org.checkerframework.checker.units.qual.Time;
```
UnsungHero97, line 34: Is this a standard dependency? What happened to JodaTime?

geekbeast: Don't know how that got in there.

geekbeast:
> Tests?

SerializationTests are automatic now and the other code changes is simply changing from one key to multi-key execution. For coverage, I added a test uses the new APIs on multiple aclkeys.

I was on the fence about unit testing the map listener, but the only thing we'd be testing is that the isMaterializationLogic is the isMaterializationLogic.

PR #855, opened 2020-06-13
Fix/lil things in life

makosblade:
This PR:
- removes some dead code
- uses enum set serializer to switch from O(4Bytes*#permissions in set) to constant 6 bits
- fixes an NPE in an entryprocessor


src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -460,78 +460,6 @@ class HazelcastOrganizationService(
 
     }
 
-    private fun fixOrganizations() {
```
katherinebernstein, line 463: 😍

geekbeast: Wait until we have to fix organizations again ;)

src/main/kotlin/com/openlattice/edm/processors/GetPartitionsFromEntitySetEntryProcessor.kt
```diff
@@ -5,7 +5,10 @@ import com.openlattice.rhizome.hazelcast.entryprocessors.AbstractReadOnlyRhizome
 import java.util.*
 
 class GetPartitionsFromEntitySetEntryProcessor : AbstractReadOnlyRhizomeEntryProcessor<UUID, EntitySet, Set<Int>>() {
-    override fun process(entry: MutableMap.MutableEntry<UUID, EntitySet>): Set<Int> {
-        return entry.value.partitions
+    override fun process(entry: MutableMap.MutableEntry<UUID, EntitySet?>): Set<Int> {
+        if ( entry.value == null ){
+            return setOf()
+        }
+        return entry.value!!.partitions
```
UnsungHero97, line 12: I don't know Kotlin; just curious. Is the `!!` since you already check for `null` above?

makosblade: Ah, good catch. I just needed to assign entry.value to a variable and then its able to do a smart cast with it on line 12.
I was a little surprised that it can't detect this current code as not needing !!

PR #857, opened 2020-06-19
Bugfix/unexpected null principal

geekbeast:


geekbeast:
Closing as making fields protected was a better fix.

PR #859, opened 2020-06-22
Partition fixes, wrapper type fixes

makosblade:
This PR:
- adds wrapper types around Set<Int> usage in entryprocessor


src/main/kotlin/com/openlattice/organizations/tasks/OrganizationsInitializationTask.kt
```diff
@@ -50,8 +50,9 @@ class OrganizationsInitializationTask : HazelcastInitializationTask<Organization
         logger.info("Running bootstrap process for organizations.")
         val sw = Stopwatch.createStarted()
         val organizationService = dependencies.organizationService
+        val partitionManager = dependencies.partitionManager
         val globalOrg = organizationService.maybeGetOrganization(GLOBAL_ORG_PRINCIPAL)
-        val defaultPartitions = organizationService.allocateDefaultPartitions(organizationService.numberOfPartitions)
+        val defaultPartitions = partitionManager.getAllPartitions()
```
geekbeast, line 55: Are you making it so that global organization entity sets are by default spread across all partitions?

src/main/kotlin/com/openlattice/data/storage/partitions/PartitionManager.kt
```diff
@@ -199,6 +145,11 @@ class PartitionManager @JvmOverloads constructor(
 
 }
 
-
-private val ALL_PARTITIONS = "SELECT ${PARTITION.name}, COALESCE(count,0) as $COUNT FROM (SELECT unnest(?::integer[]) as ${PARTITION.name}) as partitions LEFT JOIN ${PARTITION_COUNTS.name} USING (${PARTITION.name}) ORDER BY $COUNT ASC "
-private val EMPTIEST_PARTITIONS = "$ALL_PARTITIONS LIMIT ?"
+private val EMPTIEST_PARTITIONS = """
```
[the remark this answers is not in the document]
geekbeast: Also on second look this doesn't seem to compute the emptiest partitions. It also seems to filter empty entity sets from the join.

src/main/kotlin/com/openlattice/data/storage/partitions/PartitionManager.kt
```diff
@@ -160,29 +120,15 @@ class PartitionManager @JvmOverloads constructor(
...
-    }
-
-    private fun isValidResize(partitionCount: Int) {
-        checkArgument(partitionCount > partitionList.size, "Only support resizing to a larger number of partitions.")
+    private fun getEmptiestPartitions(numPartitions: Int): List<Int> {
+        val partitionCounts = mutableMapOf<Int, Long>()
+        BasePostgresIterable(StatementHolderSupplier(hds, EMPTIEST_PARTITIONS)) {
+            ResultSetAdapters.count(it) to ResultSetAdapters.partitions(it)
+        }.forEach { (count, partitions) ->
+            val avg = count / partitions.size
```
geekbeast, line 128: This seems to be computing emptiest partitions from a computation that assumes data is evenly spread across partitions for entity sets. We know the actual sizes from PostgresMaterializedViews for partition counts, why aren't we using that?

src/main/kotlin/com/openlattice/organizations/tasks/OrganizationsInitializationTask.kt
makosblade: Yes but this code should be equivalent to the previous call, just a bit clearer IMO

src/main/kotlin/com/openlattice/data/storage/partitions/PartitionManager.kt
makosblade: It appears as though the previous get emptiest partitions call isn't actually working and creates a non-trivial performance impact whenever the counts get updated. 

I think assuming the data is evenly distributed should be fine since ekid -> partition mapping is random. Also I think this implementation should work about as well and end up reducing DB load.

The idea also is to add repartitioning to fix the ton of badly partitioned data 😢

makosblade: (repartitioning is l8r)

src/main/kotlin/com/openlattice/data/storage/partitions/PartitionManager.kt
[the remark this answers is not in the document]
makosblade: Yeah can definitely rename this for accuracy

PR #862, opened 2020-06-24
Delete AuthorizationQueryService

katherinebernstein:


katherinebernstein:
> I assume that we will implement paging again at some point in the future?

We can, but I don't think it's being used at all right now

PR #863, opened 2020-06-25
Remove unneeded

geekbeast:
Should get rid of the annoying SQL error at startup.


geekbeast:
This is safe to remove since we don't create user accounts in internal db anymore, only external db.

PR #866, opened 2020-06-25
composite function for managing cluster locking

makosblade:


src/main/kotlin/com/openlattice/linking/LinkingQueryService.kt
```diff
@@ -46,6 +47,23 @@ interface LinkingQueryService {
             clusterId: UUID,
             scores: Map<EntityDataKey, Map<EntityDataKey, Double>>): Int
 
+    /**
+     * Performs the following operataions:
+     *
```
anzioka, line 52: typo

PR #869, opened 2020-06-26
Move HazelcastAuthorizationService to kotlin

katherinebernstein:


src/main/kotlin/com/openlattice/authorization/HazelcastAuthorizationService.kt
```diff
@@ -0,0 +1,526 @@
...
+
+
+class HazelcastAuthorizationService(
+        hazelcastInstance: HazelcastInstance,
+        eventBus: EventBus
+) : AuthorizationManager {
+
+    private var securableObjectTypes: IMap<AclKey, SecurableObjectType> = HazelcastMap.SECURABLE_OBJECT_TYPES.getMap(hazelcastInstance)
+    private var aces: IMap<AceKey, AceValue> = HazelcastMap.PERMISSIONS.getMap(hazelcastInstance)
+    private var eventBus: EventBus = checkNotNull(eventBus)
```
makosblade, line 43: I don't think you need the checkNotNull anymore b/c eventbus isn't nullable. Probably also want to make eventBus constructor arg a val and make the IMaps vals as well

katherinebernstein: ➕

PR #872, opened 2020-07-01
Bugfix/reduce entity key id locking

geekbeast:
This reduces hazelcast pressure doing integrations, fixes uuid list serializaton during id generation, improves local queue reservoir logic, and makes more ids available for generation.

This will now leak ~325K ids every time service is restarted, so at some point we should build some service that looks for gaps in the sequence, but not a priority at the moment.

### DIFF
--- a/src/main/java/com/openlattice/ids/IdsGeneratingEntryProcessor.java
+++ b/src/main/java/com/openlattice/ids/IdsGeneratingEntryProcessor.java
@@ -23,6 +23,7 @@ package com.openlattice.ids;
 import com.google.common.annotations.VisibleForTesting;
 import com.hazelcast.core.Offloadable;
 import com.kryptnostic.rhizome.hazelcast.processors.AbstractRhizomeEntryProcessor;
+import com.openlattice.rhizome.hazelcast.DelegatedUUIDList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map.Entry;
@@ -44,7 +45,7 @@ public class IdsGeneratingEntryProcessor extends AbstractRhizomeEntryProcessor<I
         final Range range = entry.getValue(); //Range should never be null in the EP.
         final UUID[] ids = getIds( range );
         entry.setValue( range );
-        return Arrays.asList( ids );
+        return DelegatedUUIDList.wrap( Arrays.asList( ids ) );
     }
 
     public int getCount() {
--- a/src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
+++ b/src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
@@ -23,6 +23,7 @@ package com.openlattice.data.ids
 
 import com.geekbeast.hazelcast.HazelcastClientProvider
 import com.google.common.base.Preconditions.checkState
+import com.google.common.collect.Queues
 import com.google.common.util.concurrent.ListeningExecutorService
 import com.openlattice.IdConstants
 import com.openlattice.data.EntityKey
@@ -39,13 +40,13 @@ import com.openlattice.postgres.PostgresTable.*
 import com.openlattice.postgres.ResultSetAdapters
 import com.openlattice.postgres.streams.BasePostgresIterable
 import com.openlattice.postgres.streams.PreparedStatementHolderSupplier
-import com.openlattice.postgres.streams.StatementHolder
 import com.zaxxer.hikari.HikariDataSource
 import org.apache.olingo.commons.api.edm.EdmPrimitiveTypeKind
 import org.slf4j.LoggerFactory
 import java.sql.Connection
+import java.sql.PreparedStatement
 import java.util.*
-import java.util.function.Supplier
+import java.util.concurrent.BlockingQueue
 import kotlin.collections.HashMap
 
 /**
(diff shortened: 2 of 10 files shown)


src/main/java/com/openlattice/ids/IdsGeneratingEntryProcessor.java
```diff
@@ -44,7 +45,7 @@ public IdsGeneratingEntryProcessor( int count ) {
         final Range range = entry.getValue(); //Range should never be null in the EP.
         final UUID[] ids = getIds( range );
         entry.setValue( range );
-        return Arrays.asList( ids );
+        return DelegatedUUIDList.wrap( Arrays.asList( ids ) );
```
katherinebernstein, line 48: Does it make any difference whether the function (/entry processor) returns a List<UUID> or DelegatedUUIDList?

src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
```diff
@@ -57,193 +58,160 @@ private val entityKeyIdsSql = "SELECT * FROM ${SYNC_IDS.name} WHERE ${ENTITY_SET
...
         checkState(ids.size == entityIds.size, "Insufficient ids generated.")
         return entityIds.zip(ids).toMap()
     }
 
     private fun storeEntityKeyIdReservations(
-            entitySetId: UUID, entityKeyIds: Set<UUID>,
-            partitions: Set<Int> = partitionManager.getEntitySetPartitions(entitySetId)
+            entitySetId: UUID,
+            entityKeyIds: Set<UUID>,
+            partitions: Array<Int> = partitionManager.getEntitySetPartitions(entitySetId).toTypedArray()
```
katherinebernstein, line 88: Did we determine that accessing partitionManager in the default value function is safe? (I vaguely remember problems with this in the distant and recent past but don't remember what the resolution was)

src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
```diff
@@ -57,193 +58,160 @@ private val entityKeyIdsSql = "SELECT * FROM ${SYNC_IDS.name} WHERE ${ENTITY_SET
...
+
+        insertIds.setObject(1, entitySetId)
+        insertIds.setObject(2, entityKeyId)
+        insertIds.setInt(3, partition)
+        insertIds.addBatch()
+
+        insertToData.setObject(1, entitySetId)
+        insertToData.setObject(2, entityKeyId)
+        insertToData.setInt(3, partition)
+        insertToData.setObject(4, IdConstants.ID_ID.id)
```
katherinebernstein, line 188: (lol)

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -694,6 +694,10 @@ fun getPartition(entityKeyId: UUID, partitions: List<Int>): Int {
     return partitions[entityKeyId.leastSignificantBits.toInt() % partitions.size]
 }
 
+fun getPartition(entityKeyId: UUID, partitions: Array<Int>): Int {
```
katherinebernstein, line 697: Doesn't something like this already exist in PostgresDataQueries?

src/main/java/com/openlattice/ids/IdsGeneratingEntryProcessor.java
geekbeast: Yes, because just returning a list means it uses java serialization, but wrapping it users stream serializer instead.

katherinebernstein: (I ask because the return type of the function is still List<UUID>)

src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
geekbeast: it's safe as long field the service isn't annotated with ```@Timed``` or similar annotations that trigger spring proxying the class. Even then it's fine as long as member variable is declared protected (or public).

src/main/java/com/openlattice/ids/IdsGeneratingEntryProcessor.java
katherinebernstein: `List<UUID>`

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
geekbeast: Yes, but it takes a list instead of a typed array. I guess could have routed it to list version by doing .asList

src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
katherinebernstein: Also, what's the reason to convert `partitions` to an array?

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
katherinebernstein: oh oops I didn't see this, but why is it being converted to a typed array at all?

src/main/java/com/openlattice/ids/IdsGeneratingEntryProcessor.java
geekbeast: Hazelcast uses the class object not the type specified on the interface. Otherwise, it would have to use reflection to unreliably reify types on entry processors.

src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
geekbeast: I thought it was also being bound as a jdbc parameter, so I was trying to avoid unnecessary copy.  By the time I realized it was array everywhere. It's perfectly reason to switch to List<Int> instead to avoid having to versions of the function since no one is likely to save time in the way I originally imagined since this is used for look up only and never binding.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
geekbeast: I had to do a conversion from set and I could gone to list or array.  I did array for reason mentioned above.

src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
makosblade: Any chance we can use IntArray instead to keep it primitive?

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
makosblade: I feel like this and the above `getPartition` should both be in `PartitionManager` instead of here in `PostgresDataQueries` but I don't think there being two methods is an issue

src/main/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdService.kt
geekbeast: Sure

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
geekbeast: Done

PR #875, opened 2020-07-06
Lock entity by entity in order.

geekbeast:


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -354,8 +354,8 @@ fun buildUpsertEntitiesAndLinkedData(): String {
  */
 val lockEntitiesInIdsTable =
         "SELECT 1 FROM ${IDS.name} " +
-                "WHERE ${ENTITY_SET_ID.name} = ? AND ${ID_VALUE.name} = ANY(?) AND ${PARTITION.name} = ? " +
-                "ORDER BY ${ID_VALUE.name} " +
+                "WHERE ${ENTITY_SET_ID.name} = ? AND ${ID_VALUE.name} = ? AND ${PARTITION.name} = ? " +
+//                "ORDER BY ${ID_VALUE.name} " +
```
katherinebernstein, line 358: is this supposed to be commented out?

katherinebernstein: oh i see

geekbeast: Yes, but I should actually just remove the comment.

katherinebernstein: How will the case where there are two batch writes at the same time writing to overlapping sets of entities work? (ie if write1 locks EKID1, write2 locks EKID2, write1 tries to lock EKID2, write2 tries to lock EKID1)

katherinebernstein: ah i guess the sorted() should take care of this

PR #879, opened 2020-07-09
Feature/kotlify and update hazelcast principal service

katherinebernstein:
This ended up being much bigger in scope than I originally intended, but the main things this does are:
- Convert HazelcastPrincipalService to kotlin
- Update Auth0SyncTask to use bulk calls where possible
- Avoid unnecessary work of adding members to organizations they already belong to where possible
- Process organization connection grants on the creation of a grant or the addition of a user, instead of on every sync task


src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -75,7 +75,11 @@ class HazelcastOrganizationService(
 
     @Timed
     fun maybeGetOrganization(p: Principal): Optional<SecurablePrincipal> {
-        return securePrincipalsManager.maybeGetSecurablePrincipal(p)
+        return try {
```
makosblade, line 78: this feels very close to what `Optional.ofNullable()` does. Definitely prefer checking null over try-catching an exception

src/main/kotlin/com/openlattice/principals/PrincipalExistsEntryProcessor.kt
```diff
@@ -0,0 +1,12 @@
+package com.openlattice.principals
+
+import com.openlattice.authorization.AclKey
+import com.openlattice.authorization.SecurablePrincipal
+import com.openlattice.rhizome.hazelcast.entryprocessors.AbstractReadOnlyRhizomeEntryProcessor
+
+class PrincipalExistsEntryProcessor :  AbstractReadOnlyRhizomeEntryProcessor<AclKey, SecurablePrincipal?, Boolean>() {
```
makosblade, line 7: Offloadable?

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
katherinebernstein: I agree, although `spm.getPrincipal` is somewhat heavily used outside of this PR and I wasn't sure if any of those calls are specifically expecting an exception to be thrown in the null case. I can check/change those though

src/main/kotlin/com/openlattice/users/Auth0SyncService.kt
```diff
@@ -79,25 +96,28 @@ class Auth0SyncService(
...
 
-        processGlobalEnrollments(principal, user)
+        processGlobalEnrollments(mapOf(principal to user))
         processOrganizationEnrollments(principal, user)
 
         syncAuthenticationCache(principal.id)
-        markUser(user.id)
     }
 
-    private fun markUser(userId: String) {
```
geekbeast, line 89: Are we assuming the users will only be deleted via API and not through Auth0?

src/main/kotlin/com/openlattice/users/Auth0SyncService.kt
```diff
@@ -64,6 +53,34 @@ class Auth0SyncService(
         syncUserEnrollmentsAndAuthentication(user)
     }
 
+    fun createOrUpdateUsers(usersToUpdate: Collection<User>) {
```
geekbeast, line 56: Something I realized when looking at this fresh is that users need to be added as members of an organization in the organization and the principal service. Need to ensure this is fail safe.

src/main/kotlin/com/openlattice/users/Auth0SyncService.kt
```diff
@@ -64,6 +53,34 @@ class Auth0SyncService(
...
+
+        logger.debug("Creating users {} and updating principals {}",
+                principalsToCreate.map { it.id },
+                existingUserPrincipals.map { it.id }
+        )
+
+        val newlyCreatedUsers = principalsToCreate.filter {
+            val wasSuccessfullyCreated = tryCreateNewUserPrincipal(usersByPrincipal.getValue(it), it)
+
+            if (!wasSuccessfullyCreated) {
```
geekbeast, line 71: This doesn't seem fail safe. If datastore is killed before the body of the if executes, will it be left in an inconsistent state?

katherinebernstein: It would, and I have to double check but I'm pretty sure the auth0 initialization task still runs through every user, so every fresh restart should be a hard re-sync

geekbeast: :+1:

src/main/kotlin/com/openlattice/users/Auth0SyncService.kt
katherinebernstein: I think so, although it looks like the expiration field is never actually referenced anywhere other than being updated in `markUser`

PR #880, opened 2020-07-11
Bugfix/lock ids wherever rows are updated

katherinebernstein:


src/main/kotlin/com/openlattice/postgres/RetryableLockedIdsOperator.kt
```diff
@@ -0,0 +1,55 @@
...
+
+        private const val MAX_INTERVAL: Long = 60 * 1_000
+        private const val RETRY_OFFSET: Long = 125
+        private const val MAX_RETRY_ATTEMPTS = 32
+
+        fun lockIdsAndExecute(
+                connection: Connection,
+                entitySetIdToEntityKeyId: Map<EntityDataKey, Int>,
+                operateOnConnectionFn: (Connection) -> Int
+        ): Int {
```
geekbeast, line 22: You can make return type generic and make it match the function that is passed in.

src/main/kotlin/com/openlattice/postgres/RetryableLockedIdsOperator.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    companion object {
+
+        private const val MAX_INTERVAL: Long = 60 * 1_000
+        private const val RETRY_OFFSET: Long = 125
+        private const val MAX_RETRY_ATTEMPTS = 32
+
+        fun lockIdsAndExecute(
+                connection: Connection,
+                entitySetIdToEntityKeyId: Map<EntityDataKey, Int>,
```
geekbeast, line 20: Set<EntityDataKey> is probably fine here as well.

src/main/kotlin/com/openlattice/postgres/RetryableLockedIdsOperator.kt
```diff
@@ -0,0 +1,55 @@
...
+                operateOnConnectionFn: (Connection) -> Int
+        ): Int {
+
+            return attempt(LinearBackoff(MAX_INTERVAL, RETRY_OFFSET), MAX_RETRY_ATTEMPTS) {
+
+                connection.use { conn ->
+
+                    conn.autoCommit = false
```
geekbeast, line 28: It's a little dangerous to change autoCommit and make commit an externally passed in function. We're the only ones using it, but an alternate approach here might be the use a CTE with SELECT ... ORDER BY ... FOR  UPDATE follow by the original SQL query.

src/main/kotlin/com/openlattice/postgres/RetryableLockedIdsOperator.kt
```diff
@@ -0,0 +1,55 @@
...
+        private const val RETRY_OFFSET: Long = 125
+        private const val MAX_RETRY_ATTEMPTS = 32
+
+        fun lockIdsAndExecute(
+                connection: Connection,
+                entitySetIdToEntityKeyId: Map<EntityDataKey, Int>,
+                operateOnConnectionFn: (Connection) -> Int
+        ): Int {
+
+            return attempt(LinearBackoff(MAX_INTERVAL, RETRY_OFFSET), MAX_RETRY_ATTEMPTS) {
```
geekbeast, line 24: We should remove, because deadlocks should be impossible.

src/main/kotlin/com/openlattice/postgres/RetryableLockedIdsOperator.kt
katherinebernstein: Set?

geekbeast: Never mind ignore this one.

src/main/kotlin/com/openlattice/postgres/LockedIdsOperator.kt
```diff
@@ -0,0 +1,89 @@
+package com.openlattice.postgres
+
+import com.openlattice.postgres.PostgresColumn.*
+import com.openlattice.postgres.PostgresTable.IDS
+import java.sql.Connection
+import java.sql.PreparedStatement
+import java.util.*
+
+class LockedIdsOperator {
```
geekbeast, line 9: Why wrap this in a class?

src/main/kotlin/com/openlattice/postgres/LockedIdsOperator.kt
```diff
@@ -0,0 +1,89 @@
...
+         * the current bind index).
+         *
+         * @return The number of rows that were udpated.
+         *
+         */
+        fun lockIdsAndExecute(
+                connection: Connection,
+                query: String,
+                entitySetId: UUID,
+                entityKeyIdsByPartition: Map<Int, Collection<UUID>>,
```
geekbeast, line 40: Seem like a good candidate for a helper function somewhere.

makosblade:
Okay actually looking at this further, the lockIdsAndUpdate makes it more maintainable to do `index++` so I don't care manual or not in this case

PR #881, opened 2020-07-11
Bugfix/single transaction writes no dupe reads

geekbeast:
This adds locking to clear entities and switches to a more efficient way of reading linked entities.

This will remove the need for us to do an extra write for linked entities after each write and should have a pretty big positive performance impact. 

We will need to run a migration to set ORIGIN_ID to LINKING_ID (and it's probably worth looking at renaming the column). The migration should be fairly efficient since we do equality on the partitions column allowing the query to be pushed down to all nodes.

@makosblade This will require modifying linking queries to just do updates instead of inserts (should be way more efficient).

### DIFF
--- a/src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
+++ b/src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
@@ -427,7 +427,7 @@ class PostgresEntityDataQueryService(
             val entityKeyIdsArr = PostgresArrays.createUuidArray(connection, entities.keys)
             val lockEntities = connection.prepareStatement(lockEntitiesInIdsTable)
             //Make data visible by marking new version in ids table.
-            val upsertEntities = connection.prepareStatement(buildUpsertEntitiesAndLinkedData())
+            val upsertEntities = connection.prepareStatement(upsertEntitiesSql)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -973,31 +983,39 @@ class PostgresEntityDataQueryService(
...
+            val entityKeyIdsArr = PostgresArrays.createUuidArray(conn, entityKeyIds)
+            val partitionsArr = PostgresArrays.createIntArray(conn, partitionsMap.keys)
+
+            val numUpdated = conn.prepareStatement(updateVersionsForEntitiesInEntitySet).use { ps ->
+                ps.setLong(1, -version)
+                ps.setLong(2, -version)
+                ps.setLong(3, -version)
+                ps.setObject(4, entitySetId)
+                ps.setArray(5, entityKeyIdsArr)
+                ps.setArray(6, partitionsArr)
```
katherinebernstein, line 1013: just a note that I think my PR changed the bind order of these last 2 params, so that's something to look out for on merging these

geekbeast: :+1:

PR #892, opened 2020-07-20
Bugfix/sync job to pull in any missing twilio messages

katherinebernstein:


src/main/java/com/openlattice/postgres/PostgresColumn.java
```diff
@@ -20,22 +20,11 @@
...
-import static com.openlattice.postgres.PostgresDatatype.TEXT;
-import static com.openlattice.postgres.PostgresDatatype.TEXT_ARRAY;
-import static com.openlattice.postgres.PostgresDatatype.TIMESTAMPTZ;
-import static com.openlattice.postgres.PostgresDatatype.UUID;
-import static com.openlattice.postgres.PostgresDatatype.UUID_ARRAY;
-
 import com.openlattice.IdConstants;
 import com.openlattice.edm.type.Analyzer;
 
+import static com.openlattice.postgres.PostgresDatatype.*;
```
makosblade, line 26: 🚫 
`import static .*` considered harmful. Unless this has changed in recent JVMs, I believe this will import every static anything in the class, even if unused. Admittedly PostgresDatatype is probably the only type of class where I might accept this but even so, personally I'd pref just listing everything used explicitly.

katherinebernstein: 👍

src/main/java/com/openlattice/postgres/PostgresColumn.java
```diff
@@ -320,6 +320,12 @@
             .withDefault( "'-infinity'" )
             .notNull();
     public static final String                   LAST_WRITE_FIELD                  = "last_write";
+    public static final String                   LAST_SYNC_FIELD                   = "last_sync";
+    public static final PostgresColumnDefinition LAST_SYNC                         = new PostgresColumnDefinition(
+            LAST_SYNC_FIELD,
+            TIMESTAMPTZ )
+            .withDefault( "'-infinity'" )
```
UnsungHero97, line 327: `'-infinity'` seems like a good candidate for a constant somewhere.

katherinebernstein: Made a ticket for this as there's a few of these that should be constantified, and it'll involve finding/updating more files than just this one https://jira.openlattice.com/browse/LATTICE-2531

geekbeast:
@katherinebernstein Looking at the failing tests it looks like you may have forgotten to implement equals for the entry processor, which is causing the serialization test to fail.

PR #894, opened 2020-07-22
Bugfix/batch all the writes

geekbeast:


src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
```diff
@@ -71,20 +64,16 @@ class IndexingMetadataManager(private val hds: HikariDataSource, private val par
...
-                                updateLastLinkingIndexSql,
-                                entitySetId,
-                                partition,
-                                shouldLockEntireEntitySet = true
-                        ) { ps ->
+        return hds.connection.use { connection ->
+            val ps = connection.prepareStatement(updateLastLinkingIndexSql)
+            linkingIdsWithLastWrite.map { (entitySetId, entities) ->
+                val partitions = entitySetPartitions.getValue(entitySetId)
+                entities.entries
```
katherinebernstein, line 71: Can we pull the grouping/sorting logic outside of the connection? (Just to reduce the amount of work that needs to be done while it's been reserved)

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
```diff
@@ -247,6 +215,16 @@ private val entityKeyIdsInEntitySet =
         "AND ${ID.name} = ANY(?) " +
         "AND ${PARTITION.name} = ? "
 
+/**
+ * 1. entity set id
+ * 2. entity key ids (uuid array)
```
katherinebernstein, line 220: should be `entity key id`

src/main/kotlin/com/openlattice/data/storage/PostgresEntityDataQueryService.kt
```diff
@@ -417,42 +417,45 @@ class PostgresEntityDataQueryService(
...
+                ps.setObject(5, entityKeyId)
+                ps.setInt(6, partition)
+                ps.addBatch()
+            }
 
-            connection.autoCommit = true
-            logger.debug("Updated $updatedLinkedEntities linked entities as part of insert.")
+            val updatedEntities = ps.executeBatch().sum()
+
+//            val updatedLinkedEntities = ps.executeBatch().sum()
```
katherinebernstein, line 435: any reason to keep this here+commented instead of deleting it?

src/main/kotlin/com/openlattice/data/storage/IndexingMetadataManager.kt
geekbeast: That's what it was before. I pulled it out to avoid re-creating the prepared statement over and over in the loop. I could do something like building out all the groupings before executing the queries, but that's a bigger change.

katherinebernstein: can't we just assign the sorted map to a variable before acquiring the connection?

katherinebernstein: (then acquire the connection, prepare the statement, iterate over the sorted map)

geekbeast: Yeah that would work, but you still have this: https://jdbc.postgresql.org/documentation/head/server-prepare.html

So to get the full benefit of both you have to prepare the full data structure ahead of time.

geekbeast: LATTICE-2530

PR #898, opened 2020-07-29
Feature/repartitioning

geekbeast:
This escalated quickly so merging into develop, before I keep going.


src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -1186,6 +1190,10 @@ private val EDGE_IDS_SQL = "${EDGE_ENTITY_KEY_ID.name} = ANY(?) AND ${EDGE_ENTIT
 private val SRC_IDS_AND_ENTITY_SETS_SQL = "${SRC_ENTITY_KEY_ID.name} = ANY(?) AND ${SRC_ENTITY_SET_ID.name} = ANY(?)"
 private val DST_IDS_AND_ENTITY_SETS_SQL = "${DST_ENTITY_KEY_ID.name} = ANY(?) AND ${DST_ENTITY_SET_ID.name} = ANY(?)"
 
+const val REPARTITION_SELECTOR = "partitions[ 1 + (('x'||right(id::text,8))::bit(32)::int % array_length(partitions,1))]"
```
katherinebernstein, line 1193: if I remember right, there needs to be an `abs()` somewhere in here for when ids end up wrapping (/any leftover ones we had from when id generation was just random uuid)

src/main/kotlin/com/openlattice/graph/Graph.kt
```diff
@@ -1186,6 +1190,10 @@ private val EDGE_IDS_SQL = "${EDGE_ENTITY_KEY_ID.name} = ANY(?) AND ${EDGE_ENTIT
...
+private val REPARTITION_COLUMNS = E.columns.joinToString(",") { if (it == PARTITION) REPARTITION_SELECTOR else it.name }
+private val REPARTITION_SQL = "INSERT INTO ${E.name} SELECT $REPARTITION_COLUMNS FROM ${E.name} INNER JOIN"
```
katherinebernstein, line 1195: this is incomplete, yeah?

src/main/kotlin/com/openlattice/graph/Graph.kt
geekbeast: We shouldn't have any, but I have a fix for this issue which will match what the back-end does. This code merge is more fore hz-4.0 and I'm going to keep working on this branch.

src/main/kotlin/com/openlattice/graph/Graph.kt
geekbeast: correct. this is just to merge for hz-4.0

PR #901, opened 2020-07-30
Feature/repartitioning

geekbeast:
Wires up admin jobs API and re-partitioning API for entity sets. This is just the bare minimum so that we can start and monitor re-partitioning jobs. 

Some work items still needed:

- Implement resumable jobs, so if stack goes down jobs can be restarted.
- Expose API to clear job status.


src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,271 @@
...
+import java.util.*
+
+/**
+ * Background job for re-partitioning data in an entity set.
+ *
+ * @author Matthew Tamayo-Rios &lt;matthew@openlattice.com&gt;
+ */
+class RepartitioningJob
+@JsonCreator constructor(
+        jobState: RepartitioningJobState
```
makosblade, line 24: Can you rename this to state to match `state` in `AbstractDistributedJob` ?

src/main/kotlin/com/openlattice/data/DataGraphManager.kt
```diff
@@ -180,14 +180,16 @@ interface DataGraphManager {
...
+            expirationPolicy: DataExpiration,
+            dateTime: OffsetDateTime,
+            deleteType: DeleteType,
+            expirationPropertyType: Optional<PropertyType>
     ): BasePostgresIterable<UUID>
 
     fun getEdgeEntitySetsConnectedToEntities(entitySetId: UUID, entityKeyIds: Set<UUID>): Set<UUID>
     fun getEdgeEntitySetsConnectedToEntitySet(entitySetId: UUID): Set<UUID>
-    fun setPartitions(entitySetId: UUID, partitions: Set<Int>) : Int
+    fun setPartitions(entitySetId: UUID, oldPartitions: Set<Int>): UUID
```
makosblade, line 194: Can you rename this to `repartitionEntitySet` or something like that? I don't think it's super clear here that it will create a job that moves all the data around.

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,271 @@
...
+        whenExcludedGreater: PostgresColumnDefinition = column,
+        otherwise: PostgresColumnDefinition = column
+): String = "${column.name} = CASE WHEN EXCLUDED.${comparison.name} > ${comparison.name} THEN EXCLUDED.${whenExcludedGreater.name} ELSE ${otherwise.name} END"
+
+
+private val REPARTITION_DATA_COLUMNS = buildRepartitionColumns(DATA)
+private val REPARTITION_IDS_COLUMNS = buildRepartitionColumns(E)
+private val REPARTITION_EDGES_COLUMNS = buildRepartitionColumns(IDS)
+
+/**
```
makosblade, line 175: Won't this take an eternity without chunking?

geekbeast: Citus 9.3 has special re-partitioning logic where the nodes talk to each other. I want to try and take advantage of that first, before we start doing manual batches.

makosblade: So is the idea to test this out on a smaller/medium sized entity set before NCRIC? I'm guessing 9.3 avoids the coordinator memory limit right?

geekbeast: The coordinator is not involved in re-partitioning queries if LIMIT, ORDER BY, or GROUP BY clauses are omitted. Yes, the goal would be to try this out on smaller entity sets.

src/test/resources/rhizome.yaml
```diff
@@ -15,9 +15,12 @@ hazelcast:
...
+    initialize-indices: true
     hikari:
-        jdbcUrl: jdbc:postgresql://localhost:5432/openlattice
+        jdbcUrl: jdbc:postgresql://localhost:5433/openlattice
         username: "oltest"
         password: "test"
-        maximumPoolSize: 10
-        connectionTimeout: 12000000 #20 minutes
\ No newline at end of file
+        maximumPoolSize: 32
```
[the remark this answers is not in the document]
geekbeast: This is my local configuration for running citus that leaked. I'll set it back.

geekbeast:
> I feel like this is unlikely to work on entity sets that are large -- if the CTE loads all ids/data/edges from the entity set and then has to ship that data to other boxes, it seems likely to hit that max CTE size error.
> 
> As a followup thought -- if gross margins is at all close to done, I feel like that should take priority over this. We definitely do still need to repartition ids and edges, but if we can avoid an extra read+write+delete of all the customer data, I feel like that could potentially save days of time / days of potential backend badness that might (but hopefully not) come up, and the majority of the warp core data is in the data table (so gross margins would free up a TON of space).

This doesn't use CTEs, it uses a new Citus re-partitioning mechanism. It's a good point what we can save some I/O, but I'm not sure on the timeline for gross margins branch and we're getting way more pressure on materialized views. This is a much simpler change that will dramatically speed up integrations on single partition entity sets that is pretty much ready to go.

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
katherinebernstein: This is probably a good initial set to test repartitioning on! Most of them (save two) are suuuper small https://jira.openlattice.com/browse/LATTICE-2412?focusedCommentId=23853&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23853

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,315 @@
...
+) : AbstractDistributedJob<Long, RepartitioningJobState>(state), MetastoreAware {
+    constructor(
+            entitySetId: UUID,
+            oldPartitions: List<Int>,
+            newPartitions: Set<Int>
+    ) : this(RepartitioningJobState(entitySetId, oldPartitions,newPartitions))
+    private var phase = 0
+
+    @Transient
+    private lateinit var hds: HikariDataSource
```
makosblade, line 38: Does Kotlin not have a runtime exception here b/c there's no init block?

geekbeast: Why would there be a run time exception? On the client-side it's not an issue unless someone invokes call() without calling setHikariDataSource. Otherwise, the stream serializer initializes the value before it's ever used.

makosblade: Oh derp, I hadn't caught the full context for it. I was pretty sure that lateinit variables cause runtime exceptions without init blocks but it looks like it would be thrown at read time 😅

src/main/java/com/openlattice/postgres/PostgresTable.java
```diff
@@ -31,7 +32,7 @@
  * Tables definitions for all tables used in the OpenLattice platform.
  */
 public final class PostgresTable {
-
+    public static final PostgresTableDefinition JOBS = PostgresJobsMapStore.JOBS;
```
makosblade, line 35: What's this here for? Can't any consumer of `PostgresTable` consume `PostgresJobsMapStore` instead? Not a big deal in any case, just curious

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+        ps.setObject(1, state.entitySetId)
+        ps.setArray(2, PostgresArrays.createIntArray(ps.connection, state.newPartitions))
+        ps.setInt(3, partition)
+    }
+
+    private fun setPartitions(entitySetId: UUID, partitions: Set<Int>) {
+        require(entitySets.containsKey(entitySetId)) {
+            "Entity set $entitySetId not found"
+        }
+        entitySets.executeOnKey(entitySetId) {
```
makosblade, line 184: I'm a little confused on this new syntax. Does this create an implicit serializer that can serialize `Set<Int>`? Or is there a requirement that only objects that are registered can be used here and hz just figures it out? 
Either way unless hz added `Set<Int>` as a built-in, partitions should probably be wrapped by a delegatedIntSet or whichever wrapper object we have for Set<Int>

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+DELETE FROM ${ID.name} 
+    USING (SELECT ${ID.name},${ENTITY_SET_ID.name},${PARTITION.name} FROM ${IDS.name} INNER JOIN (select ? as ${ENTITY_SET_ID.name},? as ${PARTITIONS.name} ) as es USING (${ENTITY_SET_ID.name})) as to_be_deleted
+    WHERE ${PARTITION.name} = ? AND ${PARTITION.name}!=$REPARTITION_SELECTOR AND to_be_deleted.${ID.name} = ids.${ID.name} and to_be_deleted.${PARTITION.name} = ids.${PARTITION.name};  
+""".trimIndent()
+
+/**
+ * Computes the actual partition and compares it to current partition. If partitions do not match deletes the row.
+ *
+ * 1. entity set id
+ * 2. partition
```
makosblade, line 346: I think this needs:

```
3. entity set id
4. partition
```

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+private val DELETE_DATA_SQL = """
+DELETE FROM ${DATA.name} 
+    USING (SELECT ${ID.name},${ENTITY_SET_ID.name},${PARTITION.name} FROM ${DATA.name} INNER JOIN (select ? as ${ENTITY_SET_ID.name},? as ${PARTITIONS.name} ) as es USING (${ENTITY_SET_ID.name})) as to_be_deleted
+    WHERE ${PARTITION.name} = ? AND partition!=$REPARTITION_SELECTOR AND to_be_deleted.${ID.name} = ids.${ID.name} and to_be_deleted.${PARTITION.name} = ids.${PARTITION.name};  
+""".trimIndent()
+
+/**
+ * Computes the actual partition and compares it to current partition. If partitions do not match deletes the row.
+ * 1. entity set id
+ * 2. partition
```
makosblade, line 334: I think this needs:

```
3. partition
```

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+    WHERE ${PARTITION.name} = ? AND ${PARTITION.name}!=$REPARTITION_SELECTOR
+    ON CONFLICT DO UPDATE SET
+        ${latestSql(VERSION, VERSION)},
+        $VERSIONS = ARRAY( SELECT DISTINCT UNNEST(${VERSIONS.name} || EXCLUDED.${VERSIONS.name} ) ORDER BY 1  )
+""".trimIndent()
+
+/**
+ * Computes the actual partition and compares it to current partition. If partitions do not match deletes the row.
+ * 1. entity set id
+ * 2. partition
```
makosblade, line 323: I think this needs:

```
3. partition
```

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+        ${latestSql(LAST_PROPAGATE, VERSION)},
+        ${latestSql(LAST_MIGRATE, VERSION)},
+        ${latestSql(LAST_LINK_INDEX, VERSION)}
+""".trimIndent()
+
+/**
+ * Query for repartition a partition of edges.
+ *
+ * 1. entity set id
+ * 2. partition
```
makosblade, line 307: I think this needs:

```
3. partition
```

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+        $VERSIONS = ARRAY( SELECT DISTINCT UNNEST(${VERSIONS.name} || EXCLUDED.${VERSIONS.name} ) ORDER BY 1  )  
+        ${latestSql(LAST_WRITE, VERSION)},
+        ${latestSql(LAST_PROPAGATE, VERSION)},
+""".trimIndent()
+
+/**
+ * Query for repartition a partition of ids.
+ *
+ * 1. entity set id
+ * 2. partition
```
makosblade, line 283: I think this needs:

```
3. partition
```

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+
+private val REPARTITION_DATA_COLUMNS = buildRepartitionColumns(DATA)
+private val REPARTITION_IDS_COLUMNS = buildRepartitionColumns(E)
+private val REPARTITION_EDGES_COLUMNS = buildRepartitionColumns(IDS)
+
+/**
+ * Query for repartition a partition of data.
+ *
+ * 1. entity set id
+ * 2. partition
```
makosblade, line 260: I think this needs:

```
3. partition
```

src/main/java/com/openlattice/postgres/PostgresTable.java
geekbeast: If the table isn't in this class the PostgresTables class won't be able to automatically pick it up for PostgresTableManager.

makosblade: Ahh okay, I didn't realize PostgresTable gets scanned. Cool 👍

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
geekbeast: Kotlin lambda functions are java or closure serialized, so hazelcast won't see the captured Set<Int>. In this case, I figured that the rare invocation of the entry processor doesn't justify implementing a separate EntryProcessor class, StreamSerializer, and test. As a side note, if I was returning a Set<Int> I could wrap with DelegatedIntSet as the return values from the entry processor will be serialized using the traditional serialization mechanisms.

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
geekbeast: Missed a delete during cleanup.

src/test/kotlin/com/openlattice/data/ids/PostgresEntityKeyIdServiceTest.kt
```diff
@@ -147,9 +147,9 @@ class PostgresEntityKeyIdServiceTest : TestServer() {
     @Test
     fun testMassiveInsertAndReadBack() {
         logger.info("Testing insertion and massive readback")
-        val smallBatchSize = 8192
-        val medBatchSize = 32768
-        val largeBatchSize = 128000
+        val smallBatchSize = 128
+        val medBatchSize = 512
+        val largeBatchSize = 2048
```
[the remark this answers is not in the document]
geekbeast: It didn't trigger the error at values of any size and I had to update the build.gradle to prevent OOMs. I left them in, because they did catch some logic errors in entity key id generation separate from the deadlock (mainly around duplicate key assignment) and you didn't need ridiculous batch sizes for that.

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
makosblade: Cool yeah the return wrapping makes sense. I don't know that I understand how serializing a lambda would avoid having to also serialize the `Set<Int>` to each node that executes this code though 🤔. If you're saying it *does* actually serialize the `Set<Int>` then cool, and I agree that seldom invoked java serializers are totally fine in this case.

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+            state.deleteCount += delete(DELETE_IDS_SQL)
+            state.deleteCount += delete(DELETE_EDGES_SQL)
+        }
+
+        result = state.repartitionCount + state.deleteCount
+        hasWorkRemaining = (++state.currentlyMigratingPartitionIndex < state.oldPartitions.size)
+
+        //TODO: Consider adding completion hook to distributable jobs framework
+        //Once we are done, set the partitions.
+        if (!hasWorkRemaining && phase == 0) {
```
katherinebernstein, line 107: Does phase 0 mean complete? (It might be helpful to use en enum or constants to name these phases)

Unrelatedly, what's the reason for waiting until completion to update the partitions on the entity set itself? Won't any incoming writes during the repartition be competing with the repartition job? (I realize we control these jobs so we could just shut down integrations while running, but seems like if there's a way to minimize/reduce interruption, it would be preferable)

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+            setPartitions(state.entitySetId, state.newPartitions)
+            //The 4*getCount was an estimate, we remove estimate and addin updated value.
+            state.needsMigrationCount /= 2
```
katherinebernstein, line 110: Where is the `4*getCount` that the comment is referring to, and why is needsMigrationCount halved here?

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+            state.needsMigrationCount += getNeedsMigrationCount()
+            state.currentlyMigratingPartitionIndex = 0
+            hasWorkRemaining = true
+            phase = 1
```
katherinebernstein, line 114: ^ okay, maybe phase 0 doesn't mean complete...

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+    override fun hashCode(): Int {
+        var result = super.hashCode()
+        result = 31 * result + phase
+        return result
+    }
+
+}
+
+
+private val REPARTITION_SELECTOR = "partitions[ 1 + ((array_length(partitions,1) + (('x'||right(${ID.name}::text,8))::bit(32)::int % array_length(partitions,1))) % array_length(partitions,1))]"
```
katherinebernstein, line 212: Seems ideal to have one central generating function for this that takes the id column name and maybe partitions column name too -- I realize this should never change, but I feel like we've copy/pasted most of this a bunch of times and seems like these could easily get out of sync

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -0,0 +1,352 @@
...
+
+/**
+ * Computes the actual partition and compares it to current partition. If partitions do not match deletes the row.
+ * 1. entity set id
+ * 2. partition
+ */
+private val DELETE_DATA_SQL = """
+DELETE FROM ${DATA.name} 
+    USING (SELECT ${ID.name},${ENTITY_SET_ID.name},${PARTITION.name} FROM ${DATA.name} INNER JOIN (select ? as ${ENTITY_SET_ID.name},? as ${PARTITIONS.name} ) as es USING (${ENTITY_SET_ID.name})) as to_be_deleted
+    WHERE ${PARTITION.name} = ? AND partition!=$REPARTITION_SELECTOR AND to_be_deleted.${ID.name} = ids.${ID.name} and to_be_deleted.${PARTITION.name} = ids.${PARTITION.name};
```
katherinebernstein, line 356: is this supposed to be `ids` and not `data` ?

katherinebernstein: (and for the edges query)

geekbeast: Yup.

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
geekbeast: This one is definitely different from all the previous versions as it can never return negative. Happy to move it somewhere else.

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
geekbeast: We need a denominator for computing progress, but we don't know what will be written while the job is running. We know a lowerbound for items that will get issued in phase 2 (which is why we double to four). Once phase 2 starts and partitions for an entity set have been updated we halve the migration count and add in the actual phase 2 count, before progress gets updated. The progress bar may seem to go backwards for a bit as a result though :grimacing:

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
geekbeast: Phase 0 populates data into new partitions, we then mark new partitions, then phase 1 merges in any changes since phase 0.

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
geekbeast: It's to maximize availability. If we set partitions, then getEntities call will use newPartitions and reads will fail. This way most of the data is copied and there is instead a small window where stale reads may happen.

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
katherinebernstein: ohhh I see. Makes sense! Yeah backwards progress bar seems fine as a ballpark indicator

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
geekbeast: IIRC, lambdas are implemented as anonymous functions which include a closure of all captured variables that are out of scope. So there's basically a small class with a Set<Int> field that gets generated and is serialized using java/closure serialization (I'd have to look up if Kryo serialization is still configured). Most implementation of Set<Int> are serializable so this works. If partitions is invoked with non-serializable set then this will explode.

There's a couple of more rules to make sure that you don't accidentally capture the enclosing class for serialization, but that only happens when you invoke something on this or pass function reference to a class function.

PR #905, opened 2020-08-04
get the "name", "given_name", and "family_name" fields from auth0

UnsungHero97:
depends on https://github.com/geekbeast/rhizome/pull/333

### DIFF
--- a/src/main/kotlin/com/openlattice/users/Auth0UserListingService.kt
+++ b/src/main/kotlin/com/openlattice/users/Auth0UserListingService.kt
@@ -68,11 +68,7 @@ class Auth0UserListingService(
      */
     private fun getUsers(auth0ApiExtension: Auth0ApiExtension): List<User> {
         val exportEntity = auth0ApiExtension.userExport()
-        val job = exportEntity.submitExportJob(
-                UserExportJobRequest(
-                        listOf(USER_ID, EMAIL, NICKNAME, APP_METADATA, IDENTITIES, NAME, GIVEN_NAME, FAMILY_NAME)
-                )
-        )
+        val job = exportEntity.submitExportJob(UserExportJobRequest(AUTH0_USER_FIELDS))
 
         // will fail if export job hangs too long with too many requests error (429 status code)
         // https://auth0.com/docs/policies/rate-limits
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/openlattice/users/Auth0Utils.kt
```diff
@@ -28,6 +28,7 @@ fun getUpdatedUsersPage(
                     .withSearchEngine(SEARCH_ENGINE_VERSION)
                     .withQuery("$UPDATED_AT={$lastSync TO $currentSync]")
                     .withFields("$USER_ID,$EMAIL,$NICKNAME,$APP_METADATA,$IDENTITIES", true)
+                    .withFields("$NAME,$GIVEN_NAME,$FAMILY_NAME", true)
```
katherinebernstein, line 31: Is there a reason this isn't on the previous `withFields` line / does this for sure not overwrite that?

UnsungHero97: Not a good reason, just to avoid long lines. I'm trying to test locally to make sure it doesn't override.

katherinebernstein: Looking at the parent class it looks like this will override.

UnsungHero97: Roger - fixing.

src/main/kotlin/com/openlattice/users/Auth0UserListingService.kt
```diff
@@ -69,7 +69,9 @@ class Auth0UserListingService(
     private fun getUsers(auth0ApiExtension: Auth0ApiExtension): List<User> {
         val exportEntity = auth0ApiExtension.userExport()
         val job = exportEntity.submitExportJob(
-                UserExportJobRequest(listOf(USER_ID, EMAIL, NICKNAME, APP_METADATA, IDENTITIES))
+                UserExportJobRequest(
+                        listOf(USER_ID, EMAIL, NICKNAME, APP_METADATA, IDENTITIES, NAME, GIVEN_NAME, FAMILY_NAME)
```
geekbeast, line 73: Since you are here, why don't you extract these into a const list?

src/main/kotlin/com/openlattice/users/Auth0Utils.kt
```diff
@@ -27,7 +27,10 @@ fun getUpdatedUsersPage(
             UserFilter()
                     .withSearchEngine(SEARCH_ENGINE_VERSION)
                     .withQuery("$UPDATED_AT={$lastSync TO $currentSync]")
-                    .withFields("$USER_ID,$EMAIL,$NICKNAME,$APP_METADATA,$IDENTITIES", true)
+                    .withFields(
+                            "$USER_ID,$EMAIL,$NICKNAME,$APP_METADATA,$IDENTITIES,$NAME,$GIVEN_NAME,$FAMILY_NAME",
```
geekbeast, line 31: then here just refer to that constant list list.joinToString(",")

PR #907, opened 2020-08-06
Implement queuing service that uses postgres to manage work batches

geekbeast:
We have a lot of different types of concurrent work queues that operate on the ids table, but no good scalable mechanism for managing that work. This service provides the building blocks for (1) queuing a stream of ids based off a query of the ids table (2) operating on that set of ids as stream (3) releasing the resources held by that set of ids.

Remaining items:
- Create an index on batch type to speed up counting operation.
- Switch linking indexing to use this service. It's already broken and existing logic is based on incorrect assumption that you must get all entity data keys for linking ids, when jvm side group by operation by entity set id is all that is need to invoke index operation.
- (One day) Switch indexing to use this service.


src/main/kotlin/com/openlattice/batching/EntityBatchingService.kt
```diff
@@ -0,0 +1,167 @@
...
+    fun enqueue(
+            batchType: String,
+            idsSql: String,
+            batchSize: Int = DEFAULT_BATCHING_SIZE,
+            bind: (PreparedStatement, Int) -> Int
+    ) {
+        hds.connection.use { connection ->
+            connection.prepareStatement(buildInsertBatch(idsSql)).use { ps ->
+                ps.setString(1, batchType)
+                val offset = bind(ps, 1)
```
katherinebernstein, line 44: Does `bind` increment the offset before binding? I feel like I would expect this to be the starting index for bind (ie 2) -- might be with either updating that or updating the comment to explicitly state that this is the last set index, not the starting index

src/main/kotlin/com/openlattice/batching/EntityBatchingService.kt
```diff
@@ -0,0 +1,167 @@
...
+        connection.autoCommit = false
+        try {
+            var remaining = batchSize
+            val batch = mutableListOf<EntityBatchEntry>()
+
+            /**
+             * Since this is a skip locked query, it won't ever deadlock and will simply skipped locked rows to process
+             * unlocked rows.
+             */
+            val result = connection.prepareStatement(SELECT_BATCH).use { ps ->
```
katherinebernstein, line 92: Any reason not to use a BasePostgresIterable here? Seems preferable to get all the built-in error handling there

geekbeast: BasePostgresIterable acquires and closes connections, which is incompatible with the skip locks needed here.

PR #909, opened 2020-08-07
copy members to set before EP

katherinebernstein:


src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -286,11 +286,13 @@ class HazelcastOrganizationService(
 
     private fun addMembers(
             orgAclKey: AclKey,
-            members: Set<Principal>,
+            membersToAdd: Set<Principal>,
             profiles: Map<Principal, Map<String, Set<String>>>
     ): Set<Principal> {
         require(orgAclKey.size == 1) { "Organization acl key should only be of length 1" }
 
+        val members = membersToAdd.toSet()
```
geekbeast, line 294: val member = if( members is Serializable) membersToAdd else membersToAdd.toSet()

PR #910, opened 2020-08-07
add Thread.sleep in auth0 job

katherinebernstein:


src/main/kotlin/com/openlattice/users/Auth0UserListingService.kt
```diff
@@ -77,6 +77,8 @@ class Auth0UserListingService(
             exportJobResult = exportEntity.getJob(job.id)
         }
 
+        Thread.sleep(1000) // TODO actually fix
```
makosblade, line 80: https://jira.openlattice.com/browse/LATTICE-2538

PR #912, opened 2020-08-08
galaxy brain fix.

geekbeast:


src/main/kotlin/com/openlattice/users/Auth0SyncService.kt
```diff
@@ -163,7 +163,7 @@ class Auth0SyncService(
             var nextAclKeyLayer = childAclKeys.toSet()
 
             while (nextAclKeyLayer.isNotEmpty()) {
-                val nextAclKeyLayer = (nextAclKeyLayer.flatMapTo(mutableSetOf<AclKey>()) {
+                nextAclKeyLayer = (nextAclKeyLayer.flatMapTo(mutableSetOf<AclKey>()) {
```
makosblade, line 166: I think IntelliJ has a warn for this that we should make sure is highly enabled. 
I'd prefer setting it to error level honestly

PR #913, opened 2020-08-10
Fix missing in map error when principal trees are properly cleaned up…

geekbeast:
… for deleting objects.


geekbeast:
> LGTM but it may be a good idea to add a map consistency checker to the AdminApi that we can run on demand
 
Yeah, I think we have a whole bunch of crud cleanup that could be setup as background jobs.

makosblade:
> > LGTM but it may be a good idea to add a map consistency checker to the AdminApi that we can run on demand
> 
> Yeah, I think we have a whole bunch of crud cleanup that could be setup as background jobs.

https://jira.openlattice.com/browse/LATTICE-2541
👍

PR #914, opened 2020-08-11
Base 64 decoder

makosblade:
Needed for decoding base64 encoded images before they're integrated


jokedurnez:
Hm, is this necessary?  In light of trying to remove transforms?  We should write `bytea` into the atlas table before integration, right?

makosblade:
> Hm, is this necessary? In light of trying to remove transforms? We should write `bytea` into the atlas table before integration, right?

It's not necessary if there's a data cleaning step that does base64 decode but it sounded like there wasn't so I made this to unblock the xml integrations

PR #919, opened 2020-08-13
LATTICE-2542 #Ready Always trigger granting process and always persis…

geekbeast:
…t updated value when membership change is requested.

### DIFF
--- a/src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
+++ b/src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
@@ -129,7 +129,7 @@ class HazelcastOrganizationService(
         when (principal.type) {
             PrincipalType.USER ->
                 //Add the organization principal to the creator marking them as a member of the organization
-                addMembers(organization.getAclKey(), ImmutableSet.of(principal), mapOf())
+                addMembers(organization.getAclKey().first(), ImmutableSet.of(principal), mapOf())
             PrincipalType.ROLE ->
                 //For a role we ensure that it has
                 logger.debug("Creating an organization with no members, but accessible by {}", principal)
@@ -273,7 +273,12 @@ class HazelcastOrganizationService(
             profiles: Map<Principal, Map<String, Set<String>>> = members
                     .associateWith { getAppMetadata(users.getValue(it.id)) }
     ) {
-        val newMembers = addMembers(AclKey(organizationId), members, profiles)
+        val nonUserPrincipals = members.filter { it.type != PrincipalType.USER }
+        require(nonUserPrincipals.isEmpty()) { "Cannot add non-users principals $nonUserPrincipals to organization $organizationId" }
+
+        val securablePrincipals = members.associateWith { principal -> securePrincipalsManager.lookup(principal) }
+
+        val newMembers = addMembers(AclKey(organizationId), securablePrincipals, profiles)
 
         if (newMembers.isNotEmpty()) {
             val securablePrincipals = securePrincipalsManager.getSecurablePrincipals(newMembers)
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -285,44 +290,47 @@ class HazelcastOrganizationService(
...
         }) as Set<Principal>
 
-        if (newMembers.isNotEmpty()) {
-            grantOrganizationPrincipals(organizationId, newMembers, orgAclKey, profiles)
-        }
+        //Always trigger as this won't cause a write to organizations table.
+        grantOrganizationPrincipals(
+                organizationId,
+                membersToAdd,
+                orgAclKey,
```
UnsungHero97, line 316: Silly question... is the `orgAclKey` param really necessary? Isn't the `organizationId` param enough info?

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -285,44 +290,47 @@ class HazelcastOrganizationService(
...
+            members: Map<Principal, AclKey>,
             orgAclKey: AclKey = AclKey(organizationId),
             profiles: Map<Principal, Map<String, Set<String>>> = mapOf()
     ) {
         /*
          * Grant each member any of the roles, for which they meet the crieria.
          */
         val organization = organizations.getValue(organizationId)
-        members.forEach { member ->
+        members.forEach { (member, _) ->
```
UnsungHero97, line 333: I see `_` being used below, but it wasn't defined prior to this change... how did the code compile?!

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -285,44 +290,47 @@ class HazelcastOrganizationService(
...
+            return@OrganizationEntryProcessor Result(newMembers)
         }) as Set<Principal>
 
-        if (newMembers.isNotEmpty()) {
-            grantOrganizationPrincipals(organizationId, newMembers, orgAclKey, profiles)
-        }
+        //Always trigger as this won't cause a write to organizations table.
+        grantOrganizationPrincipals(
+                organizationId,
+                membersToAdd,
```
UnsungHero97, line 315: Why not `newMembers`?

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
geekbeast: _ is special syntactic sugar that means this argument won't be used so there is no variable _

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
geekbeast: It's a compatibility thing for callers. You don't have to provide the parameter as it will be automatically instantiated from the provided organizationId. It can be useful to avoid making an unnecessary copy if the caller has the organization acl key around for some reason.

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
geekbeast: To be fail safe. There was a weird bug where if grantOrganizationPrincipals fails (say stack reboot) you can end up with a member of the org that isn't visible as a member (listing members uses grant). Adding the user again will cause the grant to be re-attempted even if members list is not modified.

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -285,44 +290,47 @@ class HazelcastOrganizationService(
...
             it.members.addAll(newMembers)
-            return@OrganizationEntryProcessor Result(newMembers, newMembers.isNotEmpty())
+            //We're always persisting here.
+            return@OrganizationEntryProcessor Result(newMembers)
         }) as Set<Principal>
 
-        if (newMembers.isNotEmpty()) {
-            grantOrganizationPrincipals(organizationId, newMembers, orgAclKey, profiles)
-        }
+        //Always trigger as this won't cause a write to organizations table.
```
katherinebernstein, line 312: Not a huge deal, but it does always trigger a read from the organizations mapstore (even if there's no work to be done) -- maybe we should add a 
```
if (members.isEmpty()) {
  return
}
```
at the beginning of `grantOrganizationPrincipals`

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -273,7 +273,12 @@ class HazelcastOrganizationService(
             profiles: Map<Principal, Map<String, Set<String>>> = members
                     .associateWith { getAppMetadata(users.getValue(it.id)) }
     ) {
-        val newMembers = addMembers(AclKey(organizationId), members, profiles)
+        val nonUserPrincipals = members.filter { it.type != PrincipalType.USER }
+        require(nonUserPrincipals.isEmpty()) { "Cannot add non-users principals $nonUserPrincipals to organization $organizationId" }
+
+        val securablePrincipals = members.associateWith { principal -> securePrincipalsManager.lookup(principal) }
```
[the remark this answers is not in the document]
makosblade: ➕

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
geekbeast: This seems to make more sense in the public addMembers call, since we shouldn't do any work at all if members is empty.

Optimization we could do in grantOrganizationPrincipals is push building out the list of roles to be granted to an entry processor, which is then returned the changes are completed. Also looking at it our e-mail processing is probably not RFC-compliant either.

katherinebernstein: Cool, so you mean restore the old
```
if (newMembers.isNotEmpty()) {
  grantOrganizationPrincipals(...)
}
```
code? Doesn't seem to avoid any additional work but I'm good with that (although putting that logic in `grantOrganizationPrincipals` seems like a more thorough way to avoid doing the extra load. But again, just one hazelcast read so not a huge deal)

Doing that role work in an EP also sounds ideal if possible.

How does this interact with email processing?

PR #923, opened 2020-08-20
no more edges, no more entity key ids

makosblade:
This PR:
- Removes long unused edges and entity_key_ids table objects

### DIFF
--- a/src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
+++ b/src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
@@ -24,7 +24,6 @@ package com.openlattice.data.storage
 import com.openlattice.analysis.requests.Filter
 import com.openlattice.postgres.DataTables.*
 import com.openlattice.postgres.PostgresColumn.*
-import com.openlattice.postgres.PostgresTable.ENTITY_KEY_IDS
 import com.openlattice.postgres.ResultSetAdapters
 import org.slf4j.LoggerFactory
 import java.util.*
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
```diff
@@ -234,7 +233,7 @@ internal fun buildEntitiesClause(entityKeyIds: Map<UUID, Optional<Set<UUID>>>, l
                 .filter { ids -> ids.isNotEmpty() }
                 .map { ids -> " $idsColumn IN (" + ids.joinToString(",") { id -> "'$id'" } + ") AND" }
                 .orElse("")
-        " ($idsClause ${ENTITY_KEY_IDS.name}.${ENTITY_SET_ID.name} = '${it.key}' )"
+        " ($idsClause entity_key_ids.${ENTITY_SET_ID.name} = '${it.key}' )"
```
geekbeast, line 236: Hmmm...

src/main/kotlin/com/openlattice/data/storage/PostgresQueries.kt
```diff
@@ -218,7 +217,7 @@ internal fun selectEntityKeyIdsWithCurrentVersionSubquerySql(
         }
     }
 
-    return "( SELECT $selectColumns FROM ${ENTITY_KEY_IDS.name} INNER JOIN $FILTERED_ENTITY_KEY_IDS USING($joinColumnsSql) WHERE true $entitiesClause $groupBy ) as $ENTITIES_TABLE_ALIAS"
+    return "( SELECT $selectColumns FROM entity_key_ids INNER JOIN $FILTERED_ENTITY_KEY_IDS USING($joinColumnsSql) WHERE true $entitiesClause $groupBy ) as $ENTITIES_TABLE_ALIAS"
```
geekbeast, line 220: I think entity_key_ids replace should have been switched to IDS.name

makosblade: This class is appears to be a leaf node in a chain of deprecated/unused/commented code so I just switched it to the historical name for that variable. I can swap this class to IDS but I'm fairly sure it won't work/is unused

geekbeast: Oh right. Good point, we should see if we can delete this.

geekbeast: https://jira.openlattice.com/browse/LATTICE-2547

PR #925, opened 2020-08-20
cleanup how auth0 is initialized so that we stop depending on shuttle for one method call

makosblade:
This PR:
- Uses new auth0 client configuration structures

### DIFF
--- a/src/main/java/com/openlattice/hazelcast/StreamSerializerTypeIds.java
+++ b/src/main/java/com/openlattice/hazelcast/StreamSerializerTypeIds.java
@@ -30,7 +30,7 @@ public enum StreamSerializerTypeIds {
     /**
      * Move this one, break everything.
      */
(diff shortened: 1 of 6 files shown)


src/main/java/com/openlattice/hazelcast/StreamSerializerTypeIds.java
```diff
@@ -30,7 +30,7 @@
     /**
      * Move this one, break everything.
      */
-    STREAM_SERIALIZER_IDS_MUST_BE_POSTIVE,
+    STREAM_SERIALIZER_IDS_MUST_BE_POSITIVE,
```
katherinebernstein, line 33: unsure what changed here but I'm afraid 😨

katherinebernstein: oh yes. postive.

PR #935, opened 2020-08-24
Don't use joins to drive deletes.

geekbeast:
Still need to test these queries to make sure array indexing is acceptable, but otherwise should be good.


src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -840,6 +840,15 @@ fun getPartitioningSelector(
...
 
+/**
+ * 1. partitions
+ * 2. array length ( partitions )
+ * 3. array length ( partitions )
+ * 4. array length ( partitions )
+ */
+fun getDirectPartitioningSelector(
+        idColumn: String
+) = "(?)[ 1 + ((? + (('x'||right(${idColumn}::text,8))::bit(32)::int % ?)) % ?)]"
```
makosblade, line 851: It appears as though this pattern has been here but can you add a line for what its doing in the javadoc? I cannot decipher it 😆

src/main/kotlin/com/openlattice/graph/partioning/RepartitioningJob.kt
```diff
@@ -375,71 +383,54 @@ INSERT INTO ${E.name} SELECT $REPARTITION_EDGES_COLUMNS
...
-    USING (SELECT ${ID.name},${ENTITY_SET_ID.name},${PARTITION.name},${PARTITIONS.name} FROM ${DATA.name} INNER JOIN (select ? as ${ENTITY_SET_ID.name},? as ${PARTITIONS.name} ) as es USING (${ENTITY_SET_ID.name})) as to_be_deleted
-    WHERE 
-      ${DATA.name}.${PARTITION.name} = ?
-      AND ${DATA.name}.${ENTITY_SET_ID.name} = ?
-      AND ${DATA.name}.${PARTITION.name}!=${getPartitioningSelector(DATA.name + "." + ID.name)} 
-      AND to_be_deleted.${ID.name} = ${DATA.name}.${ID.name} 
-      AND to_be_deleted.${PARTITION.name} = ${DATA.name}.${PARTITION.name};  
+    WHERE
+      ${DATA.name}.${ENTITY_SET_ID.name} = ?
+      AND ${DATA.name}.${PARTITION.name} = ?
```
katherinebernstein, line 397: Any reason not to do this as one big batch delete, ie `WHERE partition = ANY(allOldPartitions)`

geekbeast: It processes one partition at a time so that it can resume on the current partition if the job is paused or the node crashes.

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
geekbeast: Ha! Sure. I barely remember it myself. It converts uuid to text, takes the right-most 8 characters, prepends an x, casts it to 32 bit string interpreting it as hex, cast those 32 bits as an int, then uses the size of partition to compute the positive mod index into the array.

makosblade: 👍

src/main/kotlin/com/openlattice/data/storage/PostgresDataQueries.kt
```diff
@@ -831,7 +831,7 @@ fun selectPropertyTypesOfEntitySetColumnar(
  */
 fun getPartitioningSelector(
         idColumn: PostgresColumnDefinition
-) = getPartitioningSelector(idColumn.name)
+) = getPartitioningSelector(idColumn.name) + Math.mo
```
katherinebernstein, line 834: 🤔

PR #940, opened 2020-09-04
load all apps for an organization

katherinebernstein:


geekbeast (approved):

makosblade (approved):

src/main/kotlin/com/openlattice/apps/services/AppService.kt
```diff
@@ -366,6 +369,23 @@ class AppService(
...
+            appId: UUID,
+            organizationId: UUID,
+            roleMappings: Map<UUID, AclKey>
+    ) {
+        ensureAppRolesExist(getApp(appId), roleMappings.keys)
+        principalsService.ensurePrincipalsExist(roleMappings.values.toSet())
+
+        appConfigs.executeOnKey(AppConfigKey(appId, organizationId)) {
+            val setting = it.value
+            roleMappings.forEach { (roleId, roleAclKey) ->
```
geekbeast, line 382: Make sure the map itself is always serializable. Had problems in Organization Service when folks were passing in a map that wasn't serializable.

src/main/kotlin/com/openlattice/apps/services/AppService.kt
```diff
@@ -435,4 +455,12 @@ class AppService(
         }
     }
 
+    private fun ensureAppRolesExist(app: App, roleIds: Set<UUID>) {
+        val missingRoleIds = Sets.difference(roleIds, app.appRoles.map { it.id }.toSet())
```
geekbeast, line 459: roleIds  - app.appRoles, the view created by Sets.difference is expensive and you have to create another set by calling .toSet() anyway.

geekbeast (changes requested):

UnsungHero97 (approved):

geekbeast (approved):

PR #941, opened 2020-09-11
PermissionsApi.getAcls

UnsungHero97:
depends on https://github.com/openlattice/api/pull/475

### DIFF
--- a/src/main/kotlin/com/openlattice/authorization/HazelcastAuthorizationService.kt
+++ b/src/main/kotlin/com/openlattice/authorization/HazelcastAuthorizationService.kt
@@ -426,7 +426,10 @@ class HazelcastAuthorizationService(
 
     @Timed
     override fun getAllSecurableObjectPermissions(keys: Set<AclKey>): Set<Acl> {
-        return keys.map { key -> getAllSecurableObjectPermissions(key) }.toSet()
+        return aces.entrySet(hasAnyAclKeys(keys))
+                .groupBy { it.key.aclKey }
+                .map { entry -> Acl(entry.key, entry.value.map { Ace(it.key.principal, it.value.permissions) }.toSet()) }
+                .toSet()
     }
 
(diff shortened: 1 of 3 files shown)


geekbeast (approved):

src/main/kotlin/com/openlattice/authorization/HazelcastAuthorizationService.kt
```diff
@@ -421,6 +424,11 @@ class HazelcastAuthorizationService(
         return Acl(key, acesWithPermissions)
     }
 
+    @Timed
+    override fun getAllSecurableObjectPermissions(keys: Set<AclKey>): Set<Acl> {
+        return keys.map { key -> getAllSecurableObjectPermissions(key) }.toSet()
```
katherinebernstein, line 429: Given perf issues we've had with permissions calls in the past, this one worries me a little bit. Each `getAllSecurableObjectPermissions` is pretty heavy, and it seems likely that this could be called on a large set of AclKeys (if all of an org's entity sets are passed in, for instance). My vote would be to do a 
```aces.entrySet(hasAnyAclKey(keys))``` and then group by aclKey and process the result into the expected format

UnsungHero97: I have the same worry as well. My goal with this PR is to find a proper alternative to calling `getAllSecurableObjectPermissions` for each `AclKey`.

katherinebernstein: It does, if I'm understanding right -- it'll basically pull back any entries that match any of the aclKeys you pass it (which I think is what you want -- load *all* permissions for *all* the aclKeys in `keys`)

katherinebernstein (approved):

makosblade (approved):

PR #955, opened 2020-09-25
Don't assign org names a normal user id.

geekbeast:


geekbeast:
Hotfix for new org creation.

src/main/kotlin/com/openlattice/assembler/PostgresRoles.kt
```diff
@@ -9,6 +9,7 @@ private const val ADFS_PREFIX = "adfs|"
 private const val AD_PREFIX = "ad|"
 private const val WAAD_PREFIX = "waad|"
 private const val INTERNAL_PREFIX = "ol-internal"
+const val ORGANIZATION_PREFIX = "ol-interal|organization|"
```
makosblade, line 12: typo

PR #960, opened 2020-09-30
Feature/make org dbs renamable

katherinebernstein:


src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -170,6 +176,10 @@ class HazelcastOrganizationService(
         }) as Set<UUID>
     }
 
+    fun getOrganizationDatabaseName(organizationId: UUID): String {
+        return organizationDatabases.getValue(organizationId).name
```
geekbeast, line 180: How big is the return value here? If large consider using offloadable entry processor to return just the name.

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -528,6 +539,33 @@ class HazelcastOrganizationService(
...
+        executeDatabaseNameUpdate(organizationId, newDatabaseName)
+
+        try {
+            assembler.renameOrganizationDatabase(currentDatabaseName, newDatabaseName)
+        } catch (e: Exception) {
+            logger.error("An error occurred while trying to rename org {} database name to {}",
+                    organizationId,
+                    newDatabaseName,
+                    e)
+            executeDatabaseNameUpdate(organizationId, currentDatabaseName)
```
geekbeast, line 564: This seems redundant here unless it is fully transactional. If it is a SQL exception then rename failed for some reason. Also without looking at this-- since we are accepting user provided input, it must be parameterized and likely means you need to build a SQL function to do safely.

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
katherinebernstein: It's an OrganizationDatabase object (defined in api) which is just an int and a String

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
katherinebernstein: I don't think it's redundant -- this call resets the OrganzationDatabase value in the mapstore to the previous value in the event the update fails. 

Re the sql injection -- what do you mean by sql function? My understanding is that prepared statements can't be used on an ALTER DATABASE update so I wasn't 100% sure how to do the sanitization other than the validate only alphanumeric/dashes/underscores piece at the top of this function.

katherinebernstein: oh but actually none of the code path in assembler references the value in the mapstore, so it's safe to just execute the mapstore update after the atlas db name update. I had done the update in the mapstore first originally so that a duplicate name would throw an exception, but updating the db name in atlas first should have that same effect.

geekbeast (approved):

makosblade (approved):

PR #978, opened 2020-10-08
Move back to updating data rows in place and not using linking rows

makosblade:


katherinebernstein (approved):

geekbeast (approved):
Double check that ids table is also being updated.

makosblade:
Yep, it was being updated in a separate method call so I rolled them into one

PR #989, opened 2020-10-23
return boolean instead of unit from EPs

katherinebernstein:


makosblade (approved):

geekbeast (approved):
You can also return Void? if you never plan on using it.

PR #992, opened 2020-10-24
Feature/paged neighbor query

katherinebernstein:


src/main/kotlin/com/openlattice/graph/core/GraphService.kt
```diff
@@ -0,0 +1,82 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * You can contact the owner of the copyright at support@openlattice.com
+ *
+ */
+package com.openlattice.graph.core
```
makosblade, line 20: If this is meant to replace the java version, we should probably `@Deprecated` the java one

src/main/kotlin/com/openlattice/search/SearchService.kt
```diff
@@ -493,201 +495,246 @@ class SearchService(
         )
     }
```
makosblade, line 497: This diff is a bit of a mystery to me. Is there a better commit to review this one on than the final branch?

PR #994, opened 2020-10-27
add API to roll user credential

katherinebernstein:


src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -256,10 +256,10 @@ class AssemblerConnectionManager(
         }
     }
 
-    fun updateCredentialInDatabase(organizationId: UUID, unquotedUserId: String, credential: String) {
+    fun updateCredentialInDatabase(unquotedUserId: String, credential: String) {
         val updateSql = updateUserCredentialSql(quote(unquotedUserId), credential)
 
-        extDbManager.connectToOrg(organizationId).connection.use { connection ->
+        atlas.connection.use { connection ->
```
makosblade, line 262: we should replace `atlas` with `extDbManager.connect("postgres")` since (pretty sure at least) `atlas` points to a connection that could be expired

makosblade (approved):

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
katherinebernstein: hm, there are a ton of places in that class that we do `atlas.connection.use {` -- should we update all of those?

makosblade: don't need to right this second but it seems like we should eventually

geekbeast (approved):

PR #996, opened 2020-10-27
Bugfix/lattice 2564 fix duplicate ids

geekbeast:
Add job for cleaning up duplicate ids.


src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,301 @@
...
+        }) { rs ->
+            ResultSetAdapters.id(rs) to mapper.readValue<List<EntityKey>>(rs.getString(DUPLICATES_FIELD))
+        }.toMap()
+
+        entityKeysAndIds
+                .asSequence()
+                .forEach { (id, entityKeys) ->
+                    val entitySetAppearances = entityKeys.groupingBy { it.entitySetId }.eachCount()
+
+                    if (entitySetAppearances.values.all { it == 1 }) {
```
katherinebernstein, line 82: This seems doable on a per-entity set basis, yeah? so if you have
```
{
  entitySet1: 1,
  entitySet2: 1,
  entitySet3: 2
}
```
then you could assign new ids for the entities in `entitySet1` and `entitySet2` and only have to manually address the collisions from `entitySet3`
(if I'm reading this correctly)

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,301 @@
...
+    private fun migrate(
+            ps: PreparedStatement,
+            originalId: UUID,
+            entitySetId: UUID
+    ) {
+        bind(ps, entitySetId, originalId)
+        ps.addBatch()
+    }
+
+    private fun update(
```
katherinebernstein, line 139: just double checking, `migrate` and `update` are identical and just separated because the arguments are semantically different, yeah?

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,301 @@
...
+            newId: UUID,
+            originalId: UUID
+    ) {
+        bind(ps, newId, originalId)
+        ps.addBatch()
+    }
+
+    private fun bind(ps: PreparedStatement, entitySetId: UUID, entityKeyId: UUID) {
+        ps.setObject(1, entitySetId)
+        ps.setObject(1, entityKeyId)
```
katherinebernstein, line 150: should be 2

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,301 @@
...
+        ps.setObject(1, newId)
+        ps.setObject(2, entityKey.entitySetId)
+        ps.setString(3, entityKey.entityId)
+    }
+
+    private fun recordCollision(id: UUID, entityKeys: List<EntityKey>) {
+        hds.connection.use { connection ->
+            connection.prepareStatement(INSERT_COLLISION).use {ps ->
+                ps.setObject(1, id )
+                entityKeys.forEach {
```
katherinebernstein, line 163: does this work with `ps.setObject(1, id )` outside the `forEach`? like addBatch doesn't clear the prepared values?

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,301 @@
...
+
+/**
+ * Returns the duplicates for a batch of ids.
+ * 1. ids - array of uuids
+ */
+private val SELECT_NEXT_DUPLICATES = """
+SELECT  ${ID.name}, 
+        jsonb_agg(jsonb_build_object(
+            '${SerializationConstants.ENTITY_SET_ID}', ${ENTITY_SET_ID.name},
+            '${ENTITY_SET_ID.name}${SerializationConstants.ENTITY_SET_ID}${ENTITY_SET_ID.name}',${ENTITY_ID.name})
```
katherinebernstein, line 217: what is this `${ENTITY_SET_ID.name}${SerializationConstants.ENTITY_SET_ID}${ENTITY_SET_ID.name}` value?

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,301 @@
...
+        INNER JOIN (select ${ID.name} as ${SRC_ENTITY_SET_ID.name}, ${PARTITIONS.name} FROM ${ENTITY_SETS.name} ) as es 
+        USING(${SRC_ENTITY_SET_ID.name})  
+""".trimIndent()
+
+/**
+ * 1. (new) dst_entity_key_id - uuid
+ * 2. (old) dst_entity_key_id - uuid
+ */
+private val UPDATE_E_DST = """
+UPDATE ${E.name} SET $DST_ENTITY_KEY_ID = ? WHERE $DST_ENTITY_KEY_ID = ?
```
katherinebernstein, line 278: I think these need `.name`s -- i.e. `SET ${DST_ENTITY_KEY_ID.name} = ?`, etc

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,301 @@
...
+private val UPDATE_E_DST = """
+UPDATE ${E.name} SET $DST_ENTITY_KEY_ID = ? WHERE $DST_ENTITY_KEY_ID = ? 
+""".trimIndent()
+
+/**
+ * 1. (new) edge_entity_key_id - uuid
+ * 2. (old) edge_entity_key_id - uuid
+ */
+private val UPDATE_E_EDGE = """
+UPDATE ${E.name} SET $EDGE_ENTITY_KEY_ID = ? WHERE $EDGE_ENTITY_KEY_ID = ?
```
katherinebernstein, line 286: same on the `.name`s

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,301 @@
...
+        ps.setObject(2, entityKey.entitySetId)
+        ps.setString(3, entityKey.entityId)
+    }
+
+    private fun recordCollision(id: UUID, entityKeys: List<EntityKey>) {
+        hds.connection.use { connection ->
+            connection.prepareStatement(INSERT_COLLISION).use {ps ->
+                ps.setObject(1, id )
+                entityKeys.forEach {
+                    ps.setObject(2, it)
```
katherinebernstein, line 164: also looks like this update query should take entitySetId as param 2 and entityid as param 3

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
geekbeast: yes

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
geekbeast: correct

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
geekbeast: Yup. Good catch.

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,307 @@
...
+/**
+ * 1. entity set id - uuid
+ * 2. id - uuid
+ */
+private val MIGRATE_IDS = """
+WITH for_migration as (DELETE FROM ${IDS.name} WHERE ${ENTITY_SET_ID.name} = ? AND ${ID.name} = ? RETURNING *)
+    INSERT INTO ${IDS.name} 
+        SELECT $REPARTITION_IDS_COLUMNS 
+        FROM for_migration 
+        INNER JOIN (select ${ID.name} as ${ENTITY_SET_ID.name}, ${PARTITIONS.name} FROM ${ENTITY_SETS.name} ) as es
```
katherinebernstein, line 249: is this joining on partitions? how does postgres compare those?

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,307 @@
...
+
+/**
+ * 1. entity set id - uuid
+ * 2. id - uuid
+ */
+private val MIGRATE_IDS = """
+WITH for_migration as (DELETE FROM ${IDS.name} WHERE ${ENTITY_SET_ID.name} = ? AND ${ID.name} = ? RETURNING *)
+    INSERT INTO ${IDS.name} 
+        SELECT $REPARTITION_IDS_COLUMNS 
+        FROM for_migration
```
katherinebernstein, line 248: I think we should probably also set `last_write = now()` to trigger reindexing, retransporting, etc for entities with new ids

src/main/kotlin/com/openlattice/ids/IdGenerationServiceDependent.kt
```diff
@@ -0,0 +1,32 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * You can contact the owner of the copyright at support@openlattice.com
+ *
+ *
+ */
+
+package com.openlattice.ids
+
+import com.openlattice.assembler.AssemblerConnectionManager
```
makosblade, line 24: unused

makosblade (approved):

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
geekbeast: Oof, it's actually worse. This made check the primary key of the ids table. Entity set id is not part of the primary key, so when there's a collision on ids only one entity ever survives on the ids table, so we don't actually want to delete as that would erase existing ids. Similar thing no the edges table, but unfortunately there are cases where we can't even be sure that it's recoverable if there's overlapping triples (should be rare). 

In terms of indexing, I will actually mark everything at the end as both the new and old stuff needs to be reindexed.

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
geekbeast: The USING clause will join on the entity_set_id field. Not that it matters since this we don't need this query anymore.

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
geekbeast: The overhead of loading the partitions for the entity_sets and computing the partitions for marking as indexed probably exceeds the benefit so I just brute force it since it is an indexed update anyway.

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
```diff
@@ -0,0 +1,318 @@
...
+""".trimIndent()
+
+/**
+ * Returns the next batch of ids of size [BATCH_SIZE]. Proof: If a partition contains an element
+ * smaller than the max id returned from the query then it would have been returned in the query instead of the max element.
+ *
+ * 1. id - uuid
+ */
+private val SELECT_NEXT_IDS = """
+SELECT DISTINCT ${ID.name} FROM ${SYNC_IDS.name} WHERE ${ID.name} > ? LIMIT $BATCH_SIZE)
```
katherinebernstein, line 247: Looks like that`)` at the end isn't supposed to be there

src/main/kotlin/com/openlattice/data/ids/jobs/FixDuplicateIdAssignmentJob.kt
katherinebernstein: following up on this one -- this looks like a funky string value

geekbeast: Not sure how this came back. fixed again.

PR #999, opened 2020-10-29
Bugfix/spot many bugs

katherinebernstein:


src/main/kotlin/com/openlattice/authorization/listeners/PermissionMapListener.kt
```diff
@@ -58,4 +58,10 @@ class PermissionMapListener(private val eventBus: EventBus) : EntryAddedListener
     override fun equals(other: Any?): Boolean {
```
makosblade, line 58: should regenerate this

src/main/kotlin/com/openlattice/conductor/rpc/ConductorConfiguration.kt
```diff
@@ -21,15 +20,15 @@ data class ConductorConfiguration(
...
-        return key
-    }
 
     companion object {
         @JvmStatic
-        fun key(): ConfigurationKey {
-            return key
-        }
+        @get:JvmName("key")
+        val key = SimpleConfigurationKey("conductor.yaml")
```
makosblade, line 27: can probably get rid of the key on line 11/12

src/main/java/com/openlattice/shuttle/EntityDefinition.java
```diff
@@ -44,6 +45,7 @@
 import static com.google.common.base.Preconditions.checkNotNull;
 
 @JsonInclude( value = Include.NON_EMPTY )
+@SuppressFBWarnings( "SE_BAD_FIELD" )
```
makosblade, line 48: Needs reason

src/main/kotlin/com/openlattice/apps/services/AppService.kt
```diff
@@ -118,7 +116,7 @@ class AppService(
     }
 
     fun deleteRoleFromApp(appId: UUID, roleId: UUID) {
-        val app = getApp(appId)
+        getApp(appId)
```
makosblade, line 119: should probably rename to something more meaningful (re: using as a null check)
Or just use
```
        requireNotNull(appId) { "App with id $appId does not exist." }
```
though actually its not nullable so. . . can just remove this line

geekbeast: This is just checking if app exists right? Seems like there's probably an unnecessary deep load of app going on here and we should have a method that just does containsKey on the app map.

makosblade (approved):

geekbeast (approved):

PR #1001, opened 2020-10-30
Fix/check if table already exists

makosblade:


katherinebernstein (approved):

geekbeast (approved):
Looks good. I also added a note on foreign data wrapper synchronization that occurred to me while reviewing this.

PR #1006, opened 2020-11-03
Fix writing of column entities.

geekbeast:


src/main/kotlin/com/openlattice/organizations/OrganizationMetadataEntitySetsService.kt
```diff
@@ -259,18 +270,18 @@ class OrganizationMetadataEntitySetsService(private val edmService: EdmManager)
...
 
         val id = dataGraphManager.getEntityKeyIds(
-                setOf(EntityKey(column.tableId, column.name))
+                setOf(EntityKey(organizationMetadataEntitySetIds.datasets, table.id.toString())) //TODO: Consider using OID
         ).first()
 
         dataGraphManager.mergeEntities(
                 organizationMetadataEntitySetIds.columns,
-                mapOf(columnId to columnEntity),
+                mapOf(id to datasetColumnEntity),
```
katherinebernstein, line 278: this should still just be `columnEntity` I think

geekbeast: Yeah, I saw that write after I pushed.

katherinebernstein (approved):

UnsungHero97 (approved):

PR #1013, opened 2020-11-06
fix entity set creation on prod and local startup

katherinebernstein:


geekbeast (approved):

UnsungHero97 (approved):

makosblade (approved):

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
```diff
@@ -543,30 +540,24 @@ class AssemblerConnectionManager(
...
-        return BasePostgresIterable(PreparedStatementHolderSupplier(hds, PRINCIPALS_SQL) { ps ->
-            ps.setString(1, PrincipalType.ROLE.name)
-        }) {
-            securePrincipalsManager.getSecurablePrincipal(ResultSetAdapters.aclKey(it)) as Role
-        }
+    fun getAllRoles(): Set<Role> {
+        return securePrincipalsManager.allRoles
     }
 
-    fun getAllUsers(): BasePostgresIterable<SecurablePrincipal> {
```
geekbeast, line 554: This now requires the list of all roles to fit in memory forever.

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
```diff
@@ -142,14 +139,16 @@ class HazelcastOrganizationService(
             else -> throw IllegalStateException("Only users and roles can create organizations.")
         }//Fall through by design
 
-        initializeOrganizationPrincipals(principal, organization)
+        initializeOrganizationPrincipal(principal, organization)
         initializeOrganization(organization)
         organizationMetadataEntitySetsService.initializeOrganizationMetadataEntitySets(organization.id)
```
geekbeast, line 144: Why not move this to the very end?

src/main/kotlin/com/openlattice/auditing/AuditRecordEntitySetsManager.kt
```diff
@@ -137,7 +137,8 @@ class AuditRecordEntitySetsManager(
         val firstUserPrincipal: Principal
 
         try {
-            firstUserPrincipal = ownerPrincipals.first { it.type == PrincipalType.USER }
+            firstUserPrincipal = ownerPrincipals.firstOrNull { it.type == PrincipalType.USER }
+                    ?: ownerPrincipals.first { it.type == PrincipalType.ORGANIZATION }
```
geekbeast, line 141: I'm not sure we've thought through using ORGANIZATION principal as the owner. 

If anything we should be using Organization ADMIN role as the owner of the catalog role and the audit entity sets. In other words, it seems more natural to allow ROLES to be owners of objects. 

It will also cause issues down the road when roles are deleted and there are objects are left without owners.

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
katherinebernstein: that was already a requirement

geekbeast (commented):
Minor comments except for big design change:

This is allow the organization principal to be the owner of both regular entity sets and audit entity sets. This effectively grants every in the organization owner over those datasets. I'm not sure this is what we want and there a bunch of other side effects that we will need to guard against if we allow roles/organizations to own entity sets.

src/main/kotlin/com/openlattice/organizations/tasks/OrganizationsInitializationTask.kt
```diff
@@ -113,7 +114,8 @@ class OrganizationsInitializationTask : HazelcastInitializationTask<Organization
         return setOf(
                 AuthorizationInitializationTask::class.java,
                 UsersAndRolesInitializationTask::class.java,
-                PostConstructInitializerTask::class.java
+                PostConstructInitializerTask::class.java,
+                EdmSyncInitializerTask::class.java
```
geekbeast, line 118: Nice.

src/main/kotlin/com/openlattice/auditing/AuditRecordEntitySetsManager.kt
katherinebernstein: Oooh true. Yeah seems like admin role should be the owner here (+ build in tracking for those roles someday). We should also block deletion of admin role if we're not doing that already

src/main/kotlin/com/openlattice/assembler/AssemblerConnectionManager.kt
geekbeast: Without check who wrote the line, it looks like it was intended to page through role by role, instead of pulling all roles to a single box. This will work fine for a while until we have a lot of roles so it's not that big of an issue.

katherinebernstein: yeah, but then we did a `.map` on them collecting them to a list 😝 

Anyway this should make things faster since this updates the logic to not do a separate hazelcast read for every role in the system anymore, but at some point we'll probably want to set up paging for reading out principals.

src/main/kotlin/com/openlattice/organizations/HazelcastOrganizationService.kt
katherinebernstein: I was trying to avoid reordering things that were working, but this seems like it should be safe to move. What would be the reason to move it?

PR #1016, opened 2020-11-17
Fix/try all resources

makosblade:
Unclosed ResultSets, which may or may not be closed by the underlying jdbc driver


katherinebernstein (approved):
Most (if not all) of these should probably be converted to hazelcast-only calls someday

PR #1017, opened 2020-11-18
get the external database table / atlas data set schema

UnsungHero97:


src/main/kotlin/com/openlattice/organizations/ExternalDatabaseManagementService.kt
```diff
@@ -248,6 +249,16 @@ class ExternalDatabaseManagementService(
         return organizationExternalDatabaseColumns.getValue(columnId)
     }
 
+    fun getExternalDatabaseTableSchema(organizationId: UUID, tableId: UUID): String? {
```
geekbeast, line 252: What if there are two tables with the same name for a given id? You should probably use the hidden oid column if available.

UnsungHero97: The `oid` column is not available in `select * from pg_catalog.pg_tables`. There is a problem though... it's possible to have the same table name but in different schemas, so the query in `getExternalDatabaseTableSchemaSql` will return multiple results in that case.

Not sure how to handle this 🤔

